PR #4, opened 2018-09-03
Repo setup and a basic matrix-scan-then-do-nothing keyboard

klardotsh:
It can even auto-deploy to my Feather in a single Makefile step

You'll need `python-pipenv` on Arch to use this in any reasonable fashion.

Currently don't have unit tests/PR lint checks enabled. They're coming soon/eventually.

### DIFF
--- a/boards/klardotsh/twotwo_matrix_feather.py
+++ b/boards/klardotsh/twotwo_matrix_feather.py
@@ -1,0 +1,26 @@
+import board
+import time
+
+from kmk.circuitpython.matrix import MatrixScanner
+from kmk.common.consts import DiodeOrientation
+from kmk.common.keymap import Keymap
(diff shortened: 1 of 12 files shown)


boards/klardotsh/twotwo_matrix_feather.py
```diff
@@ -0,0 +1,26 @@
...
+
+    keymap = Keymap([
+        ['A', 'B'],
+        ['C', 'D'],
+    ])
+
+    while True:
+        keymap.parse(matrix.raw_scan())
+        # Yield to the controller to allow things like the REPL to take over
+        time.sleep(0.001)
```
klardotsh: It's worth noting that this line absolutely does not work. Oops.

PR #8, opened 2018-09-03
Teensy support

kdb424:
Added microcontroller support. Keyboard support coming later in another PR.

### DIFF
--- a/entrypoints/teensy31.py
+++ b/entrypoints/teensy31.py
@@ -1,0 +1,10 @@
+import sys
+
+from kmk_keyboard_user import main
+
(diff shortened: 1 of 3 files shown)


entrypoints/teensy31.py
```diff
@@ -0,0 +1,10 @@
+import sys
+
+from kmk_keyboard_user import main
+
+if __name__ == '__main__':
+    try:
+        main()
+    except Exception as e:
+        sys.print_exception(e)
```
klardotsh: This is actually the default behaviour if you just let the exception go completely unhandled, but this is still fine

PR #10, opened 2018-09-15
Basic CircleCI setup: Lint the codebase on all inbound PRs

klardotsh:
This will fail and block a merge on a PR if the linter fails (checks for all the normal Flake8 stuff, plus verifying that `[]{}()` literals are used instead of, say, `dict(...)`, plus verifying that all imports are sorted correctly (which can be fixed locally automatically with `pipenv run isort myfile.py`)

I'll be configuring GitHub to require the lint step for inbound PRs to be mergeable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,35 @@
+version: 2
+jobs:
+  lint:
+    docker:
+      - image: 'python:3.7'
+
+    environment:
+        KMK_TEST: 1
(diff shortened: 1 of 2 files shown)


klardotsh:
You can also click those inline Xes and checkmarks on this page to see build logs! Circle is... decent. It's finnicky in bigger orgs, but generally good enough that I'd use it here rather than setting up something myself.

PR #18, opened 2018-09-17
CircleCI: Ensure all boards can at least compile a hex on each commit

klardotsh:


.gitmodules
```diff
@@ -1,14 +1,11 @@
 [submodule "circuitpython"]
 	path = vendor/circuitpython
 	url = https://github.com/adafruit/circuitpython.git
-	branch = "9b98ad779468676c3d5f1efdc06b454aaed7c407"
```
klardotsh: These lines conflict with how submodules work, because I'm dumb.

klardotsh:
Repo settings updated to require these checks to pass on each PR to `master`

PR #28, opened 2018-09-25
Topic grave escape

kdb424:
Well.... Here ya go.

### DIFF
--- a/kmk/common/internal_keycodes.py
+++ b/kmk/common/internal_keycodes.py
@@ -16,13 +16,44 @@ def process_internal_key_event(state, action, changed_key, logger=None):
         return tg(state, action, changed_key, logger=logger)
     elif changed_key.code == Keycodes.Layers._KC_TO:
         return to(state, action, changed_key, logger=logger)
+    elif changed_key == Keycodes.KMK.KC_GESC:
+        return grave_escape(action, state, logger=logger)
     else:
         return state
 
 
-def tilde(state, action, changed_key, logger):
-    # TODO Actually process keycodes
-    return state
+def grave_escape(action, state, logger):
+    if action['type'] == KEY_DOWN_EVENT:
+        for key in state.keys_pressed:
+            if key in {Keycodes.Modifiers.KC_LSHIFT, Keycodes.Modifiers.KC_RSHIFT}:
+                # if Shift is held, return KC_GRAVE which will become KC_TILDE on OS level
+                return state.update(
+                    keys_pressed=(
+                        state.keys_pressed | {Keycodes.Common.KC_GRAVE}
+                    ),
+                )
+            elif key in {Keycodes.Modifiers.KC_LGUI, Keycodes.Modifiers.KC_RGUI}:
+                # if GUI is held, return KC_GRAVE
+                return state.update(
+                    keys_pressed=(
+                        state.keys_pressed | {Keycodes.Common.KC_GRAVE}
+                    ),
+                )
+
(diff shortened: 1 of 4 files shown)


kmk/common/internal_keycodes.py
```diff
@@ -16,13 +16,54 @@ def process_internal_key_event(state, action, changed_key, logger=None):
         return tg(state, action, changed_key, logger=logger)
     elif changed_key.code == Keycodes.Layers._KC_TO:
         return to(state, action, changed_key, logger=logger)
+    elif changed_key.code == Keycodes.KMK.KC_GESC.code:
```
klardotsh: We can get away with just checking `changed_key == KC_GESC` here, don't REALLY have to check the code (since `KC_GESC` is a constant, whereas things like `KC_MO` are generated objects and required that `_MO` code hackaround)

kmk/common/internal_keycodes.py
```diff
@@ -16,13 +16,54 @@ def process_internal_key_event(state, action, changed_key, logger=None):
...
         return state
 
 
-def tilde(state, action, changed_key, logger):
-    # TODO Actually process keycodes
-    return state
+def grave_escape(action, state, logger):
+    if action['type'] == KEY_DOWN_EVENT:
+        for key in state.keys_pressed:
+            if (key == Keycodes.Modifiers.KC_LSHIFT or
```
klardotsh: This and the `elif` block can be condensed into a single `set` object with an `in` check, but this technically still works

kmk/common/internal_keycodes.py
```diff
@@ -16,13 +16,54 @@ def process_internal_key_event(state, action, changed_key, logger=None):
...
+                # else return KC_ESC
+                newstate = state.update(
+                    keys_pressed=(
+                        state.keys_pressed | {Keycodes.CodeKeys.CODE_ESC}
+                    ),
+                )
+                return newstate
+
+    elif action['type'] == KEY_UP_EVENT:
+        newstate = state.update(
```
klardotsh: Can flatten this into a single `update` call by checking for `if key not in (CODE_ESC, KC_GRAVE)`, but this technically still works. Turns an `O(n^2)` operation into a `2O(n^2)`, though, which may matter on weaker chips (but still... probably not).

kmk/common/keycodes.py
```diff
@@ -406,6 +406,9 @@ class ShiftedKeycodes(KeycodeCategory):
         KC_RIGHT_ANGLE_BRACKET = KC_RABK = KC_GT = Keycode(55, (CODE_LSHIFT,))
         KC_QUESTION = KC_QUES = Keycode(56, (CODE_LSHIFT,))
 
+    class CodeKeys(KeycodeCategory):
+        CODE_ESCAPE = CODE_ESC = Keycode(41)
```
klardotsh: Why is this needed over just checking for `KC_ESC` in the above loops?

klardotsh:
I guess while we're at it let's squash this thing down, this is a ton of commits, only a few of which were actually relevant (and for history's sake we are still fine with just one giant commit for this feature, given the relatively small diff)

kmk/common/internal_keycodes.py
kdb424: I did in fact have to check .code here. When I did not, it did not match, and was actually bypassed. I manually triggered this function during testing until I sorted this out. Not sure why.

kdb424:
This is great. Thanks.

PR #29, opened 2018-09-26
Makefile cleanup and useful utils

kdb424:


util/bootloader.py
```diff
@@ -0,0 +1,8 @@
+try:
+    import machine
+    machine.bootloader()
+
+except ImportError:
+    import microcontroller
+    microcontroller.on_next_reset(microcontroller.RunMode.BOOTLOADER)
```
klardotsh: oh shit, this is way cool. I assume this is the way CircuitPython expects things to be? I almost like it better than `machine.bootloader()`.

Notes on this and the reset file: this will actually trigger the reboot/reset immediately upon file import. Wrap both of these into functions, merge them both into a single file (no need for two files for this - also we have `kmk/common/util.py` which may be a good fit for this), and I'm good with shipping. Thanks for doing this!

The reason I want these merged into actual KMK source tree is because we can then change `KC_RESET` (and optionally add `KC_SOFT_RESET`) to point at these, reducing duplicated code.

To retain `ampy run` ability, we can keep a `util/reset.py` (though I think I might prefer something like `bin/force_keyboard_reset.py`) which imports the correct functions from `kmk.common.util` and runs them `if __name__ == '__main__'`

klardotsh: I guess I'll retract the `bin/` move request, because `bin/*` should be runnable on the host PC, which `force_keyboard_reset` would not be. So if we move the functionality itself to within kmk and then call it from the `util/reset.py` file I'm okay with just that much change.

kdb424: I assumed since it's so simple of what is in those files, I would simply leave these and add correct things later in the firmware. For example, other things that I'd like to add to the util folder are things like linux_deps.sh which will detect the distro, and install any deps needed on each distro, usb drivers for windows, ect. Keeping them this simple allows easy interaction from the development point of when working on code, simply importing them if the dev wants to push them to the root directory. LTDR, having this documented and in a very easy to repurpose spot was my design goal and why they ended up here.

PR #31, opened 2018-10-01
Macros: Support basic sequences of keys (either raw or specified by a string) and cross-platform Unicode sequences (ex. (ノಠ痊ಠ)ノ彡┻━┻)

klardotsh:


klardotsh:
I mean, this isn't advertised as "the unicode solution", this is advertised as an ibus-specific macro that you can opt into. I have no idea what all the other OSes use for this stuff and no realistic way to test it even if I implemented.

klardotsh:
The best I can really do right now is hack in support for the options that don't suck from https://docs.qmk.fm/#/feature_unicode and hope to hell they actually work, but I have exactly zero Macs or Windows machines in the house, so testing those things is officially Your Problem™ or we'll just merge untested.

PR #36, opened 2018-10-03
Advanced layers support basically done.

kdb424:
We need to convert at least TT to tap dance, but holding the rest back for that is not something I would call needed.


kmk/common/internal_state.py
```diff
@@ -134,7 +150,7 @@ def kmk_reducer(state=None, action=None, logger=None):
     if action.type == NEW_MATRIX_EVENT:
         matrix_keys_pressed = {
             find_key_in_map(state, row, col)
-            for row, col in action.matrix
+            for row, col, in action.matrix
```
klardotsh: I would love to get to the bottom of why this is needed

klardotsh:
Apparently this needs a rebase and/or merge of master

PR #45, opened 2018-10-07
Add support for Adafruit Feather+ItsyBitsy M4 Express boards through CircuitPython

klardotsh:
The bulk of this work was just "port KMK to CircuitPython and deal with the new abilities and restrictions thereof". We can't fully flash these devices (or at least... I can't fully flash mine) through just the Makefile, so `flash-feather-m4-express` will build a [uf2](https://github.com/Microsoft/uf2) file and provide the user with instructions to flash it.

On my Arch Linux system with no auto-mounting of USB (read: no desktop environment or other background utilities managing my USB), I was able to fully build and flash an ItsyBitsy with the following:

```sh
# double tap reset button on device (once KMK is flashed the `KC_RESET` button also works for dropping to bootloader, assuming firmware has not crashed)
sudo mount -o uid=1000,gid=1000 /dev/sdf mnt
cp build/circuitpython/ports/atmel-samd/build-itsybitsy_m4_express/firmware.uf2 mnt/
# Red light should start flashing and flickering. Wait for it to turn off before continuing
sudo umount -f mnt
```

This branch integrates `internal_keycodes` with `internal_state` to avoid a `RuntimeError` in CircuitPython where the max recursion depth (seemingly 16 on these SAMD51 boards) is reached during the import process (`from kmk_keyboard_user import main` from REPL). This is the same bug we saw on the NRF MDK port, however an actual exception is thrown on these boards, rather than hardlocking everything like the NRF. That's nice. In general this change is a warning to us to watch our import depth going forward, lest we break non-Pyboards.

This branch adds a utility file (`kmk.common.pins`) which exposes two variables to the user: `get_pin(pin_str)` and `Pin`. These allow cross-implementation pin abstraction, so that the user doesn't have to import anything from Micro/CircuitPython directly - they can address pins on their board as, for example, `get_pin('D20')` or simply `Pin.D20`. I've updated all keymaps in this branch to use this new syntax, and it's awesome if I say so myself.

This branch cleans up the Makefile targets a bit and also banishes the remnants of Teensy support (or lack thereof) entirely, from the Makefile as well as the CircleCI config.


klardotsh:
Tests are failing because of the Docker environment that the test builds run in. I'll have to update the Circle config to use images based off a newer Ubuntu/Debian (or move to Alpine, or whatever) which provide the newer GCC cross compiler that CircuitPython expects.

klardotsh:
CI _finally_ passes, merging. Opened #46 to track a quality of life addition we'll want going forward, discovered by this hours-long process of unbreaking CI.

PR #51, opened 2018-10-10
Correct SAMD chip number

tannewt:


klardotsh:
Heh... oops. Crossed brain wires at some point (SAMD21, SAMD52)

klardotsh:
Something's up with Circle's config here - going to force merge since it's docs only. Thanks again!

Issue #46, opened 2018-10-08
Create a kmkfw Docker image


klardotsh:
This will not only allow for significantly easier user flashing (hopefully...), but will also cut down CircleCI runtimes, which are currently bonkers due to all the bootstrapping and system dependencies we have.

klardotsh:
A potentially useful enhancement for this task would be to allow Circle to build the base image as the first step in each build workflow, then use that base image across all future steps in the workflow. This effectively would ensure our Docker base image is ALWAYS up to date with whatever Dockerfile is in the repo, regardless of whether I forget to run the make task locally.

klardotsh:
This is now almost entirely unnecessary given we target upstream builds (UF2-enabled) of CircuitPython and don't freeze modules. At this point our only use for Docker is to run CI, which is not really the spirit of what this ticket was getting at. Closing.

PR #60, opened 2018-10-12
Some backports from work on #46 to greatly speed up Circle builds

klardotsh:
This backports from topic-docker the base image, which we can now use in
Circle to MASSIVELY speed up our CI situation. This at least caches all the Linux-level dependencies (Python, GCC, etc. etc.) and should shave 3-5 minutes off each build stage.

It also merges the lint and test stages for simplicity's sake.

The base image can currently only be pushed by @klardotsh as nobody else
has access to the Docker Hub org, but that can change in the future.

Not finishing up `topic-docker` or #46 just yet because I need to do some Makefile foolery to ensure we are not redistributing the Nordic BLE stack within our Docker images, which I'm _pretty_ sure is a violation of their licensing terms (thus why MicroPython and CircuitPython both call out to a bash script to pull those firmware blobs at build time).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,134 +1,66 @@
 version: 2
 jobs:
-  lint:
-    docker:
-      - image: 'ubuntu:bionic'
-
-    environment:
-        KMK_TEST: 1
-        PIPENV_VENV_IN_PROJECT: 1
-
-    steps:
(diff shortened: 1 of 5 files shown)


klardotsh:
Looks like each stage takes about 1m30s to complete now, which is fantastic compared to the 5-7 we were at.

PR #62, opened 2018-10-12
Basic Docs

kdb424:


docs/keycodes.md
```diff
@@ -0,0 +1,299 @@
...
+
+
+## [Internal Keycodes]
+
+|Key                    |Description                                                          |
+|-----------------------|---------------------------------------------------------------------|
+|`KC.RESET`             |Put the keyboard into DFU mode for flashing                          |
+|`KC.DEBUG`             |Toggle debug mode                                                    |
+|`KC.GESC`              |Escape when tapped, <code>&#96;</code> when pressed with Shift or GUI|
+|`KC.LEAD`              |The [Leader key]                                                     |
```
klardotsh: This link goes nowhere

docs/keycodes.md
```diff
@@ -0,0 +1,299 @@
...
+|`KC.RIGHT_ANGLE_BRACKET`|`KC.RABK`, `KC.GT` |`>`                |
+|`KC.QUESTION`           |`KC.QUES`          |`?`                |
+
+
+## [Internal Keycodes]
+
+|Key                    |Description                                                          |
+|-----------------------|---------------------------------------------------------------------|
+|`KC.RESET`             |Put the keyboard into DFU mode for flashing                          |
+|`KC.DEBUG`             |Toggle debug mode                                                    |
```
klardotsh: Does this actually work?

docs/keycodes.md
```diff
@@ -0,0 +1,299 @@
...
+|`KC.GESC`              |Escape when tapped, <code>&#96;</code> when pressed with Shift or GUI|
+|`KC.LEAD`              |The [Leader key]                                                     |
+|`KC.UC_MODE_NOOP`      |Sets UnicodeMode to NOOP                                             |
+|`KC.UC_MODE_LINUX`     |Sets UnicodeMode to NOOP                                             |
```
klardotsh: `NOOP` copy-pasta where not appropriate

docs/keycodes.md
```diff
@@ -0,0 +1,299 @@
...
+|`KC.MACRO_SLEEP_MS(ms)`|Sleeps in a macro. Check MACROS for more information.                |
+
+
+## [Layer Switching]
+
+|Key         |Description                                                                  |
+|-----------------|------------------------------------------------------------------------|
+|`KC.DF(layer)`      |Switches the default layer                                           |
+|`KC.MO(layer)`      |Momentarily activates layer, switches off when you let go            |
+|`KC.LM(layer, mod)` |As `MO(layer)` but with `mod` active                                 |
```
klardotsh: Do we actually have this support? I have no idea.

docs/keycodes.md
```diff
@@ -0,0 +1,299 @@
...
+|`KC.LSFT(kc)`|Hold Left Shift and press `kc`                      |
+|`KC.LALT(kc)`|Hold Left Alt and press `kc`                        |
+|`KC.LGUI(kc)`|Hold Left GUI and press `kc`                        |
+|`KC.RCTL(kc)`|Hold Right Control and press `kc`                   |
+|`KC.RSFT(kc)`|Hold Right Shift and press `kc`                     |
+|`KC.RALT(kc)`|Hold Right Alt and press `kc`                       |
+|`KC.RGUI(kc)`|Hold Right GUI and press `kc`                       |
+|`KC.HYPR(kc)`|Hold Left Control, Shift, Alt and GUI and press `kc`|
+|`KC.MEH(kc)` |Hold Left Control, Shift and Alt and press `kc`     |
+|`KC.LCAG(kc)`|Hold Left Control, Alt and GUI and press `kc`       |
```
klardotsh: We do not support this or anything below it within this section

docs/keymap.md
```diff
@@ -0,0 +1,51 @@
+# Keymap
+The keymap is a very simple set of instruction for your keyboard.
+An example keymap for a 3x3 macro pad
```
klardotsh: Nit: you wrote a 4x3

docs/keymap.md
```diff
@@ -0,0 +1,51 @@
...
+        [KC.GESC,  KC.A, KC.B],
+        [KC.C,     KC.D, KC.E],
+        [KC.F,     KC.G, KC.H],
+        [KC.MO(1), KC.I, KC.J],
+    ],
+    [
+        [KC.TRNS, KC.K, KC.RESET],
+        [KC.L,    KC.M, KC.N],
+        [KC.O,    KC.P, KC.Q],
+        [KC.R,    KC.S, KC.T],
```
klardotsh: `TRNS` should be down here, not at the top of this column

docs/keymap.md
```diff
@@ -0,0 +1,51 @@
...
+
+# Handwire
+When hand wiring custom keyboards, you will have to set the board up so KMK
+knows where things are. This is the basic SETUP.
+
+```python
+from kmk.consts import DiodeOrientation
+from kmk.keycodes import KC
+from kmk.pins import Pin as P
+from kmk.entrypoints.handwire.feather_m4_express import main
```
klardotsh: let's update this to the latest state, `kmk.entrypoints.circuitpython_samd51`

docs/keymap.md
```diff
@@ -0,0 +1,51 @@
...
+cols = (P.D11, P.D10, P.D9)
+rows = (P.A2, P.A3, P.A4, P.A5)
+
+diode_orientation = DiodeOrientation.COLUMNS
+```
+
+The main things that need set up are:
+1. Your microcontroller will need to be set here.
+```python
+from kmk.entrypoints.handwire.feather_m4_express import main
```
klardotsh: ditto

docs/unicode.md
```diff
@@ -0,0 +1,68 @@
+# Macros And Unicode
+Macros are used for sending multiple keystrokes in a single action. This is useful for
+things like unicode input, sending strings of text, or other automation.
+
+## Basic Macros
+The most basic macro is send_string(). It can be used to send any typed keycode, including the return and tab key.
```
klardotsh: Maybe "any standard ASCII keycode" rather than "any typed keycode"?

docs/unicode.md
```diff
@@ -0,0 +1,68 @@
...
+WOW = send_string("Wow, KMK is awesome!")
+
+keymap = [...WOW,...]
+```
+# Unicode
+Before using unicode mode, you will need to set your platform. This can be done either of these ways.
+
+	unicode_mode = UnicodeModes.LINUX
+	OR
+	keymap = [...KC.UC_MODE_LINUX,...]
```
klardotsh: If they do this, they'll need to tap that keycode before unicode mode will work at all, this is absolutely worth noting.

It's also worth noting that using that key, they can change at runtime, allowing them to flip from a Linux to MacOS machine with no keyboard flashing, for example.

docs/unicode.md
```diff
@@ -0,0 +1,68 @@
...
+
+TABLE_FLIP = unicode_codepoint_sequence([
+	"28", "30ce", "ca0", "75ca","ca0", "29",
+	"30ce", "5f61", "253b", "2501", "253b",
+])
+
+keymap = [...TABLE_FLIP,...]
+```
+
+To send a simple unicode symbol
```
klardotsh: Very critical note: emojis like this are actually only supported on Linux, because ibus doesn't restrict codepoint length to 4 (it restricts to like 6 or 8 I think). MacOS and (I believe) Windows only support 4-character hex codepoints, which explicitly precludes emojis right now (but works for things like table flips)

klardotsh: I'd also love to show an example of an inline table flip here even though table flips were used above in the codepoint sequence.

Further, let's bump this section up above codepoints. Codepoints manually are such a fringe usecase I doubt they'll ever be used, but sending a unicode string is _awesome_ and something we should publicize.

docs/keycodes.md
kdb424: We don't have the server set up, so I don't know where to send it yet.

docs/keycodes.md
kdb424: I'm in the works, so, VERY soon.

docs/keycodes.md
kdb424: We support all layers. I handled those.

docs/keycodes.md
kdb424: Can we?

docs/keymap.md
kdb424: Oops.

docs/keycodes.md
klardotsh: ahhh, fair

docs/keycodes.md
klardotsh: cool, then leave it for now

docs/keycodes.md
klardotsh: open a ticket for them, you can base the implementations off MEH and HYPER in #65 if you get to it before I do

docs/unicode.md
kdb424: I would argue tho opposite. I have seen much more use with people asking for this to type other language characters such as german. In QMK, I've seen people tap dance the "u" key to "ü" on double tap for this reason. I am willing to change to the table flip however here, but that's regardless of order.

kdb424: Misunderstood. Fixing

klardotsh: right, but this send string implementation works for German characters just as well as it does for table flips, without having to manually convert to `0252` :)

let's do the demo with umlaut U and/or table flip, and maybe keep an emoji example for Linux folks. And put this section above the manual-codepoint section.

docs/keycodes.md
kdb424: Removed. Not worth the refactor.

docs/keycodes.md
kdb424: This is on the later list after tap dance as those fit as tap dance codes with a shortcode.

PR #82, opened 2018-10-26
Topic split keyboards

kdb424:
This is a solid start to #3 bringing a solid base for i2c and bluetooth to come later.

### DIFF
--- a/docs/split_keyboards.md
+++ b/docs/split_keyboards.md
@@ -4,7 +4,6 @@ UART is supported, though other modes will come later such as Bluetooth and i2c.
 
 Useful config options:
 ```python
-keyboard.split_type = "UART"  # Sets split mode to UART
 keyboard.split_flip = True  # If your boards are identical but one is flipped, this option is for you 
 keyboard.split_offsets = [6, 6, 6, 6]  # This is the how many keys are on each column on the "Master" half
 ```
@@ -34,5 +33,6 @@ from kmk_side import split_side
 # UART
 To enable uart it's as simple as adding this line, of course changing the pin
 ```python
+keyboard.split_type = "UART"
 keyboard.uart = keyboard.init_uart(tx=board.SCL)
 ```
--- a/docs/split_keyboards.md
+++ b/docs/split_keyboards.md
@@ -1,0 +1,38 @@
+# Split Keyboards
+Split keyboards are mostly the same as unsplit and very easy to adapt a keymap for. Currently
+UART is supported, though other modes will come later such as Bluetooth and i2c.
+
+Useful config options:
+```python
+keyboard.split_type = "UART"  # Sets split mode to UART
+keyboard.split_flip = True  # If your boards are identical but one is flipped, this option is for you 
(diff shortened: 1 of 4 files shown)


kmk/firmware.py
```diff
@@ -75,12 +83,83 @@ def _send_key(self, key):
...
         assert self.keymap, 'must define a keymap with at least one row'
         assert self.row_pins, 'no GPIO pins defined for matrix rows'
         assert self.col_pins, 'no GPIO pins defined for matrix columns'
         assert self.diode_orientation is not None, 'diode orientation must be defined'
 
+        if self.split_flip and not self._master_half():
+            self.col_pins = list(reversed(self.col_pins))
+
+        if self.split_side == "Left":
+            if supervisor.runtime.serial_connected:
```
klardotsh: This can be flattened into `self.split_master_left = supervisor.runtime.serial_connected`, and the block of these below can be `self.split_master_left = not supervisor.runtime.serial_connected`

kmk/firmware.py
```diff
@@ -75,12 +83,83 @@ def _send_key(self, key):
...
+    def _receive_from_slave(self):
+        if self.uart is not None and self.uart.in_waiting > 0:
+            update = bytearray(self.uart.read(3))
+            if self.split_master_left:
+                update[1] -= self.split_offsets[update[0]]
+            return update
+
+        return None
+
+    def _master_half(self):
```
klardotsh: Cache this to a variable rather than using a function here, maybe? Saves some cycles.

kmk/firmware.py
kdb424: It was left this way to leave room for bluetooth in the future.

kmk/firmware.py
kdb424: Same as before, bluetooth functionality will be different, and therefor this was left open ended for changes.

docs/split_keyboards.md
```diff
@@ -0,0 +1,38 @@
...
+
+and then in your keymap, add the line
+```python
+from kmk_side import split_side
+```
+
+# UART
+To enable uart it's as simple as adding this line, of course changing the pin
+```python
+keyboard.uart = keyboard.init_uart(tx=board.SCL)
```
klardotsh: We should be recommending they use the `P.` abstraction from `from kmk.pins import Pin as P`, never the raw pins in `board`

klardotsh:
This is good to go modulo that one comment about consistent pin stuff

docs/split_keyboards.md
kdb424: Whoops. Totally missed that as I moved it from another file originally.

kdb424: File "kmk/firmware.py", line 136, in init_uart
TypeError: Expected a Pin
This is actually not possible unfortunately

kmk/firmware.py
kdb424: I retract this statement. It's being resolved.

kmk/firmware.py
kdb424: Caching now.

PR #83, opened 2018-11-01
added hardware directory with an Adafruit Itsy Bitsy to the Pro Micro…

siddacious:
… footprint. This is just a first draft to get some feedback on. I'm happy to add whatever silks or tweaks you might want.


klardotsh:
Hey @siddacious, thanks for this! Finally back from a weekend trip out of town - I'll take a look at these this evening.

Had to look into what Eagle was - for sake of documentation (in case someone later stumbles across this PR), it looks like Eagle is proprietary subscriptionware from Autodesk which has a limited free version. [A project exists on Github](https://github.com/lachlanA/eagle-to-kicad) to convert Eagle files to [KiCAD](http://kicad-pcb.org/) format, which is an open-source PCB CAD tool. I may also poke at that tool to see if these Eagle files open correctly there.

klardotsh:
@kdb424 Assigning you as final arbiter of approval on this even if I do a review of my own, since you "own" the hardware design and support end of these types of things and it was your idea originally. See also the above comment on ways you'll be able to open and review these.

siddacious:
I do intend to eventually start using KiCAD for some things but I'm too busy at the moment. The current 5.0 version of KiCAD should have Eagle integration build in.

In spite of the few issues Eagle may have, it is widely used and much more user friendly than KiCAD; it's near close to the lingua franca for ECAD software in the maker/DIY world. With that in mind I would suggest that Eagle files be distributed as the primary format and KiCAD as the secondary.

klardotsh:
I trust your judgement far more than my cursory glance on that - I'm not opposed to proprietary software that's the best tool for the job. For sake of folks who _do_ care and are of the "never ever ever am I installing proprietary software" type, I may make a README in this folder that alludes to whatever I come across in testing tonight.

I'll submit my review based solely on opening the files in actual Eagle though - good to know it's the standard (again, I'm not a hardware guy usually! heh)

Especially if KiCAD can open Eagle files directly, I guess we don't even need to ship KiCAD files as a secondary (keeping "build assets" in sync in a Git repo is a problem I tend to hate solving, and if it's not necessary, great, I'll skip it)

klardotsh:
Still deferring to @kdb424 for final approval but this looks _suuuuuper_ clean, thanks! Clarification question for you - I see a bunch of wire objects (colored white/gray in my view, at least) on layer `49 - Reference`, mostly leading off the bottom five pins of the ItsyBitsy. I assume those are just for alignment / future use (if we ever found something to do with them in a future revision? Maybe compatibility with the [Elite C Pro Micro replacement](https://keeb.io/collections/frontpage/products/elite-c-usb-c-pro-micro-replacement-arduino-compatible-atmega32u4)?), and wouldn't end up in a print if we sent this off to DirtyPCB/wherever?

Also can confirm these files are a no-go in KiCAD - I get a segfault / core dump immediately upon it opening the actual board designer (the schematic window appeared to show up). This could be a system config issue on my side, but I'm not going to dive too far into it.

siddacious:
The lines/wires on reference are indeed just for reference in making the vias look nice and that layer isn't used by any fab house that I know of.

[This page](https://docs.oshpark.com/design-tools/eagle/eagle-layers-explained/) has a good explanation of the layers in Eagle.

I'm bringing up a board tonight but I'll take a pass at a `README.md` and `LICENSE` when I have a minute.

kdb424:
Looks great! Once licence and readme are good to go, I'll call this an approval. Thanks a ton!

kdb424:
Approved, and updating!

PR #94, opened 2018-12-05
Update README.md

kdb424:
Small typo and added better readability for translating to code, which is why most users will be here (adding a keyboard)


klardotsh:
Needs a rebase but otherwise looks good

klardotsh:
cc @siddacious, we can't actually request reviews from folks outside the org so this is mostly an FYI (I'll fix that issue by just... inviting you to the org)

PR #105, opened 2018-12-29
Congressional Bill 122918 Forgot To Sleep Edition: Refactor everything about how key definitions work

klardotsh:
There's a TON here, so here goes:

- Fully rewrite kmk.keycodes and banish the worst offenses of OOP-abuse in the codebase. Keys are now defined using a spattering of (mostly nested) helper functions ultimately delegating to the KC AttrDict (which is now the single point of truth for both userspace and internal uses) and Keycode class. There's still some cleanup to be done in this file, but I didn't want to rock the boat too much in one night.

- Handle basically everything that can happen as a result of a user pressing a physical key through the same execution flow (in simpler speak, everything is a "macro" in a sense), regardless of whether the handler was written by the Core Dev team, some random contributor, or a non-contributor making their own macros in their keymap (think backlight, LCDs, whatever). There's an example in my Planck keymap, ft. Shrek, of how a user would write such a thing. The function signature for handlers includes basically everything a user would ever want to know about the state of the application, without them needing to import things from far-flung lands, but most are ignorable through args/kwargs globbing.

- In line with the above, move a bunch of functions around while leaving their functionality 90%+ untouched. All KMK-provided handlers have been moved to one of three files: a file of layer handlers, a file of "everything else that used to live in InternalState" handlers, and a file of sequence handlers, containing everything that used to live under kmk.macros.simple and kmk.macros.unicode. Some functionality hasn't been moved out of InternalState because it'd be way too much heavy lifting - leader and tap dance. Those will one day work through some sort of "key swallowing" API that I haven't thought through yet, so instead, I'm for now taking the stance that things as advanced as those functionalities will still live in InternalState, triggered by a macro-style key. This will change one day, probably.

- **Lifecycle change**: Macros now fire inline inline in InternalState rather than after state calculation/settlement in Firmware. Since we're now committed to "everything speaks the same execution language", I wanted to discard of special cases, and the old macro system (through which all sequence macros ran) was one.

- Remove vestigial and non-working (for several months now) references to rotary encoders. At this point it's not even remotely worth trying to catch them up with two rounds of massive internal refactors.

- *Quality of life update*: Leader key sequences can now be defined with strings as the dictionary keys. No more glds spam.

- **Add a (required) boot.py file** - we're now at the point where the recursion limits in CircuitPython are crippling. I finally found a value (1024 greater than the default of 4096) that allows KMK to boot and also handle the crazy trees of keycode generation and macro execution KMK requires. Makefile will install it for you when flashing "everything", or through an individual target if desired.


klardotsh:
I've tested this on the Planck - so far nothing seems broken. Layers work, shifted keys work, unicode and simple sequences (WPM, whatever) work... I haven't tested tap dance yet, but I have zero reason to suspect it would have regressed given its migration was handled the same way as leader mode's migration.

I need to HEAVILY rewrite docs, but before getting to that, I need to sleep. Wanted to get the ball rolling for code review, which can happen in parallel to writing 900,000 words.

klardotsh:
I haven't updated any keymaps other than my Planck. I may do that after sleep, or may say it's Not My Problem (TM). Probably the former.

kdb424:
I'll be testing this over the coming few days, and reporting any issues, likely following new docs due to timing. Expect on Jan 1st a review pending no issues for daily use.

kdb424:
Code seems stable at this point. I'll have to put more time into testing adding keycodes as a user (testing underglow support as a user controlled thing), but looks generally stable at least for such a large refactor.

kdb424:
Approved. Tested quite a bit of it, and looks great over all. Much easier to work with when adding new features as well.

PR #119, opened 2019-03-09
Literally this is all that is needed to restore functionality (GESC)

### LINKED ISSUE #114: Fix gesc yet again.
Seems like GESC is now partially working. If you don't let go of super, it will not allow you to type it again. It  discards all keys pressed as intended, but does not re add it to the state.

kdb424:
QMK allows configuration which we don't currently have any of to allow unbreaking of OS keybinds, but without any workarounds, this is literally what it is with QMK. We don't need to throw away MOD4/Super/Win Resolves #114

### DIFF
--- a/kmk/handlers/stock.py
+++ b/kmk/handlers/stock.py
@@ -51,9 +51,6 @@ def gesc_pressed(key, state, KC, *args, **kwargs):
     GESC_TRIGGERS = {KC.LSHIFT, KC.RSHIFT, KC.LGUI, KC.RGUI}
 
     if GESC_TRIGGERS.intersection(state.keys_pressed):
-        # First, release GUI if already pressed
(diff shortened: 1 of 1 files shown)


klardotsh:
Fairly certain this reverts us to the state we previously had, which broke GESC entirely on my system (and I believe yours) because the OS was interpreting this as `Super+Tilde`, which would normally be an OS/WM-level keybind?

Still not a fan of this key in general. Such a pain code-side when `MO(1) -> KC.GRV` solves the issue in the same number of keypresses user-side.

kdb424:
I honestly recommend that we remove this keybind all together. It's a convenience bind, and has always been hacky. That's how QMK implemented it,and you didn't seem to like it there either. I had absolutely no issues with QMK's implementation, you did.

PR #120, opened 2019-03-09
Adds delete/backspace key. Resolves #93

### LINKED ISSUE #93: Add backspace/Delete key combo
Through MOD4, similar to GESC.

kdb424:


klardotsh:
I have the same concern here with colliding with the OS that I had in #119 - whatever solution we end up with, let's be sure to test it in a few different environments and ensure it works somewhat sanely.

kdb424:
These are both "hacky" They are not guaranteed to work in every situation, not should be expected to. Convenience binds are there for those it works for. It obviously isn't for everyone, or every system.

PR #124, opened 2019-03-16
Literally hurting performance now. Needs removed

kdb424:
This has been tested to HURT performance quite a lot. While animations are running, I was hitting a border of performance that I could not roll keys. With this one line change, almost all of it is solved. I recommend testing this branch before accepting, but I don't think that we still need this. If we do, than we will have to change limits as without this, performance can get very bad in some situations even with "plenty" of ram free.

### DIFF
--- a/kmk/firmware.py
+++ b/kmk/firmware.py
@@ -23,7 +23,6 @@ import kmk.types  # isort:skip
 import kmk.util  # isort:skip
 
 import busio  # isort:skip
-import gc  # isort:skip
 
 import supervisor  # isort:skip
 from kmk.consts import LeaderMode, UnicodeMode  # isort:skip
@@ -223,5 +222,3 @@ class Firmware:
 
             if self.debug_enabled and state_changed:
                 print('New State: {}'.format(self._state._to_dict()))
-
-            gc.collect()


klardotsh:
Have we tested heavy facerolling/etc with this? My worry is we'll be back in the same boat we were in when the GC was added - when the GC _does_ have to kick in, it would have to clear so many bytes that the massive "stop the world" event would degrade performance for a long period (but only every so often).

That said - I'm fine with this in principle. That workaround was added when we still targeted PyBoards which have about as much RAM as my microwave, give or take. The SAMD51/NRF52840 chips have enough RAM that GC probably won't fire as often, and can fire in "downtime" assuming CircuitPython works that way (which I'm not sure it does?)

kdb424:
On my Nyquist, I see absolutely no change in performance what so ever no matter the situation. On my luddite, it seems to be costing more memory than the nyquist, which I assume is due to the larger matrix scan, being an 8x8 matrix and not a 6x5. This is where I see most memory change. Normally on the luddite (60% non split, 8x8) all is well on mainline, but on the LED branch, animations cost a slight increase in memory "waste" each cycle. Due to that buildup of temp variables, the GC every cycle kicks into the "end of the world" mode. GC does in fact fire in the "downtime" as temp variables are constantly used in animations and trigger the clean. If I quickly roll the home row, with the GC every cycle, and the animation on, I get literally 1 or 2 characters MAX out of the board. Without GC I normally will drop 1 or 2 characters. I believe that there is more performance to be had, and I need to profile animations (honestly it's mostly the HSV to RGB conversion) to see if I can use less memory, but animations will always cause more memory to be used, and is to be expected that it will GC occasionally as well.

PR #128, opened 2019-05-13
Resolves #121: Use flattened keymaps, which can visually represent the logical layout, rather than the physical wiring

### LINKED ISSUE #121: Convert keymap from list of lists, to flat list.
This is needed for "non square" pcb's or ones that don't align the colums due to pin limitation. For example, the luddite is a "square" in software, each row is 8 columns, but that makes the keymap a lot harder to read. We should consider making this change during boot and forming the rows and columns for internal used based on pins, and not a user defined list of lists.

Example of why this is not good.
https://github.com/KMKfw/kmk_firmware/blob/d5272ced192018137246ebe700640d87afe0638c/user_keymaps/kdb424/luddite.py#L49-L59

klardotsh:
@kdb424 This fixes the hair-pull-out-ing the Iris and your grid-but-not keyboard cause. I did not fix said keyboard or layout for you, but I updated basically all other keymaps in the tree (including your Levinson and Planck) to use this new system.

For a great example of how this works for "weird" boards, see the firmware for the Iris. For boards that only used to require `swap_indicies`, I replicated that behavior in the `klaranck` tree (arguably that could be pulled out to `kmk.util` as `swap_indicies` boards are probably not uncommon in the wild?))


kmk/internal_state.py
```diff
@@ -47,10 +47,26 @@ def _to_dict(self):
         return ret
 
     def _find_key_in_map(self, row, col):
+        ic = intify_coordinate(row, col)
+
+        try:
+            idx = self.config.coord_mapping.index(ic)
```
klardotsh: This is technically super inefficient (introduces an `O(n)` lookup where we once had an `O(1)`, but `coord_mapping` will need to change to some form of map/dict to bring this back to an `O(1)`. This can likely be done as a compile-down at firmware boot time, much like what we do with leader key sequences.

kdb424:
This one will take a bit longer to review as I need to test it with boards in a different branch to ensure it works as intended, but am looking at.

elvis-epx:
It seems that user_keymaps/kdb424/* were left behind.

kdb424:
@elvis-epx I'll be updating these separately as I have a branch I live on for the moment. Thanks for the heads up.

PR #129, opened 2019-05-17
Make matrix scanner class configurable

elvis-epx:
Allows for future matrix classes (e.g.  MCP23017-based matrix, PS/2 converter)


klardotsh:
Thanks @elvis-epx !

klardotsh:
Had to merge `master` into your branch to allow merges (branches have to be up to date to merge around here - which may or may not actually be a good policy now that we have multiple contributors....), so forgive the branch clutter.

PR #131, opened 2019-05-31
Add KC_APP ("Win menu") key

elvis-epx:


klardotsh:
Oh, yep. Good catch, thanks!

PR #139, opened 2019-07-25
Enforce the Black autoformatter and code style

klardotsh:
I care less about what the actual style is and more that it's consistent. Black is now an official Python Software Foundation project and has mostly-sane rules. I overrode one that, [per the upstream ticket](https://github.com/psf/black/issues/594), it appears many or most major projects disagree with and also override, so I don't feel bad.

Surprisingly, we already followed Black almost to a T, and very little code was reformatted. Included is a `Makefile` target to let Black automatically fix your commit before it pushes (you could run this as a pre-commit hook in Git if you want)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,11 @@ docker-base-deploy: docker-base
 devdeps: .devdeps
 
 lint: devdeps
+	@$(PIPENV) run black --check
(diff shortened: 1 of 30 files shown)


klardotsh:
typing this response from this branch flashed over rsync

PR #151, opened 2020-01-01
add handwire ghirkin

kdb424:
Hoping to use this board to add BLE HID eventually.

### DIFF
--- a/kmk/boards/handwire_ghirkin.py
+++ b/kmk/boards/handwire_ghirkin.py
@@ -1,0 +1,25 @@
+import board
+
+from kmk.kmk_keyboard import KMKKeyboard as _KMKKeyboard
+from kmk.matrix import DiodeOrientation
+
+
+class KMKKeyboard(_KMKKeyboard):
+    col_pins = (
+        board.SDA,
+        board.A0,
+        board.A1,
+        board.A3,
+        board.A2,
+        board.A4
+    )
+    row_pins = (
+        board.MOSI,
+        board.MISO,
+        board.RX,
+        board.TX,
+        board.D2
+    )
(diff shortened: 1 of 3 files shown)


kdb424:
./kmk/boards/handwire_ghirkin.py:8:17: BLK100 Black would make changes.
I have no idea what changes it would make, but yet it complains.

klardotsh:
It probably thinks it will cram those tuples onto a single line... stand by.

kmk/boards/handwire_ghirkin.py
```diff
@@ -0,0 +1,12 @@
+import board
+
+from kmk.kmk_keyboard import KMKKeyboard as _KMKKeyboard
+from kmk.matrix import DiodeOrientation
+
+
+class KMKKeyboard(_KMKKeyboard):
+    col_pins = (board.SDA, board.A0, board.A1, board.A3, board.A2, board.A4)
+    row_pins = (board.MOSI, board.MISO, board.RX, board.TX, board.D2)
```
klardotsh: For hand-wire boards, `col_pins` and `row_pins` should live in the user keymap instead of a global definition, yeah?

Heck, this whole file probably shouldn't exist if we're roughly keeping with historical trend (the `klarank` should probably also not be in `boards` but whatever). The `user_keymap` should likely be the only file for handwires.

kdb424: I seemed to have some issues putting this in a keymap file, but it could have been the constant corruption on my microcontroller. I separated the lines based on another board I found, so oops.

klardotsh:
Can we fix or close this?

PR #152, opened 2020-01-05
Add the ability to use seesaw I/O extender for scanning

jepler:
This is "tested" by manually creating contacts between the pins on an Itsy Bitsy M4 and the samd09 breakout.  I'm entering it in Draft status because it's a proof of concept and I don't really know how it should work, organizationally speaking, to get the seesaw instance across the layers it needs to span.

At one point I stated on chat that only pull UP was supported, but that has now been fixed in current adafruit_seesaw and pull DOWNs are available.  However, my test was arranged so that the seesaw pins were driving pins and the local pins were input pins.  This is probably preferable for performance, since it means fewer I2C transactions.

- [ ] "plumb" the seesaw object in a palatable way
- [ ] document the functionality appropriately
- [ ] benchmark that performance is okay
- [ ] provide a tested sample configuration that is properly organized

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -191,7 +191,7 @@ class KMKKeyboard:
         else:
             return busio.UART(tx=pin, rx=None, timeout=timeout)
 
-    def go(self, hid_type=HIDModes.USB):
+    def go(self, hid_type=HIDModes.USB, seesaw=None):
         assert self.keymap, 'must define a keymap with at least one row'
         assert self.row_pins, 'no GPIO pins defined for matrix rows'
         assert self.col_pins, 'no GPIO pins defined for matrix columns'
@@ -270,6 +270,7 @@ class KMKKeyboard:
             rows=self.row_pins,
             diode_orientation=self.diode_orientation,
             rollover_cols_every_rows=getattr(self, 'rollover_cols_every_rows', None),
+            seesaw=seesaw,
         )
 
         # Compile string leader sequences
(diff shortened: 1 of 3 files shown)


klardotsh:
Thanks for this! I should have time this or tomorrow evening to review

jepler:
@klardotsh any feedback?  It turns out I don't need this, so if you're not interested either we can just close it up.

kdb424:
This looks like a clean change. We should probably consider actually approving this assuming a rebase.

jepler:
I have rebased the branch.

An alternative did occur to me in the meantime:  instead of hard-coding this as being "about" seesaw, still introduce _getio.  The base implementation would say something like
```
    def _getio(self, pin):
        if hasattr(pin, 'value'): return pin
        return digitalio.DigitalInOut(pin)
```
so now you can say in your keyboard layout `S9 = adafruit_seesaw.digitalio.DigitalIO(seesaw, 9)...` and put `S9` in your `col_pins`.

A seesaw subclass could override _getio, if it's still desirable for ints to be magic (I'm not convinced of this):
```
    def _getio(self, pin):
        if isinstance(pin, int):
             return adafruit_seesaw.digitalio.DigitalIO(self.seesaw, pin)
        return super()._getio(pin)
```

kdb424:
I don't happen to have a seesaw on hand to test. It's possible that @klardotsh will have some input here.

dzervas:
hey @jepler take a look at my PR: #161 
My idea comes with a user-facing boilerplate (and I don't know if we want that). Converting my example to work with the seesaw is pretty easy, why don't you give it a try?

kdb424:
This has been inactive for a long while. Closing unless we see more activity.

PR #154, opened 2020-06-05
Basic BLE functionality

ekx:
I've implemented basic BLE functionality. Before merging it should be considered that to get it to run on a Adafruit Feather nRF52840 Express I had to increase `upervisor.set_next_stack_limit(4096 + 1024)` in `boot.py` to 4096 + 4096 and compile everything to mpy files. So at least on that hardware memory limits seem to be pretty tight

### DIFF
--- a/kmk/ble.py
+++ b/kmk/ble.py
@@ -1,0 +1,60 @@
+from adafruit_ble import BLERadio
+from adafruit_ble.advertising import Advertisement
+from adafruit_ble.advertising.standard import ProvideServicesAdvertisement
+from adafruit_ble.services.standard.hid import HIDService
+
+from kmk.hid import AbstractHID, HIDUsage, HIDUsagePage, HIDReportTypes, HID_REPORT_SIZES
+
+
+BLE_APPEARANCE_HID_KEYBOARD = 961
+
+
+class BLEHID(AbstractHID):
+    def post_init(self):
+        self.devices = {}
+
+        hid = HIDService()
+ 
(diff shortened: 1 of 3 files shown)


kmk/ble.py
```diff
@@ -0,0 +1,64 @@
...
+        advertisement = ProvideServicesAdvertisement(hid)
+        advertisement.appearance = BLE_APPEARANCE_HID_KEYBOARD
+        scan_response = Advertisement()
+        scan_response.complete_name = 'KMK Keyboard'
+
+        ble = BLERadio()
+        if not ble.connected:
+            ble.start_advertising(advertisement, scan_response)
+            while not ble.connected:
+                pass
```
klardotsh, line 31: This is *really* aggressive - it means we'll theoretically "snap out of it" faster when a connection is established, but in the mean time will draw more power.

This is okay for now - I eventually want to do a second pass on BLE support some day anyway.

kdb424: This is a pretty standard behavior on keyboards as it's meant to "snap out of it" fast. We may be able to do some sort of software switch to "turn off" though most keyboards have a physical switch to the battery. If we have a key to switch between "bluetooth mode" and "hid mode" this could also force a bypass as a "sort of off" state none the less.

klardotsh:
Ugh, Circle is broken because... reasons I guess. `s3cmd` isn't finding its config file and I'm too lazy to deal with it right now. Merging, and this commit won't hit the CDN that nobody (probably?) uses anyway.

PR #161, opened 2020-09-13
MCP23017 support & lab68 user keymap

dzervas:
There is some `matrix` hackery to be able to pass a `DigitalInOut` instead of a pin to columns/rows. This is required as MCP23017 (IO expander) does not expose the `Pin`s.

Also I added my keymap for lab68 - the keyboard that I'm making: https://github.com/dzervas/lab68


kdb424:
Seems like once you can clean up the formatting issues with black shown in the circleci fail log, it should be good to merge!

kdb424:
Going to merge this as it's not your fault that this one failed. That's a deployment issue that caused this hangup that @klardotsh will need to resolve at some point.

PR #164, opened 2020-09-14
Move CI to Github Actions

### LINKED ISSUE #145: (Investigate?) Move CI to Github Actions
This perhaps invalidates or changes #144 

Github Actions CI is available for our repo finally. It's probably sane to keep everything in one place, so investigate moving to Github Actions. We should generally be able to do all the same things sanely I hope.

dzervas:
Closes #145 

I've created two workflows:

One that is executed on all pushes and MRs to test the code
One to deploy when either a push to master or a tag creation has happened. I'm using the current makefile as-is (hence the `CIRCLE_` env var hack)

I'm guessing that some github secrets are needed?


dzervas:
For some reason my repo is not even with master? I have my commits that were already made in this repo. Weird shit. Is that a problem for you? Should I rebase?

kdb424:
@klardotsh  handles the workflow of this repo. Hopefully he'll be able to review this one.

dzervas:
Also maybe on each tag creation we should create a release and upload the artifact there with a sum of the commits? We want that?

klardotsh (changes requested):
@dzervas yeah I"d recommend always working from a feature branch (`git checkout -b dzervas/my-thing`) and rebasing that off upstream master (I find it helpful to keep a git remote around called `upstream` when working on 3rd party projects - ex. `git remote add upstream git@github.com:KMKfw/kmk_firmware; git fetch --all; git rebase -i upstream/master` if I want to rebase a forked feature branch off the upstream master)

Generally a fan of this. While we're at it, we can outright remove the need for env var secrets by just moving to Github releases instead of the `s3cmd` push to DigitalOcean Spaces (aka the `cdn.kmkfw.io`). I'm happy to take that work on if you don't want to, but if you think it's not too much work, that could get the permissions thing out of the way.

I don't see GH Actions triggering on this PR, which is a regression from the current CircleCI behavior - this seems to just be related to the `on: push: branch: ['master']` declaration in `deploy.yml`, but ultimately, `make test` at the very least needs to be run on each inbound commit, whether it's a feature branch within the org, a pull request, or a release branch (eg. `master`). I'm happy to pick this up if you don't want to - I've got experience with GH Actions on other projects at this point so like to think I have an idea of what's going on now

dzervas:
the action wouldn't trigger on this PR - these commits are still on my repo, not yours. also i could steal secrets that way

but also yes, it's due to branch master

kdb424:
@klardotsh Is this something that we are still interested in, or should I search for other options as circleCI is still broken.

klardotsh:
I misread this the first time. This looks... basically fine as-is, and it's easy to patch going forward if anything is wonky. I'd rather have green MRs than absolutely-perfect forever. Merging, and copying env vars over from Circle.

klardotsh:
Looks like there's a massive collection of linter failures to figure out, but that's not your fault. This is good to go, I'll follow up the linter stuff as a follow-up commit/PR.

PR #165, opened 2020-09-14
[BLE] Add support to change BLE name from KMKKeyboard.go

dzervas:
The following can be used to change advertising BLE name:

```python
    keyboard.go(hid_type=HIDModes.BLE, ble_name='Lab68')
```

Also I'm passing keyword arguments from `KMKKeyboard.go` to `KMKKeyboard.hid_helper()` to allow for various possible configurations later on (USB pid/vid, bluetooth stuff, etc.)

### DIFF
--- a/kmk/ble.py
+++ b/kmk/ble.py
@@ -14,19 +14,20 @@ BLE_APPEARANCE_HID_KEYBOARD = 961
 
 
 class BLEHID(AbstractHID):
-    def post_init(self):
+    def post_init(self, ble_name='KMK Keyboard', **kwargs):
         self.devices = {}
 
         hid = HIDService()
 
         advertisement = ProvideServicesAdvertisement(hid)
         advertisement.appearance = BLE_APPEARANCE_HID_KEYBOARD
-        scan_response = Advertisement()
-        scan_response.complete_name = 'KMK Keyboard'
 
         ble = BLERadio()
+        ble.name = ble_name
+        # ble.tx_power = 2
+
         if not ble.connected:
-            ble.start_advertising(advertisement, scan_response)
+            ble.start_advertising(advertisement)
             while not ble.connected:
                 pass
 
--- a/.gitignore
+++ b/.gitignore
@@ -125,3 +125,4 @@ mnt2/
 kmk/release_info.py
 kmk/release_info.mpy
 *.mpy
+.vscode
--- a/kmk/hid.py
+++ b/kmk/hid.py
@@ -41,7 +41,7 @@ HID_REPORT_SIZES = {
 class AbstractHID:
     REPORT_BYTES = 8
 
-    def __init__(self):
+    def __init__(self, **kwargs):
         self._evt = bytearray(self.REPORT_BYTES)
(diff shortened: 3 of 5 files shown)


kdb424:
Looks good to me.

kdb424:
If not given an name to advertise will this be an issue? I am waiting on my bluetooth boards to come in to test, but seeing as the only assignment that I see is in your keymap, I thought it was worth an ask.

dzervas:
default name is `KMK`

kdb424:
Totally missed that. Awaiting on final review by @klardotsh or a week from today, whatever comes first.

kdb424 (approved):
Looks reasonable to merge.

kmk/ble.py
```diff
@@ -14,19 +14,20 @@
...
         hid = HIDService()
 
         advertisement = ProvideServicesAdvertisement(hid)
         advertisement.appearance = BLE_APPEARANCE_HID_KEYBOARD
-        scan_response = Advertisement()
-        scan_response.complete_name = 'KMK Keyboard'
 
         ble = BLERadio()
+        ble.name = ble_name
+        # ble.tx_power = 2
```
klardotsh, line 27: generally commented-out code doesn't stick around in the tree, but I'm not going to block merge over it. more an FYI for the future.

kmk/ble.py
```diff
@@ -14,19 +14,20 @@
...
+
         if not ble.connected:
-            ble.start_advertising(advertisement, scan_response)
+            ble.start_advertising(advertisement)
```
klardotsh, line 30: I'm not super familiar with the BLE API: why is `scan_response` now not needed?

klardotsh (commented):
Is there a reason for the kwargs-tinkering throughout here? It doesn't look like it's ultimately used anywhere in this PR

klardotsh:
I guess I forgot to mention in the review comment: this generally looks good, I'm only pausing final approval over the curiosity questions; this is explicitly *not* a blocked merge ultimately unless more questions come out of the above. Thanks for the code!

kmk/ble.py
dzervas: Yea, it's a placeholder. TX power is not yet supported, but it should be in the near future

kmk/ble.py
dzervas: it's already created by default and we don't need to keep track of more variables than we should. Memory is already very tight

dzervas:
About the `**kwargs` the idea was to be add even more settings in the future, but now that i think of it, it's not need it.

Rebasing this MR is a bit of a nightmare RN (got to rebase 6+ branches) so if only this is left, merge it and I'll fix it on a later MR

dzervas:
Oh wait, you mean the whole play where I feed down kwargs from `keyboard.go` to `BLEHID.__init__`. Yea, in order to pass some options during ONLY the initialization of HID class, that's what I came up with. A regular class attribute wouldn't do it, as the user could change it on runtime and expect it to be reflected on the controller

PR #167, opened 2020-09-21
BLE Multiple Connections

dzervas:
Adds support for more than one simultaneously connected centrals. The keys `BT_NXT` and `BT_PRV` are added to pivot between the active connection (where to send the keystrokes)

PR #170, opened 2020-09-29
Allows continuation of firmware if bluetooth can't be imported

kdb424:
This was completely missed for boards that do not have bluetooth. Ultimately I'd like to not import this if we don't have a reason, but guarding it would at least allow for boards that can't import to actually work on mainline kmk.

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -7,7 +7,12 @@ import busio
 import gc
 
 from kmk import led, rgb
-from kmk.ble import BLEHID
+
+try:
+    from kmk.ble import BLEHID
+except ImportError:
+    print('Bluetooth is unsupported ')
+
 from kmk.consts import KMK_RELEASE, LeaderMode, UnicodeMode
 from kmk.hid import USBHID, AbstractHID, HIDModes
 from kmk.internal_state import InternalState
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -8,13 +8,8 @@ import gc
 
 from kmk import led, rgb
 
-try:
-    from kmk.ble import BLEHID
-except ImportError:
-    print('Bluetooth is unsupported ')
-
 from kmk.consts import KMK_RELEASE, LeaderMode, UnicodeMode
-from kmk.hid import USBHID, AbstractHID, HIDModes
+from kmk.hid import AbstractHID, HIDModes
 from kmk.internal_state import InternalState
 from kmk.keys import KC
 from kmk.kmktime import sleep_ms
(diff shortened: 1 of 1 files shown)


dzervas:
Even if bluetooth is supported and the `adafruit_ble` library is there, there's no reason to load it into memory if it's not actually used (a person could be pivoting between USB/BT). I suggest moving the import line without the exception handling under the `elif hid_type == HIDModes.BLE:` line in `kmk/kmk_keyboard.py`

kdb424:
I am looking at moving a lot of imports to a better place ultimately to save on memory, and ultimately agree with you. This was a quick reaction to get KMK running on Circuitpython 6.0 as the nice!nano and several other controllers have no older firmware, and `adafruit_ble` is no longer included, or supported, I am not sure. Currently I'm using this branch to even boot on that board unfortunately.

I do completely agree with you, though I offered this as it's less likely to become a debate and is only a fast solution to a problem that ultimately needs solved and discussed.

dzervas:
You'll have to download the library - [circup](https://github.com/adafruit/circup) can help. It works flawlessly on my Particle Xenon (nrf52840) on CircuitPython 6

kdb424:
I had tried to do that on my nice!nano, but it ran out of ram. I increased the stack size in order to avoid the running out of memory on recursion depth, and then ran out of memory at other places. Boot ram needs managed more now regardless, and if we add with something like git subtree, we can compile it down as well.

kdb424:
Even on #167 I am unable to boot an Cpy 6 and had to resort to not importing bluetooth. I added adafruit_ble, but ended up with even more errors. I will look for a more long term solution, but it's still a problem.
`main.py output:
  File "main.py", line 1, in <module>
  File "kmk/boards/converter/fourtypercentclub/luddite.py", line 3, in <module>
  File "kmk/kmk_keyboard.py", line 10, in <module>
  File "kmk/ble.py", line 1, in <module>
  File "adafruit_ble/__init__.py", line 38, in <module>
ImportError: no module named '_bleio'`

kdb424:
I've delayed the imports. This should be done regardless, and allows us to switch HID modes and not load the ones not needed otherwise. I feel this should be a good middleground, though many imports could be delayed in order to save even more resources.

dzervas (approved):

dzervas:
`ImportError: no module named '_bleio'` on a bluetooth enabled board or on a SAMD? If it's on nrf it's SUPER weird

kdb424:
The M4 does not have bluetooth, which is what causes the firmware to fail on that controller. Seeing as the converter in this repo is made for that controller, it's probably best if we can unbreak this quickly in master. Any additional input, or may I merge this to get a working master branch?

dzervas:
merge

PR #171, opened 2020-09-29
Introduce rotary encoder

### LINKED ISSUE #79: Re-support rotary encoders
`kmk/rotary_encoder.py` has not been updated to support the new event loop/internals. Fix this, and possibly as part of this work also use `rotaryio` before falling back to our internal implementation.

dzervas:
Closes #79.

The way I introduced the new encoder is a bit bizarre, but the overhead is almost nothing and can be used on different layers, etc.

To add an encoder (more than one are supported), `keyboard.rotaries` should be populated with the `rotaryio.IncrementalEncoder` objects (array), like so:

```py
from rotaryio import IncrementalEncoder
keyboard.rotaries = [IncrementalEncoder(board.A4, board.A5)]
```

It may not be the best in terms of user-friendlyness, but it's just two lines and easy enough.
Now to define which keys are pressed when the encoder is turned right or left, an additional column should be added to each layout (will be skipped during regular matrix scan) for each encoder. The first (top-most) row key is pressed when the wheel turns right (position is incremented) and the second (from the top) is pressed when the wheel turns left (position is decremented), like so:

```py
keyboard.keymap = [
    # Qwerty
    # ,--------------------------------------------------------------------------------------------------------.
    # |   `  |   1  |   2  |   3  |   4  |   5  |   6  |   7  |   8  |   9  |   0  |   -  |   =  | Bksp | Del  | Right
    # |------+------+------+------+------+------+------+------+------+------+------+------+------+------+------|
    # | Tab  |   Q  |   W  |   E  |   R  |   T  |   Y  |   U  |   I  |   O  |   P  |   [  |   ]  |   \  | PgUp | Left
    # |------+------+------+------+------+-------------+------+------+------+------+------+------+------+------|
    # | Esc  |   A  |   S  |   D  |   F  |   G  |   H  |   J  |   K  |   L  |   ;  |   '  |      |Enter | PgDn |
    # |------+------+------+------+------+------+------+------+------+------+------+------+------+------+------|
    # | Shift|   Z  |   X  |   C  |   V  |   B  |   N  |   M  |   ,  |   .  |   /  |Shift |      |  Up  | Ins  |
    # |------+------+------+------+------+------+------+------+------+------+------+------+------+------+------|
    # | Ctrl | GUI |  Alt  |      |      |Space |      |      |  Fn  | Alt  | Ctrl | Left |      | Down | Right|
    # `------------------------------------------------------------------------------------------+------+------'
    [
        KC.GRV,  KC.N1,   KC.N2,   KC.N3,   KC.N4,     KC.N5,   KC.N6,   KC.N7,   KC.N8,   KC.N9,   KC.N0,   KC.MINS, KC.EQUAL, KC.BSPC,   KC.DEL,   KC.RIGHT,
        KC.TAB,  KC.Q,    KC.W,    KC.E,    KC.R,      KC.T,    KC.Y,    KC.U,    KC.I,    KC.O,    KC.P,    KC.LBRC, KC.RBRC,  KC.BSLASH, KC.PGUP,  KC.LEFT,
        KC.ESC,  KC.A,    KC.S,    KC.D,    KC.F,      KC.G,    KC.H,    KC.J,    KC.K,    KC.L,    KC.SCLN, KC.QUOT, XXXXXXX,  KC.ENTER,  KC.PGDN,  XXXXXXX,
        KC.LSFT, KC.Z,    KC.X,    KC.C,    KC.V,      KC.B,    KC.N,    KC.M,    KC.COMM, KC.DOT,  KC.SLSH, KC.RSFT, XXXXXXX,  KC.UP,     KC.INS,   XXXXXXX,
        KC.LCTL, KC.LGUI, KC.LALT, XXXXXXX, XXXXXXX,   KC.SPC,  XXXXXXX, XXXXXXX, FN,      KC.RALT, KC.RCTL, KC.LEFT, XXXXXXX,  KC.DOWN,   KC.RIGHT, XXXXXXX,
```

(Note the rightmost column. It's beyond my keyboards matrix. In the above example, when I rotate right, the right key is pressed and left when I rotate left. It's very convenient BTW...

On every step, the apparent key is pressed and then unpressed, so if you very quickly turn the wheel to the left 10 steps, the second key from the right-most column will be pressed AND unpressed 10 times - this of course introduces some lag on very fast turns, but not much, also no keypress is lost.

To add more encoders, just add more objects to `keyboard.rotaries` and more columns. In a 15 column keyboard, first rotary will use 16th column, 2nd rotary 17th, etc.

The performance impact is minimal (printed time taken to scan the rotaries and it's mostly less than a millisecond), though I'm a bit worried about the memory impact due to the way I wrote the implementation.

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -23,7 +23,7 @@ class KMKKeyboard:
 
     row_pins = None
     col_pins = None
-    rotaries = None
+    rotaries = []
     diode_orientation = None
     matrix_scanner = MatrixScanner
     uart_buffer = []
@@ -191,7 +191,7 @@ class KMKKeyboard:
             self.coord_mapping = []
 
             rows_to_calc = len(self.row_pins)
-            cols_to_calc = len(self.col_pins)
+            cols_to_calc = len(self.col_pins) + len(self.rotaries)
 
             if self.split_offsets:
                 rows_to_calc *= 2
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -23,6 +23,7 @@ class KMKKeyboard:
 
     row_pins = None
     col_pins = None
+    rotaries = None
     diode_orientation = None
     matrix_scanner = MatrixScanner
     uart_buffer = []
@@ -70,6 +71,7 @@ class KMKKeyboard:
             'leader_timeout={} '
             'hid_helper={} '
             'extra_data_pin={} '
+            'rotaries={} '
             'split_offsets={} '
             'split_flip={} '
             'target_side={} '
@@ -95,6 +97,7 @@ class KMKKeyboard:
             self.leader_timeout,
             self.hid_helper.__name__,
             self.extra_data_pin,
+            self.rotaries,
             self.split_offsets,
             self.split_flip,
             self.target_side,
@@ -264,6 +267,7 @@ class KMKKeyboard:
             rows=self.row_pins,
             diode_orientation=self.diode_orientation,
             rollover_cols_every_rows=getattr(self, 'rollover_cols_every_rows', None),
+            rotaries=self.rotaries,
         )
 
         # Compile string leader sequences
--- a/kmk/matrix.py
+++ b/kmk/matrix.py
@@ -32,8 +32,7 @@ class MatrixScanner:
         # When the encoder is incremented column n from the last column and row
         # 1 is pressed and when decremented row 2 - n is the index of the encoder.
         if rotaries:
-            if self.len_rows < 2:
-                raise ValueError('There are less than 2 rows but there are rotary encoders, which need 2 or more rows')
+            assert self.len_rows > 2, 'There are less than 2 rows but there are rotary encoders, which need 2 or more rows'
 
             self.len_cols += len(rotaries)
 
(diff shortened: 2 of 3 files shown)


kmk/kmk_keyboard.py
```diff
@@ -7,7 +7,6 @@
 import gc
 
 from kmk import led, rgb
-from kmk.ble import BLEHID
```
kdb424, line 10: I believe that this is a mistake. This needs reverted in order to cleanly merge.

kmk/kmk_keyboard.py
```diff
@@ -223,6 +225,7 @@ def go(self, hid_type=HIDModes.USB):
         elif hid_type == HIDModes.USB:
             self.hid_helper = USBHID
         elif hid_type == HIDModes.BLE:
+            from kmk.ble import BLEHID
```
kdb424, line 228: Again, this is not a BLE PR, please revert this as well.

kdb424:
Seems generally clean. Conflicts need to be resolved, and some reverts to match master, but much appreciated!

kdb424 (approved):
Looks fantastic. I see no problems with this. Will give @klardotsh  a bit to review but I will not block merge.

kdb424:
Quick thought before merging, please add documentation for this before merging.

klardotsh:
This is neat, and is a fairly concise PR for the functionality, but I'm unsure how I feel about rotaries being added to the matrix. Is there a way we can separate rotary encoders to some separate construct (`self.rotary_scan` or something, idk) such as how it was in the 2018 version of rotary encoders? Or is jamming these things into the matrix the most elegant solution we have for the level of flexibility we want?

klardotsh:
Either way on the above, this should be cleanly rebased on `master` as it is currently unmergable after all the other PRs went through

dzervas:
no, you're right, needs to be done using a separate matrix. I thought it was a good idea at the moment, but really it just complicates code & docs a lot. Leave the PR hanging and I'll find some time to fix it (rebase, implementation and add docs)

kdb424:
Incompatible with the new target. Porting should be much easier once #189 is merged. Closing for now.

PR #172, opened 2020-09-29
Switch to initiator/target

kdb424:
Wording change to go along with world trends on open source code. Simple wording change, no functionality changed at all.

### DIFF
--- a/docs/split_keyboards.md
+++ b/docs/split_keyboards.md
@@ -14,22 +14,22 @@ keyboard.uart_pin = board.SCL
 Useful config options:
 ```python
 keyboard.split_flip = True  # If your boards are identical but one is flipped, this option is for you 
-keyboard.split_offsets = [6, 6, 6, 6]  # This is the how many keys are on each column on the "Master" half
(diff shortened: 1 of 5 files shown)


docs/split_keyboards.md
```diff
@@ -14,22 +14,22 @@ keyboard.uart_pin = board.SCL
 Useful config options:
 ```python
 keyboard.split_flip = True  # If your boards are identical but one is flipped, this option is for you 
-keyboard.split_offsets = [6, 6, 6, 6]  # This is the how many keys are on each column on the "Master" half
+keyboard.split_offsets = [6, 6, 6, 6]  # This is the how many keys are on each column on the target" half
```
dzervas, line 17: missed a "

dzervas (commented):
Correct me if I'm wrong but didn't you change both "master" and "side" words to target? wouldn't that introduce confusion?

docs/split_keyboards.md
kdb424: This is correct. The "target" is the master half. This terminology can be found in fibre networks.

kdb424: Don't know what I thought I was responding to. Fixing. Thanks.

PR #173, opened 2020-09-30
iris port to nicenano

kdb424:
Ported the Iris Rev 2 for the nice!nano. Needs tested on a real board, but testing on a bare PCB works great.


dzervas (approved):
LGTM!

kdb424:
I'll be waiting on a report of a working board in case I got the pins incorrect. @klardotsh actually owns that board, hence why I am waiting on his review.

klardotsh:
Holding this at least temporarily since I do indeed own two Irises (Irii?), one of which came home from the office when The World Ended (TM) and thus can be a donor board for testing this

kdb424:
Closing as an automated tool can guarantee correctness for #189

PR #178, opened 2020-10-10
Remove GC. Not needed on newer Cpy versions

kdb424:
Reduces boot times by half, and boots fine even on the M4.


kdb424:
Accidentally stuck some formatting in on a file as well, though it's unrelated, and was also tested on that board.

PR #181, opened 2020-10-15
fix: everyone's linting; loosen one linter rule on user keymaps

klardotsh:


kmk/kmk_keyboard.py
```diff
@@ -279,20 +277,16 @@ def go(self, hid_type=HIDModes.USB, **kwargs):
                 del self.leader_dictionary[k]
 
         while True:
-            state_changed = False
```
kdb424, line 282: I actually make use of this feature in Bluetooth splits when it comes to power save. Can we keep thees? There are several instances below as well.

kmk/kmk_keyboard.py
```diff
@@ -279,20 +277,16 @@ def go(self, hid_type=HIDModes.USB, **kwargs):
...
-
             if self.split_type is not None and self.is_target:
                 update = self._receive_from_initiator()
                 if update is not None:
                     self._handle_matrix_report(update)
-                    state_changed = True
```
kdb424, line 288: Here as well

kmk/kmk_keyboard.py
```diff
@@ -305,8 +299,6 @@ def go(self, hid_type=HIDModes.USB, **kwargs):
             new_timeouts_len = len(self._state.timeouts)
 
             if old_timeouts_len != new_timeouts_len:
-                state_changed = True
```
kdb424, line 308: This as well.

kdb424 (changes requested):
Other than the changes listed, seems great to me.

kdb424 (approved):
Close enough. Please the linter, I'll fix my stuff later.

PR #184, opened 2020-10-16
Fully wireless bluetooth splits

### LINKED ISSUE #169: Bluetooth split support
There is currently no way to pair devices to each other and send data between them for things like split keyboard support fully wireless. This may be best after a change on how splits communicate, though I believe that it can easily work with our current codebase.

kdb424:
Tested fully working, and running on a split corne (crkbd) with no wires on a nice_nano, a TRRS connected Nyquist on an M4, and a luddite with no splits, RGB, split wired/wireless, and bluetooth all tested.

Closes #169

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -1,3 +1,3 @@
 import supervisor
 
-supervisor.set_next_stack_limit(4096 + 1024)
+supervisor.set_next_stack_limit(4096 + 768)
--- a/kmk/ble_split.py
+++ b/kmk/ble_split.py
@@ -51,9 +51,3 @@ def advertise():
         while ble.connected:
             print('Advertising complete')
             return uart
-
-def receive(uart):
-    update = uart.read(3)
-    if update == b'':
-        update = None
-    return update
--- a/kmk/ble_split.py
+++ b/kmk/ble_split.py
@@ -1,0 +1,59 @@
+import time
+
+from adafruit_ble import BLERadio
+from adafruit_ble.advertising.standard import ProvideServicesAdvertisement
+from adafruit_ble.services.nordic import UARTService
+
+def scan():
+    ble = BLERadio()
+
+    uart_connection = None
+    # See if any existing connections are providing UARTService.
+    if ble.connected:
+        for connection in ble.connections:
+            if UARTService in connection:
+                uart_connection = connection
+            break
+
+    if not uart_connection:
+        print("Scanning...")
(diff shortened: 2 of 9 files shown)


boot.py
```diff
@@ -1,3 +1,3 @@
 import supervisor
 
-supervisor.set_next_stack_limit(4096 + 1024)
+supervisor.set_next_stack_limit(4096 + 768)
```
kdb424, line 3: This change is 100% required unless we freeze the library in. Any higher and it would not have enough memory to run, and lower and we would have to small of a stack, and also crash. I believe that we should release frozen builds for the supported boards, but in the mean time, this works on every single config I have available to test.

kmk/kmk_keyboard.py
```diff
@@ -200,49 +213,75 @@ def go(self, hid_type=HIDModes.USB, **kwargs):
...
-        # Split keyboard Init
         if self.split_type is not None:
-            try:
-                # Working around https://github.com/adafruit/circuitpython/issues/1769
-                self._hid_helper_inst.create_report([]).send()
-                self.is_target = True
-
-                # Sleep 2s so target portion doesn't "appear" to boot quicker than
-                # dependent portions (which will take ~2s to time out on the HID send)
-                sleep_ms(2000)
```
kdb424, line 233: This is no longer needed as boot times do not appear to be noticeably mismatched. Test more if you will, but I found it to feel better removed.

kmk/ble_split.py
```diff
@@ -0,0 +1,66 @@
+from adafruit_ble import BLERadio
+from adafruit_ble.advertising.standard import ProvideServicesAdvertisement
+from adafruit_ble.services.nordic import UARTService
+
+ble = BLERadio()
+
+
+def check_all_connections(is_target):
```
kdb424, line 8: I probably need to test this further honestly as I don't have a way currently to test WHICH connection is dead if I lose connection. If may deadlock if it has no host to connect to. Ideas are very welcome here.

kdb424: On further inspection, we establish HID before reaching here, so at least 1 connection should also exist. If all connections are dropped via keycode, a simple device reboot would also allow reconnecting. I'm still interested in thoughts on this, but I'm not worried about it causing a deadlock on first setup, nor "normal use" as I haven't not actually run into it once.

kdb424:
Closing in favor of #189

PR #189, opened 2020-10-28
Refactor: Plugin Architecture & Explicit Data Lifecycles

kdb424:
Todo List
- [x] Finalize BLE Split full communication
- [x] Fix right half processing keys internally as left. Apply offsets at boot. Will require kmk_keyboard change.
- [x] Reduce stack size somehow
- [x] Clean up code
- [x] Decide how boards/shields are defined if changes are needed. Looking for input.
- [x] Make sure debugging messages all work
 - [x] Test power management for BLE keyboards
 - [ ] Add combos (vim style) EDIT: will come in a separate PR
 - [x] Port modtap to extension
Documentation changes for the following.
- [x]  A getting started guide that's able to be followed from 0 to running.
- [x]  Extensions changed a lot, and docs need to be updated to reflect this
- [x]  A "porting from QMK" guide for both boards and layouts. A tool may be written as well.
- [x]  Each supported "Keyboard" needs it's own readme. They can all be very similar, but links to where to buy them, and where the open source designs are online if either are available. 
- [x] Add docs for BLE HID
- [ ] Add example on building a basic extension

What else needs done?


kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,154 @@
...
+        self._advertising = False
+
+    def __repr__(self):
+        return f'BLE_SPLIT({self._to_dict()})'
+
+    def _to_dict(self):
+        return f'BLE_Split( _ble={self._ble} _ble_last_scan={self._ble_last_scan} _is_target={self._is_target} _uart_buffer={self._uart_buffer} _split_flip={self.split_flip} _split_side={self.split_side} _split_offsets={self.split_offsets} )'
+
+    def during_bootup(self, keyboard):
+        if self.split_side == 'Left':
```
hmoco, line 36: this could probs be simplified to `self._is_target = self.split_side == 'Left'`

kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,154 @@
...
+            self._is_target = True
+        else:
+            self._is_target = False
+
+        if self.split_flip and not self._is_target:
+            keyboard.col_pins = list(reversed(keyboard.col_pins))
+
+        # Attempt to sanely guess a coord_mapping if one is not provided.
+        if not keyboard.coord_mapping or (self.split_flip and not self._is_target):
```
hmoco, line 45: woah this is cool

kmk/extensions/layers.py
```diff
@@ -0,0 +1,164 @@
+import kmk.handlers.modtap as modtap
+from kmk.extensions import Extension
+from kmk.key_validators import layer_key_validator, mod_tap_validator
+from kmk.keys import make_argumented_key
+from kmk.kmktime import ticks_diff, ticks_ms
+
+
+class Layers(Extension):
```
hmoco, line 8: this whole class is so cool! I feel like I understand how layers are implemented here

kmk/extensions/layers.py
```diff
@@ -0,0 +1,164 @@
...
+    def __init__(self):
+        # Layers
+        make_argumented_key(
+            validator=layer_key_validator,
+            names=('MO',),
```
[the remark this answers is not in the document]
hmoco: in this topic, is there a permutation of extensions that is incompatible?

kmk/extensions/media_keys.py
```diff
@@ -0,0 +1,28 @@
+from kmk.extensions import Extension
+from kmk.keys import make_consumer_key
+
+
+class Layers(Extension):
```
hmoco, line 5: Is there a particular reason this is named the same as the other `Layers` extension?

kmk/extensions/split.py
```diff
@@ -0,0 +1,111 @@
...
+from kmk.matrix import intify_coordinate
+
+
+class SplitType:
+    UART = 1
+    I2C = 2  # unused
+    ONEWIRE = 3  # unused
+
+
+class Split(Extension):
```
hmoco, line 13: would this benefit from a bit of oop as it seems to overlap in functionality to the `BLE_Split`?

kmk/boards/converter/keebio/viterbi_r2.py
```diff
@@ -13,3 +16,8 @@ class KMKKeyboard(_KMKKeyboard):
     split_type = 'UART'
     split_flip = True
     split_offsets = [7, 7, 7, 7, 7]
+
+    rgb_ext = RGB(pixel_pin=rgb_pixel_pin, num_pixels=12)
+    layers_ext = Layers()
+    split = Split(uart_pin=uart_pin, split_offsets=split_offsets)
+    extensions = [rgb_ext, split, layers_ext]
```
hmoco, line 23: After reading this through, I kinda get why this is done this way (as opposed to OOP). It's very readible and if you want anyone to write a new converter for a keeb, you dont want them to have to worry about inheritance patterns (although I think oop would be helpful on the overlapping extensions themselves)

hmoco (commented):
Did a quick run through of this PR. Lots of incredibly cool things!

kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,154 @@
...
+from adafruit_ble.services.nordic import UARTService
+from kmk.extensions import Extension
+from kmk.hid import HIDModes
+from kmk.kmktime import ticks_diff, ticks_ms
+from kmk.matrix import intify_coordinate
+
+
+class BLE_Split(Extension):
+    def __init__(
+        self, split_flip=True, split_side=None, split_offsets=[], hid_type=HIDModes.BLE
```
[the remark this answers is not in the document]
kdb424: Extensions run once at boot, and are never called again. Good to know though.

kmk/extensions/layers.py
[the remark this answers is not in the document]
kdb424: Keys are loaded in order.This allows users/extensions to forcefully override stock/other keys. Check kmk/keys.py for more comments on this topic I believe.

kmk/extensions/leader.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class LeaderMode:
+    TIMEOUT = 0
+    TIMEOUT_ACTIVE = 1
+    ENTER = 2
+    ENTER_ACTIVE = 3
+
+
+class Leader(Extension):
```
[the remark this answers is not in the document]
kdb424: Basically this https://beta.docs.qmk.fm/using-qmk/advanced-keycodes/feature_leader_key

kmk/extensions/media_keys.py
kdb424: CopyPaste error. good catch.

kmk/extensions/split.py
kdb424: Some functions overlap, though some of BLE extends past what I can do with a wire right now. The main reason for not loading anything BLE in this fie is to prevent issues on boards without BLE, and to reduce memory usage, though I'm open to ideas that take these into consideration.

user_keymaps/klardotsh/klarank_featherm4.py
```diff
@@ -55,16 +64,9 @@
...
 
-_______ = KC.TRNS
-xxxxxxx = KC.NO
-HELLA_TD = KC.TD(
-    KC.A,
-    KC.B,
-    send_string('macros in a tap dance? I think yes'),
-    KC.TG(1),
-)
+keyboard.extensions = [leader_ext]
```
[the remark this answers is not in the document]
kdb424: Order only matters if an extension would add keys that replace other keys.

kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,154 @@
...
+        self._uart = None
+        self._uart_connection = None
+        self._advertisment = None
+        self._advertising = False
+
+    def __repr__(self):
+        return f'BLE_SPLIT({self._to_dict()})'
+
+    def _to_dict(self):
+        return f'BLE_Split( _ble={self._ble} _ble_last_scan={self._ble_last_scan} _is_target={self._is_target} _uart_buffer={self._uart_buffer} _split_flip={self.split_flip} _split_side={self.split_side} _split_offsets={self.split_offsets} )'
```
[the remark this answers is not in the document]
kdb424: That part of python I'm not super familiar with. Feel free to contribute some code, and I can give it a test. It's mostly cpython compat, and I believe that works as well, I'm just unfamiliar.

kmk/extensions/ble_split.py
kdb424: Not super familiar with that syntax, but yeah, it definitely could. That was a dirty port from when split_side could change due to wired being different. Good catch. I think that `self._is_target = bool(self.split_side == 'Left')` is also valid.

kmk/extensions/ble_split.py
kdb424: It was actually wrong. Updated.

kmk/boards/converter/keebio/viterbi_r2.py
kdb424: Decided that "boards" needed to come out anyways, and be a really slimmed down thing as most are just pinouts. Docs will give simple copy paste extentions on turning on RGB, or enabling splits ect.

kmk/extensions/power.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def usb_rescan_timer(self):
+        return bool(ticks_diff(ticks_ms(), self.usb_last_scan) > 5000)
+
+    def usb_time_reset(self):
+        self.usb_last_scan = ticks_ms()
+        return self
+
+    def usb_scan(self):
+        # TODO Add USB detection here. Currently lies that it's connected
```
dzervas, line 92: As per https://circuitpython.readthedocs.io/en/latest/shared-bindings/usb_hid/#usb_hid.devices shouldn't this be `return len(usb_hid.devices) > 0`?

kmk/extensions/power.py
```diff
@@ -0,0 +1,93 @@
...
+        self.enable = not bool(self.usb_scan)
+
+    def before_matrix_scan(self, keyboard_state):
+        if self.usb_rescan_timer():
+            self.enable = not bool(self.usb_scan)
+
+    def after_matrix_scan(self, keyboard_state, matrix_update):
+        if matrix_update:
+            self.psave_time_reset()
+        else:
```
dzervas, line 30: `elif not self.usb_scan()` - Why sleep when you have "infinite" power? (muahahahahaha)

kmk/hid.py
```diff
@@ -221,3 +222,109 @@ def hid_send(self, evt):
...
             evt[1 : HID_REPORT_SIZES[reporting_device_const] + 1]
         )
+
+
+class BLEHID(AbstractHID):
+    try:
+        # TODO FIXME This is wrapped in a try block to prevent crashes on M4 devices
+        from adafruit_ble import BLERadio
+        from adafruit_ble.advertising.standard import ProvideServicesAdvertisement
+        from adafruit_ble.services.standard.hid import HIDService
```
dzervas, line 232: shouldn't these be inside `__init__`?

dzervas (changes requested):

kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,147 @@
...
+            self.initiator_scan()
+
+    def initiator_scan(self):
+        self._uart = None
+        self._uart_connection = None
+        # See if any existing connections are providing UARTService.
+        self._connection_count = len(self._ble.connections)
+        if self._connection_count > 0 and not self._uart:
+            for connection in self._ble.connections:
+                if UARTService in connection:
```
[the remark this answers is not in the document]
kdb424: Yeah, this needs refined by someone that knows bluetooth. I "made it work" but it so far has not connected to anything but the correct device even downtown. That said, better safe than sorry. When not connected, both parts seize up completely, so it's pretty easy to notice attacks at a security level, but still better if there are multiple split boards in the same place.

kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,147 @@
...
+                    self._uart_connection = connection
+                    self._uart = self._uart_connection[UARTService]
+                    break
+
+        if not self._uart:
+            print('Scanning')
+            self._ble.stop_scan()
+            for adv in self._ble.start_scan(ProvideServicesAdvertisement, timeout=20):
```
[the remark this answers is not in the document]
kdb424: May be worth it, but generally this is just the other board being reset/ect. I'll add that if I remember though.

kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,147 @@
...
+                print('Advertising complete')
+                break
+        self._ble.stop_advertising()
+
+    def ble_rescan_timer(self):
+        return bool(ticks_diff(ticks_ms(), self._ble_last_scan) > 5000)
+
+    def ble_time_reset(self):
+        self._ble_last_scan = ticks_ms()
+        return self
```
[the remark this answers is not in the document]
kdb424: Porting error. Good catch. Made sense on the branch it came from, but no longer does.

kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,147 @@
...
+        self._ble.stop_scan()
+        return
+
+    def target_advertise(self):
+        self._ble.stop_advertising()
+        print('Advertising')
+        # Uart must not change on this connection if reconnecting
+        if not self._uart:
+            self._uart = UARTService()
+        advertisement = ProvideServicesAdvertisement(self._uart)
```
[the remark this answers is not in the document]
kdb424: Willing to accept commits on this part. I know very little about BLE in general. It would need to be configured in some sane way that the keyboard knew on both sides what to look for as well as the only thing that will look for this, is the right half. Maybe uuid?

kmk/extensions/power.py
kdb424: If that works, that would be amazing. This is what we are currently doing to detect if USB is detected or not https://github.com/KMKfw/kmk_firmware/blob/015e0cdfd74dce94610aa95bef913a2d3eb717e7/kmk/kmk_keyboard.py#L227

kmk/extensions/power.py
kdb424: This is the answer that didn't used to exist, but now does. Beat me to finding the answer that I hadn't looked for yet.

kmk/hid.py
kdb424: If it didn't blow memory out on devices that have these libs, but aren't using BLE, yes. I need to get all of this section cleaned though. Thanks for the reminder.

kmk/extensions/power.py
dzervas: No idea, I just googled "circuitpython usb_hid" and got that result. Needs test

kmk/hid.py
dzervas: oh you mean just the class definition blows up the memory? damn. Maybe we can split all the HID implementations in different files and import them at will

kdb424: adafruit_ble is pretty heavy alone. The main reason it's in the try block is boards like the M0 (Itsy bitsy) will crash without the try on it. This needs to at least be split into BLE/USB.

kmk/extensions/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+
+    # The below methods should be implemented by subclasses
+
+    def on_runtime_enable(self, keyboard):
+        pass
+
+    def on_runtime_disable(self, keyboard):
+        pass
+
+    def during_bootup(self, keyboard):
```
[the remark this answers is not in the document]
kdb424: If you are willing to put in the code, and document/test it, would gladly accept it if it runs within the specs.

kmk/extensions/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def disable(self, keyboard):
+        self._enabled = False
+
+        self.on_runtime_disable(self, keyboard)
+
+    # The below methods should be implemented by subclasses
+
+    def on_runtime_enable(self, keyboard):
+        pass
```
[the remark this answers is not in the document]
kdb424: Will be in an upcoming commit. Good catch.

kmk/extensions/ble_split.py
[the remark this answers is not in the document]
kdb424: Forgotten note that there is a timeout that will be hit every 5 seconds that forces a rescan in a forever loop as half of a keyboard is not useful.

kmk/extensions/power.py
kdb424: https://github.com/adafruit/circuitpython/pull/3513/files
Looks like I'm waiting on this PR based on the comments. Same exact use case, and power related. Works with the deep sleep branch as well

kmk/extensions/ble_split.py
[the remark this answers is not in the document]
klardotsh: IIRC `__dict__` basically doesn't exist (at least in a useful form) on CircuitPython, a decision that was inherited from MicroPython. That said, I haven't looked into it in quite a while.

kmk/extensions/rgb.py
```diff
@@ -119,7 +175,7 @@ def hsv_to_rgb(self, hue, sat, val):
 
         else:
             base = ((100 - sat) * val) / 100
-            color = int((val - base) * ((hue % 60) / 60))
+            color = (val - base) * ((hue % 60) / 60)
```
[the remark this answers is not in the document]
klardotsh: Testing in a SUPER primitive form existed back in 2018; I gave up bothering when I realized I needed to mock out so much of CircuitPython's IO classes to test basically anything (a side effect of how non-separated this codebase was prior to my initial work on this branch and @kdb424's completion/adoption of it).

The new setup is almost certainly something we could run through `pytest` more easily, but I'm not gonna say this is a requirement for merge here.

user_keymaps/klardotsh/klarank_featherm4.py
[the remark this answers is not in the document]
klardotsh: @kdb424 that may be true in the current extensions implemented but is factually incorrect with regards to the big picture. Behavior is fully dependent on extension order; all lifecycle methods are executed in the insertion order of each extension in the list (one example of many is in https://github.com/KMKfw/kmk_firmware/pull/189/files#diff-33f8629da9c68d11cab09019d0863c9f5c911b8fef213df8475b227bd7256eacR392)

kmk/boards/converter/keebio/viterbi_r2.py
klardotsh: @hmoco fwiw a lot of the non-OOP patterns in this codebase are inherited from my general distaste for it. I stick to an OOP-lite style with FP influence in Python in general, and inheritance chains more than a layer or two deep tend to set off my spidey sense.

klardotsh:
Alright, this is a big one to get through, so it'll probably take me a bit. First off, thanks @kdb424 for adopting my long-neglected-and-abandoned branch and finishing it up; I really hope the plugin architecture and streamlined/standardized data flow patterns end up making more sense to future folks who have to dive in (rather than the tangly spaghetti that previously existed during the "just get an MVP out the door" stage).

My general plan of attack is:

- finsh chiming in on any relevant existing conversations in the review, I see a few folks have already jumped in with some discussions, thanks y'all!
- review `kmk_keyboard.py`, which is of course the main meat and potatoes of a KMK board (given that it controls the event loop), and also the file that needed the most surgery (since most of a keyboard's lifecycle used to live in that file, and now is being split out)
- review each extension (I expect this should go fairly quickly)
- anything else

Not super sure of an ETA but I'll try to make some time for it today and/or tomorrow so revisions can get rolling somewhat quickly. Cheers!

klardotsh:
@kdb424 though one thing right off the bat - would you mind doing a rebase+squash (`git rebase -i origin/master` - https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d) to trim the noise out of the commit log? I think there's likely a few themes in here where there was a burst of commits for a given feature/extension, then moved on to the next one and another burst of commits, etc. If we can get that down to a commit per extension or a commit per couple of extensions at a time, that'd make the log way less chatty to sift through

kmk/extensions/rgb.py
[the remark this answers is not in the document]
kdb424: Not to mention that I plan to completely redo/replace all of this due to lack of matrix support, and it's also just too damn long.

kmk/boards/converter/keebio/viterbi_r2.py
hmoco: @klardotsh fair enough! I totally dont feel like I have enough context that I can confidently say oop would be better haha. As far as spidey senses go, I can understand why several layers of inheritance would not be ideal, especially if these files are meant to be played with by people that aren't necessarily programmers (although the repetition triggered my spidey senses lol which is why I commented on it)

kdb424: I think that I've decided that these will be offered in a different way. I needed to get "boards" out of `/kmk` anyways as they were about to grow massive. I can offer many more iterations of choice in a git repo similar to how QMK does it, and still allow imports. This solves all of the "I don't have enough control" problems, as well as the "which microcontroller is which one" problem.

kdb424:
The only other change that I would propose other than what I have put here is pulling powersave natively into KMK. That would allow hooks for extentions to do thins like "on powersave enable" and the like. Things that I could see that useful are bluetooth using slower connection intervals, RGB turning off or limiting brightness, OLED's turning off, slowing timers ect.

kmk/extensions/ble_split.py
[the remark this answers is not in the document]
kdb424: I couldn't get a solid yes out of anyone in the Circuitpython discord, including the core dev team. I'm going to assume it's a solid "no" it's not supported.

kmk/extensions/split.py
kdb424: Looking back at this further, there are a lot of functions that must act differently, and jamming all of that into a single class would get more difficult to read. When extending this later, I could import functions from the wired split into the wireless where there is overlap, keeping readability, and that allows BLE to stay completely out of memory on wired only boards. Wired keyboards will need to be brought up to spec with BLE, which requires writing a onewire protocol first, and is out of scope for this PR, much the same as doing RGB matrix is out of scope. If I'm wrong, please correct anything here.

hmoco: Makes sense! I appreciate the thought youve put into responding to my comments!

kmk/extensions/ble_split.py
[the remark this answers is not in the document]
kdb424: @dzervas I have looked into this and for some reason (possibly bug), I can't get an advertisement name. I may add connecting by mac address on the scanning side as that I can get. I'm aware that you don't have the time to really look at KMK any more, but if you end up coming back to prove me wrong.

kmk/extensions/ble_split.py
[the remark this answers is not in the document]
kdb424: I've actually done this though it doesn't have an effect. It changes it's name now based on it's volume name, and can even get which side it is by the last letter "L" or "R".

kdb424:
Power is still an extension, but a standardized extension. Docs will up updated accordingly.

kdb424:
There is a critical bug in this branch that needs debugged. I don't know what could be affecting just my luddite, but I can not boot this branch as board.D1 (board.TX) is in use already. The only way this should happen is if split keyboards are being loaded, and they are not in extensions.

kmk/extensions/power.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+        make_key(
+            names=('PS_OFF',), on_press=self._ps_disable, on_release=handler_passthrough
+        )
+
+    def __repr__(self):
+        return f'Power({self._to_dict()})'
+
+    def _to_dict(self):
+        return f'''Power(
```
klardotsh, line 34: as with the others, return an actual dictionary here

kmk/extensions/power.py
```diff
@@ -0,0 +1,155 @@
...
+        return
+
+    def on_powersave_disable(self, keyboard):
+        keyboard._trigger_powersave_disable = False
+        self.disable_powersave()
+        return
+
+    def enable_powersave(self, keyboard):
+        '''Enables power saving features'''
+        print('Psave True')
```
klardotsh, line 85: print -> debug

kmk/extensions/power.py
```diff
@@ -0,0 +1,155 @@
...
+            if not self._psp:
+                self._psp = digitalio.DigitalInOut(self.powersave_pin)
+            self._psp.direction = digitalio.Direction.OUTPUT
+            self._psp.value = True
+
+        self.enable = True
+
+    def disable_powersave(self):
+        '''Disables power saving features'''
+        print('Psave False')
```
klardotsh, line 99: print -> debug

kmk/hid.py
klardotsh: I don't like the idea of wrapping the world in a giant try/catch. Move these imports to the top of the file and wrap the imports alone in a try/catch if necessary for whatever reason. We can solve the memory blowout in docs - that's the whole point of extensions. If you don't need `BLEHID`, don't' import it or you'll explode your RAM. If you do need it, run KMKPython. Seems straightforward to me unless I'm missing / forgetting some context here.

kmk/hid.py
```diff
@@ -221,3 +222,110 @@ def hid_send(self, evt):
...
+        from storage import getmount
+
+        BLE_APPEARANCE_HID_KEYBOARD = const(961)
+        # Hardcoded in CPy
+        MAX_CONNECTIONS = const(2)
+
+        def post_init(self, ble_name=str(getmount('/').label), **kwargs):
```
klardotsh, line 239: This inherently means if I have two halves of a keyboard running stock firmware and I haven't changed the filesystem labels on either, I'll have name collisions. While BT addresses based on MAC addresses/GUIDs in general and thus it's not the end of the world, it's still confusing to say the least. Is there another way we can handle this?

kmk/kmk_keyboard.py
```diff
@@ -90,223 +87,354 @@ def __repr__(self):
...
-        :param update:
-        '''
         if update is not None:
+            self._on_matrix_changed(update[0], update[1], update[2])
+            self.state_changed = True
 
-            self._state.matrix_changed(update[0], update[1], update[2])
+    #####
+    # SPLICE: INTERNAL STATE
+    # TODO FIXME CLEAN THIS
```
klardotsh, line 120: well.... is it cleaned? can we get rid of these comments?

kmk/kmk_keyboard.py
```diff
@@ -90,223 +87,354 @@ def __repr__(self):
...
+        for k, v in timeouts:
+            if k <= current_time:
+                v()
+                del self._timeouts[k]
+
+        return self
+
+    #####
+    # SPLICE END: INTERNAL STATE
+    # TODO FIXME REMOVE THIS
```
klardotsh, line 286: same remarks as above

kmk/kmk_keyboard.py
```diff
@@ -90,223 +87,354 @@ def __repr__(self):
...
+        '''
+        Attempt to sanely guess a coord_mapping if one is not provided. No-op
+        if `kmk.extensions.split.Split` is used, it provides equivalent
+        functionality in `on_bootup`
+
+        To save RAM on boards that don't use Split, we don't import Split
+        and do an isinstance check, but instead do string detection
+        '''
+        if any(
+            x.__class__.__module__ == 'kmk.extensions.split' for x in self.extensions
```
klardotsh, line 313: these two ifs can be simplified:

```python
if any(x.__class__.__module__ in {'kmk.extensions.split', 'kmk.extensions.ble_split'} for x in self.extensions)
```

kmk/kmk_keyboard.py
```diff
@@ -90,223 +87,354 @@ def __repr__(self):
...
-            if not isinstance(k, tuple):
-                new_key = tuple(KC[c] for c in k)
-                self.leader_dictionary[new_key] = v
+        return self
 
-        for k, v in self.leader_dictionary.items():
-            if not isinstance(k, tuple):
-                del self.leader_dictionary[k]
+    # Only one GC to allow for extentions to have room.
+    # There are random memory allocations without this
```
klardotsh, line 354: are they random or just fragmented?

kmk/kmk_keyboard.py
```diff
@@ -90,223 +87,354 @@ def __repr__(self):
...
-                    # This keyboard is a initiator, and needs to send data to target
-                    self._send_to_target(update)
+        for ext in self.extensions:
+            try:
+                ext.during_bootup(self)
+            except Exception:
+                print('Failed to load extention', ext)
+                import time
+
+                time.sleep(5)
```
klardotsh, line 371: time should basically always be abstracted by `kmktime`, yeah? `sleep_ms(5000)`

kmk/preload_imports.py
```diff
@@ -1,45 +0,0 @@
-# Welcome to RAM and stack size hacks central, I'm your host, klardotsh!
```
klardotsh, line 1: oh thank zeus this file is dead

have you tested that this is backwards compatible to M4? M4 was originally where this hack had to be developed

klardotsh (changes requested):
I'm withholding final approval for now but broadly speaking have few things to say about the implementation other than "thanks for picking up my stale branch and finishing it!". It seems to largely stick to the spirit and style I had in mind when I started on this work last fall.

Most of the comments here are in one of a few categories:

- documentation/copy changes, particularly for clarity to beginners
- small code cleanliness things
- gating debug features behind debug flags appropriately
- questions around testing "edge" cases

It's a lot of commentary but I think each bit is fairly small on its own, and I don't see anything HUGE in the way of merging this fairly soon. Cheers!

boards/converter/keebio/nyquist_r1.py
```diff
@@ -10,6 +13,12 @@ class KMKKeyboard(_KMKKeyboard):
...
     rgb_pixel_pin = board.TX
     uart_pin = board.SCL
+    extra_data_pin = board.SDA
     split_type = 'UART'
     split_flip = True
     split_offsets = [6, 6, 6, 6, 6]
+
+    rgb_ext = RGB(pixel_pin=rgb_pixel_pin, num_pixels=12)
+    layers_ext = Layers()
+    split = Split(uart_pin=uart_pin, split_offsets=split_offsets)
```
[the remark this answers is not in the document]
kdb424: None of these matter. They are all getting changed. That's an artifact. Corne is the one to pay attention to.

boards/crkbd/README.md
```diff
@@ -0,0 +1,26 @@
...
+Corne LP  
+[Boardsource](https://boardsource.xyz/store/5f2efc462902de7151495057)  
+
+Extentions enabled by default  
+- Layers
+- RGB
+- BLE_Split
+
+Recommended Extentions
+- Split
```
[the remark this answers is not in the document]
kdb424: BLE_Split or split. They have different functions and therefor are different plugins.

boot.py
```diff
@@ -1,3 +1,3 @@
 import supervisor
 
-supervisor.set_next_stack_limit(4096 + 1024)
+supervisor.set_next_stack_limit(4096 + 4096)
```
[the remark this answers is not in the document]
kdb424: This boot.py is correct for stock circuitpython. The forked version will not need this, but stock users will in fact still need it.

kmk/kmk_keyboard.py
kdb424: That one is on you. I couldn't figure out how to get this one out cleanly, which is why it's there. would mark this one as help_required.

kmk/kmk_keyboard.py
kdb424: Clean hard resets on the controller. I believe that I sorted it, but it's only fragmentation if Circuitpython itself has bugs in it. These are clean resets on the controller, not soft resets.

kmk/kmk_keyboard.py
kdb424: This can probably be removed. It's just been an artifact mostly. Up to you on which you prefer.

kmk/preload_imports.py
kdb424: It's running on the live code in this PR on Nyquist (wired split) luddite (also M4) and corne BLE split nice!nano on custom firmware. Full test suite.

kmk/kmk_keyboard.py
klardotsh: at this point you should be able to remove the lines assuming the splicing work is done, right? just add another commit to the stack and make sure it's at the end of the rebase and not early on?

kmk/kmk_keyboard.py
klardotsh: ah. let's clean up that comment to be more explicit then about what happens/the risks thereof

kmk/kmk_keyboard.py
klardotsh: yeah, if we can remove the sleep, awesome, if not, then let's `kmktime` it

kmk/hid.py
kdb424: That's fair enough then.

docs/Getting_Started.md
```diff
@@ -0,0 +1,41 @@
...
+ ## Getting KMK
+ You can always find the latest releases on our CDN, in [compiled and optimized](https://cdn.kmkfw.io/kmk-latest.zip) and 
+ [raw, hackable text file](https://cdn.kmkfw.io/kmk-latest.unoptimized.zip) forms. These follow the `master` branch here on GitHub. 
+ Just get the KMK folder and drop this directly in the CIRCUITPYTHON directory (not in a sub folder). Make sure to extract the zip, 
+ and put the `kmk` folder on the root of the CIRCUITPY drive on the microcontroller
+
+## Turning a controller into a keyboard
+### Supported keyboards
+If your keyboard and microcontroller are officially supported, it's as easy as visiting the page for your files, and dropping them 
+on the root of the "flash drive". Those pages can be found [here](https://github.com/KMKfw/boards). You will need the `kb.py` and `main.py`. More advanced instructions
```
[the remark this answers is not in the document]
kdb424: This is to allow for boards with multiple controllers. I can have a single main.py with a seperate kb.py for nice!nano, and another for itsybitsy as there will be some boards that can take both, but act the same otherwise. This also allows converter/nicenano with a single keymap.

docs/Getting_Started.md
```diff
@@ -0,0 +1,41 @@
...
+If you are doing a hand wire, check [here](handwiring.md)
+
+## Additional features
+Want to have fun features such as RGB, split keyboards and more? Check out what extensions can do [here](extensions.md)
+
+## Debugging
+Debugging help can be found [here](debugging.md)
+
+## Additional help and support
+ If you need support with KMK or just want to say hi, find us in [#kmkfw:klar.sh on Matrix](https://matrix.to/#/#kmkfw:klar.sh).  This channel is bridged to Discord [here](https://discordapp.com/widget?id=493256121075761173&theme=dark) for convenience.
```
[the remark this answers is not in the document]
kdb424: Must have missed that one. I had a kernel panic, and it must have not saved.

docs/Offiially_Supported_Microcontrollers.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Nice!Nano
+Features include
+- Pro-micro pinout
+- Both USB HID and Bluetooth support
+- Can do bluetooth split keyboards with no wires at all
+- Has battery support including charging
+
+Downsides
+- \$25 USD per microcontroller at most retailers
```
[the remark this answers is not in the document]
kdb424: Showed incorrect in my editor without, and showed correctly in preview, but works for me.

docs/Offiially_Supported_Microcontrollers.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Adafruit ItsyBitsy nRF52840 Express
+Features include
+- Both USB HID and Bluetooth support
+- More affordable than the Nice!Nano at only \$18
+
+Downsides
+- Needs adapted to work with pro-micro pinout keyboards. Adapter can be found [HERE](https://github.com/KMKfw/kmk_firmware/tree/master/hardware)
+- No battery support without addon board found[HERE](https://www.adafruit.com/product/2124) 
+
```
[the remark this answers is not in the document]
kdb424: Recommendations, yes, support, no. Those will live in the category of handwire niche which is why they are not directly mentioned, as well as no major keyboard manufacturer will get support, so they were dropped as reference.

docs/config_and_keymap.md
```diff
@@ -2,9 +2,10 @@
 
 KMK is configured through a rather large plain-old-Python class called
 `KMKKeyboard`. Subclasses of this configuration exist which pre-fill defaults
-for various known keyboards (for example, many Keebio keyboards are supported
-through our ItsyBitsy to ProMicro pinout adapter). This class is the main
-interface between end users and the inner workings of KMK. Let's dive in!
+for various known keyboards (for example, many QMK keyboards are supported
```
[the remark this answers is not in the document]
kdb424: QMK is what most boards are supported from, so it's a refrence that people running QMK can likely come over. Things like ZMK are not well supported, and TMK died and got expanded on my QMK which is why it's not mentioned. I can remove this, though it was to show compatability.

docs/config_and_keymap.md
```diff
@@ -77,29 +75,10 @@ keyboard.keymap = [[KC.A, KC.B]]
...
   console. This is very rarely needed, but can provide very valuable information
   if you need to open an issue.
 
-- `unicode_mode` from `kmk.consts.UnicodeMode`, which defines the default
-  operating system implementation to use for unicode sequences (see examples
-  below, or `unicode.md`. This can be changed after boot with a key (see
-  `keycodes.md`)
-
-- `tap_time` which defines how long `KC.TT` and `KC.LT` will wait before
+- `beyboard.tap_time` which defines how long `KC.TT` and `KC.LT` will wait before
```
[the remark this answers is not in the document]
kdb424: Things that were stripped should be moved to other sections and not as a giant blob, or the other likely thing with some of the stripping was things that no longer work the same way, and needed re-written.

kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def on_runtime_enable(self, keyboard):
+        return
+
+    def on_runtime_disable(self, keyboard):
+        return
+
+    def during_bootup(self, keyboard):
+        self._ble.name = str(getmount('/').label)
+        if self._ble.name.endswith('L'):
```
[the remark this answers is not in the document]
kdb424: While I would agree that it's rare, it's also about impossible to tell "which is which" in the case of split keyboards. That is what left me to rename them, document it in the guide, and offer it as an option. I didn't want to FORCE the opinion, so if you manually use split_side it will override the KLAR name. If not, well, it wasn't working anyways as you didn't give a split_side.

kmk/extensions/ble_split.py
```diff
@@ -0,0 +1,195 @@
...
+        return
+
+    def on_powersave_enable(self, keyboard):
+        if self._uart_connection and not self._psave_enable:
+            self._uart_connection.connection_interval = self._uart_interval
+            self._psave_enable = True
+
+    def on_powersave_disable(self, keyboard):
+        if self._uart_connection and self._psave_enable:
+            self._uart_connection.connection_interval = 11.25
```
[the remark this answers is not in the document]
kdb424: We can move it, but that is a standard that would not likely change. That is the minimum speed that BLE HID devices can act on spec. The variable number is set for powersave changes, but this number being defined in a spec is why it's hard coded and not moved to a variable using more memory for no reason.

kmk/kmk_keyboard.py
kdb424: I was under the impression that these functions were still used see calls to them.

kmk/kmk_keyboard.py
kdb424: Error spew is common when an extension is bad which is why I left the sleep. Maybe I'll leave this in under debug.

docs/config_and_keymap.md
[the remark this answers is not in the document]
kdb424: Regardless, I'll go along with it, but a quick grep of the docs, I see that as the only QMK refrence other than a link to their hand wire guide.

kmk/extensions/ble_split.py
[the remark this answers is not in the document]
kdb424: I was incorrect. Patch incoming to make that statement correct. split_side priority, fallback to last letter.

kmk/extensions/layers.py
```diff
@@ -0,0 +1,212 @@
...
+            validator=layer_key_validator, names=('TO',), on_press=self._to_pressed
+        )
+        make_argumented_key(
+            validator=layer_key_validator,
+            names=('TT',),
+            on_press=self._tt_pressed,
+            on_release=self._tt_released,
+        )
+
+    start_time = {
```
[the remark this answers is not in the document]
kdb424: simple slip. Good catch.

kmk/extensions/layers.py
```diff
@@ -0,0 +1,212 @@
...
+        # under almost all normal use cases, this will
+        # disable the layer (but preserve it if it was triggered
+        # as a default layer, etc.)
+        # this also resolves an issue where using DF() on a layer
+        # triggered by MO() and then defaulting to the MO()'s layer
+        # would result in no layers active
+        try:
+            del_idx = state.active_layers.index(key.meta.layer)
+            del state.active_layers[del_idx]
+        except ValueError:
```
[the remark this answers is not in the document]
kdb424: That was just moved, not changed. Yup.

kmk/hid.py
kdb424: The split guide recommends changing your names in the readme. More importantly though, only one half will advertise HID functionality, so it should not be an issue. At least I only see 1 device on my split keyboard.

kmk/kmk_keyboard.py
kdb424: Totally need sleep. Left these is to review as I didn't know if something else needed to get out. Intended to remove but wanted to leave markers for review reference.

kmk/hid.py
kdb424: I actually can't move that to the top level. I get errors. Feel free to give it a shot.

docs/config_and_keymap.md
```diff
@@ -2,9 +2,10 @@
 
 KMK is configured through a rather large plain-old-Python class called
 `KMKKeyboard`. Subclasses of this configuration exist which pre-fill defaults
-for various known keyboards (for example, many Keebio keyboards are supported
-through our ItsyBitsy to ProMicro pinout adapter). This class is the main
-interface between end users and the inner workings of KMK. Let's dive in!
+for various known keyboards (for example, many QMK keyboards are supported
+through our ItsyBitsy to ProMicro pinout adapter, or with a nice!nano).
```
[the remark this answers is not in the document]
kdb424: Let me know what you think of the changes, or mark resolved if acceptable.

kmk/kmk_keyboard.py
kdb424: Let me know if that's clear enough.

kmk/extensions/layers.py
```diff
@@ -0,0 +1,212 @@
...
+
+    @staticmethod
+    def _mo_pressed(key, state, *args, **kwargs):
+        '''
+        Momentarily activates layer, switches off when you let go
+        '''
+        state.active_layers.insert(0, key.meta.layer)
+        return state
+
+    @staticmethod
```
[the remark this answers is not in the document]
kdb424: There is a performance benefit to this, and seeing as it is going to be a common thing on battery power boards, I'll take everything that I can get. If you insist that it must go, I'll see to it, but I see no reason to not take battery and performance measures anywhere we can.

docs/Getting_Started.md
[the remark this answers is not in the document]
kdb424: There are now examples under /board such as the keyboard that you use. both converter board itsy and nice!nano are supported and the same main.py is reused depending on the file picked.

kmk/extensions/ble_split.py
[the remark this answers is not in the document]
kdb424: I'll address this in a follow up PR. This one is too big.

kmk/kmk_keyboard.py
```diff
@@ -90,223 +87,338 @@ def __repr__(self):
...
 
-            if self._state.hid_pending:
+        self._print_debug_cycle(init=True)
+
+        while True:
+            self.state_changed = False
+
+            for ext in self.extensions:
+                try:
+                    self._handle_matrix_report(ext.before_matrix_scan(self))
```
kdb424, line 366: Can we store and handle this and not just blind fire assuming it's a matrix report? I believe this would save power as well as allow split keyboards to report matrix updates to extensions from all parts, and not just what it's on. A simple example of why one would need to do this is OLED's timeout, calculating WPM, and several other ideas come to mind.

kmk/extensions/ble_split.py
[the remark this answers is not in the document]
klardotsh: FWIW that's not how numbers in Python work - the RAM would be similar either way, as the number has to be allocated either way, and since it's being assigned to a class instance that outlives the function being executed, it will guaranteed go onto the heap instead of the stack. I won't speak to whether the current implementation uses more RAM and/or allocation cycles or not, because that gets into quirks of the CPython VM that I don't know whether they translate to MicroPython and don't care enough right now to look up (however, if you're ever up for reading C, I'd recommend reading the relevant MicroPython source - it's a pretty approachable codebase that's reasonable to understand!)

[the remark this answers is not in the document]
kdb424: I'm willing to put that into a variable, though that number should not be touched until the BLE spec changes. My reason was mostly so it would not be touched.

kmk/kmk_keyboard.py
kdb424: This actually did not fix that 100% as I believe that there are other issues in Circuitpython. No changes to software, 3 hard reboots, some run into mem aloc issues, so do not. This may as well be removed as it doesn't solve the problem.

kmk/extensions/layers.py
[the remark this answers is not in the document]
klardotsh: I'm not aware of any performance benefits to this, other than *maybe* in the `mpy-cross` case, which isn't generally used right now. Decorators in Python are just functions that wrap other functions, `@staticmethod`s do not implicitly save CPU cycles by way of being attached to the class _definition_ instead of the _instance_.

It still stands that this is the one and only place in the extensions codebase, to my knowledge, that uses `@staticmethod` instead of instance variables.

The other thing I'll note, since I now had an excuse to read over this extension in totality, is that the fact that `@staticmethod` can even be used is a design problem. Extensions should store their state in `self`, not in `state` (which in other extensions was called `keyboard`, to make it clear that that was the global keyboard state). If a method on an extension doesn't require `self` to change the state of the board, that's concerning.

Are we able to reach a place where `Layers` is fully self-contained? Even if this involves some new methods added to the extension API (perhaps it's necessary to intercept matrix scans in a different way than the API currently allows, for example), I'd love to see `KMKKeyboard.active_layers` go away completely. If that's not possible, then we've proven that `Layers` isn't independent enough to be an `Extension`, and thus must stay part of core, much as tap dance currently must.

I guess I should have done this deep of a read on the first pass, but the ultimate verdict is that I consider addressing this thread, somehow, a merge-blocker.

kmk/hid.py
klardotsh: I guess for now the strategy can be "trust that people will follow the docs"; we can address this further if it becomes an issue.

klardotsh (changes requested):
I have one remaining merge blocker before this can go (sorry this wasn't brought up in the original review, eeks).

I've discussed privately in DM, but for publicity's sake: this branch and PR are closed to non-trivial changes at this point. Bug fixes, showstopper fixes, or removals/deferrals are still fair game, but new features should not be added here at this point. I think if we can get that MB thread out of the way, this is good to merge within a day or two of resolution.

kmk/extensions/layers.py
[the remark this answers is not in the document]
kdb424: Layers maybe should fall back to core then. I will note that @staticmethod in extensions can be useful, such as HSV to RGB conversions are only useful within RGB. As for storing state in keyboard for an extension, this will also be useful to users. I have been working on an external extension (not core) that calculates WPM, and another extension that reads that live and displays it on an OLED extension. Unless there is a better way to pass data back and fourth, this will still remain the case. I'm fine with layers going into core, but I still feel that things will still break both of these "rules".

kmk/hid.py
kdb424: It's not a matter of "trust" as it is, if you fail to set either a name, or a split_side, it won't work, and never would work. This assumes that at least one is correct, and if none, the outcome is the same.

kmk/extensions/layers.py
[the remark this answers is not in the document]
kdb424: I also intend to ultimately pull tap dance out, but leave all functions that make it work in core. Tapdange in it's current state is basically both "tapdance" as we understand it, as well as the only way to add/remove keys other than matrix events. I would like to decouple them and that would allow a nice distinction between an "extension" and "core". Extensions are something that allows you to manipulate the state, as well as extend it (layers is my prime example), while core offers you the bare functionality to get things working and helper functions to make extensions not repeat code as well as making it cleaner code.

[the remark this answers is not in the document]
klardotsh: Cross-extension communication, as in the WPM<->OLED case, should be handled as a separate issue to this Layers case. In the WPM<->OLED case, there's a clear need to share data between extensions, and until/unless there is a specific API to handle that, the `Keyboard` object is the only place to stage data for handoff. In this case, however, the data is never, and should never be, shared, and should live fully within `self`. If it cannot, it's too tightly bound to core and should not be an extension.

Extensions that leave traces of themselves in core are by definitions not extensions - maybe they're modules, but they're not extensions. I cannot implement something like `Layers` without modifying core (and the underlying point here is of course that non-KMK-core developers also would not be easily able to implement something like `Layers` out-of-tree / in a private fork, and therefore IMO `Layers` is, again, not an extension at all). Thus, Tap Dance would also need to _fully_ remove itself from core to be an extension (and if core needs to add more callbacks to the plugin API to allow for that, then that is an engineering problem we should embrace in the name of correctness, and write the methods).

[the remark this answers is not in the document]
kdb424: Layers.py in it's current form is only able to manipulate layers and is possibly incorrectly named. Maybe it should be something like layer_manipulations. It is not what is intending to BE layers, only MANIPULATE the layers. There are many extension, or I guess "module" ideas that I can come up with that would manipulate key's, manipulate layers, and otherwise alter the state. It's possible that extensions should just be removed if this sort of use is beyond the scope of what it was it was intended, and replaced with a "module" placement. Most extensions that exist will alter the state of the keyboard directly, split and ble_split both come to mind. I will concede that layers in out of place if we also pull splits back into core, as these also massively alter the state and functionality of the core.

kdb424:
Incorrect use of extensions. Closing.

Issue #192, opened 2021-02-17
Non-matrix, individual-GPIO key support, alternate RGB LED drivers


Gadgetoid:
We've been putting together an RP2040-based mech keyboard product that uses individual GPIO pins for keys in lieu of row/column scanning. From a "basic hobbyist tinkering" perspective, individual GPIOs make sense for us because they're really easy to read. It would seem they're a bit alien to the mech keyboard world, though.

RP2040 has a CircuitPython port, so it makes sense to use kmk as the go-to for firmware. However kmk is currently (and the whole mech keyboard space, perhaps?) very strongly minded toward row/col scanning.

I've been thinking carefully about how we could enhance kmk to suit our needs, without treading on any toes and- hopefully- resulting in some features that might be useful to other end users.

If we could add support for individual GPIO keyboards- perhaps by building an alternate `MatrixScanner` class that handles individual GPIOs under the hood - is this something you would be interested in a PR of?

We'd additionally need to support the i2c device we use to multiplex the RGB LEDs, but right now `rgb` is *very* tightly coupled into the `KMKKeyboard` class, with no opportunity to pass in another "LED driver". `rgb` seems to be exclusively for WS281x/SK6812 style pixels (NeoPixel) and while it could be modified to include support for other devices (sticking with the grain of your current LED config setup) it feels like it would fare better as a base class with the common elements, therefore I might try and:

1. Remove any neopixel specific code from `RGB` into a `neopixel` class, subclassing `rgb`
2. Add our driver as a class also subclassing `RGB` (might be a good time to add APA102 support, too- aka DotStar)
3. Update `KMKKeyboard` to include an `rgb` class variable, defaulting to `neopixel` that specifies the driver used for RGB
4. Use `self.rgb.RGB` in lieu of `rgb.RGB` here-  https://github.com/KMKfw/kmk_firmware/blob/b8cb4bda98057bfb3fe39ac07987f7af1c75efce/kmk/kmk_keyboard.py#L249-L254
5. `self.rgb_pixel_pin` could easily be a tuple providing data/clock for i2c or SPI-based pixels

That way a "board" can then add the relevant LED driver into the class variables of the `KMKKeyboard` subclass. This could be anything that subclasses `RGB`.

(Note: perhaps the default `RGB` functionality should be NeoPixel, to avoid name collisions and any subclass can just overload the relevant methods- ie: move the setup from `__init__` to `setup`)

Thank you!

aerialviews007:
[...]

Issue #71, opened 2018-10-16
Implement at least basic tests of the InternalState


klardotsh:
It should be reasonable to implement a shim/patch `firmware` object which manually injects various matrix scans into the "reducer" (that will be going away with #70 but the ideas will generally stand), and validate the state afterwards. This will let us ensure there are no regressions going forward (and actually give us something useful to run on Circle now that we're not stuck validating patched CircuitPython builds)

xs5871:
We have unittests now. Those sound exactly like what you're describing.

PR #193, opened 2021-02-17
Add support for alternate RGB LED types

Gadgetoid:
To demonstrate what I mean by the steps laid out in #192 I have thrown together a custom (as yet untested) APA102/DotStar LED driver that bitbangs the APA102 protocol onto the two pins given (as a tuple) by `rgb_pixel_pin`.

This involves minimal changes to `KMKKeyboard` but I've made some bigger changes to `RGB` in order to support it.

Crucially RGB's docstrings now do not mention "NeoPixel" since the term "RGB LED" is all-encompassing.

Additionally I have reduced the number of methods in the `RGB` class that directly manipulate the `neopixel` instance variable, reducing the surface area that needs implemented in a subclass. Right now a subclass need only implement `setup`, `show`, `set_rgb` and `set_rgb_fill`.

I noticed `_to_dict` also exists on `RGB` and exposes the `neopixel` variable, making it potentially something we'd want to overload for different drivers. However I can't see `_to_dict` being used anywhere, and the Python magic method name for coercion to a dictionary is `__dict__` anyway. Does this method serve a purpose I'd need to be mindful of when subclassing?

I've also brought some test fixtures into the mix here, since I used them at to do some basic due-diligence and catch any silly syntax errors (lockdown has me working with a pretty minimal setup so I need all the automation I can get to stop me being foolish). This may relate to #71 since I couldn't find any existing tests and it seems like there's only linting set up?

I submit this as a draft for your perusal and welcome feedback, since I appreciate I'm storming in here with only a limited amount of knowledge of your project/roadmap and suggesting completely tangential changes/features.

Cheers!


Gadgetoid:
Adding new dependencies to the Pipfile has regenerated Pipfile.lock which brought up a bunch of new isort errors that I've fixed in 78826d6046309b840c88e25df76b5a135e42cced

I chose to add this as a separate commit to highlight these changes, rather than rebasing it into "Set up tests for LED drivers"

They *seem* pretty reasonable, but such a huge change of CI packages in a PR feels pretty nasty.

I originally had "tox" set up, totally against the grain of your current workflow- I've added a rudimentary `pytest` setup into the `Makefile` now and it should run in your CI.

It would probably be more idiomatic to have `make test` and `make lint` part of separate GitHub Actions workflows, so tests still run in the case of linting errors.

kdb424:
Sorry for the late reply. I like the changes in here. Supporting more LED types, as well as supporting a matrix system for per key lighting animations is planned "eventually" so thanks for taking this part up. As a heads up, we are trying to ready up this, quite massive, rework to the core at some point soon, and everything we can is getting moved out to modules to save ram for those not using features. This also helps us clean up kmk_keyboard.py by preventing the changes you are doing in there, and allow that to live in the same file. The RGB code that runs in the new way can be seen here, though I think what you have could be easily adapted to work with the new system, as not many changes were needed to get the master branch worked over to that system.

https://github.com/KMKfw/kmk_firmware/blob/topic-merge-keyboard-and-state-rebased/kmk/extensions/rgb.py

Gadgetoid:
> we are trying to ready up this, quite massive, rework to the core at some point soon

Music to my ears, the `KMK_Keyboard` class handled and knew way too much about its child modules. I had to get.... uhhh.... *creative* to add a Matrix Scanner that didn't accept the standard set of constructor arguments: https://github.com/pimoroni/kmk_firmware/commit/8f82b395e1149693a28ad164b637886e40549e60

Though I notice this is still happening in your rework presently:

https://github.com/KMKfw/kmk_firmware/blob/af0324cf488907bd47246c99204c7db7ecae4d7d/kmk/kmk_keyboard.py#L329-L337

I mention it a bit here, too - https://github.com/KMKfw/kmk_firmware/issues/192

Behind the scenes I've actually bounced off this project somewhat since it was difficult to bend to my whims and - when I did get it up and running - was crashing... perhaps I was running into the very memory issues you're trying to fix, I didn't have the time to investigate. I'll try and spare the time to make another attempt when the core update is done!

kdb424:
The new core is merged. This can be ported over at some point if you have the time, or I'll pick the work up if no one gets around to it.

PR #199, opened 2021-05-16
Update docs/rgb.md for config clarity

DonutCables:
Update to move `keyboard.pixel_pin` and `keyboard.num_pixels` into the "Configuration" table for added clarity and ease of access. These are arguably the most important configs of using RGB and should be much more up front than the bottom of the doc.

Of note, is that the editing of this file shows a weakness of Tab-based alignment. Checking the "raw" versions of items such as tables shows them to be aligned, but viewing via a text editor shows a much messier situation
image
Automated formatting such as the `cformat` tool I've used for QMK edits may be worth looking into.

### DIFF
--- a/docs/rgb.md
+++ b/docs/rgb.md
@@ -42,6 +42,8 @@ Changing the **Value** sets the overall brightness.
 ## Configuration
 |Define                               |Default      |Description                                                                  |
 |-------------------------------------|-------------|-----------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


klardotsh:
@kdb424 is the original author of these docs so I'll let him speak for the table alignment stuff (though yeah, that screenshot looks.... wonky for sure)

Thanks for the PR though, agreed with moving the important stuff upwards!

PR #200, opened 2021-05-17
Dynamically create core key definitions as required

Gigahawk:
Only calling `make_key` as required can significantly reduce memory footprint on boards that don't use a lot of keycodes.

On a Feather STM32F405 Express (with the usb_hid import removed), this is the difference between being able to run and running into a memory allocation error.
Specifically:
```
import gc
print(gc.mem_free())
from kmk.keys import KC
from kmk.kmk_keyboard import KMKKeyboard

keyboard = KMKKeyboard()

keyboard.keymap = [
    [
        KC.A, KC.B,
        KC.C, KC.D,
    ]
]
print(gc.mem_free())
```
will print
```
104736
59056
```

On `topic-merge-keyboard-and-state-rebased`, this prints
```
104736
  File "main.py", line 4, in <module>
MemoryError: memory allocation failed, allocating 188 bytes
```

So this can save upwards of 40kB on boards that don't need all the key definitions.

### DIFF
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -16,6 +16,7 @@ KEY_SIMPLE = const(0)
 KEY_MODIFIER = const(1)
 KEY_CONSUMER = const(2)
 
+
 class KeyAttrDict(AttrDict):
     def __getattr__(self, key):
         try:
@@ -407,7 +408,9 @@ class KeyAttrDict(AttrDict):
         elif key in ('MEDIA_STOP', 'MSTP'):
             make_consumer_key(code=183, names=('MEDIA_STOP', 'MSTP'))  # 0xB7
         elif key in ('MEDIA_PLAY_PAUSE', 'MPLY'):
-            make_consumer_key(code=205, names=('MEDIA_PLAY_PAUSE', 'MPLY'))  # 0xCD (this may not be right)
+            make_consumer_key(
+                code=205, names=('MEDIA_PLAY_PAUSE', 'MPLY')
+            )  # 0xCD (this may not be right)
         elif key in ('MEDIA_EJECT', 'EJCT'):
             make_consumer_key(code=184, names=('MEDIA_EJECT', 'EJCT'))  # 0xB8
         elif key in ('MEDIA_FAST_FORWARD', 'MFFD'):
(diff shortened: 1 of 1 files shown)


Gigahawk:
Adding the comments and running the linter seems to have slightly reduced the RAM available:
```
104736
59008
```

But this is an acceptable hit, and would presumably go away with compilation.

klardotsh:
Holy smokes this is cool, thank you!

I'm curious if this can be made a bit more (subjectively) readable while still getting the same RAM benefits - have you by chance played with something like this? I haven't actually run this on a board, so consider this "fancy pseudocode" at best

```python
key_makers = {
    KEY_SIMPLE = make_key,
    KEY_MODIFIER = make_mod_key,
    KEY_CONSUMER = make_consumer_key
    # TODO FIXME: add shifted key support
}

key_lookups = {
    ('LEFT_CONTROL', 'LCTRL', 'LCTL'): (0x01, KEY_MODIFIER),
    ('LEFT_SHIFT', 'LSHIFT', 'LSFT'): (0x02, KEY_MODIFIER),
    # ... etc ...
}

class KeyAttrDict(AttrDict):
    def __getattr__(self, key):
        try:
            return super(KeyAttrDict, self).__getattr__(key)
        except:
            pass

        for names, (keycode, kind) in key_lookups.items():
            if any(n == key for n in names):
                key_makers[kind](code=keycode, names=names)
```

Gigahawk:
I think this is a fairly reasonable implementation of what you've asked (without comments for now).
I'm not sure it's really that much more readable, but I'll leave that for you to decide.

As for memory footprint, doing it this way seems to be a little more intensive (I think dicts internally have some kind of memory alignment requirement that makes them less efficient or something), but in the grand scheme of things I think we're grasping at peanuts now:
```
104736
51552
```

If you prefer this way of storing the keycodes I'll be happy to put the comments back in.

klardotsh:
Huh, you know, you have a point - it doesn't *actually* make things that much cleaner now that I see it in its fully blown-out form. @kdb424 you have any thoughts on https://github.com/KMKfw/kmk_firmware/pull/200/commits/49ac7c2e40f0222be63d1102b0584e0c15b085c2?

I'm almost inclined to say we can roll the most recent commit back (eek, sorry) to save whatever RAM we can, but let's see Kyle's thoughts. I'm overall a LGTM vote whichever implementation we land on though!

kdb424:
I'm going to vote with whatever saves us the most ram in all cases. That is by FAR the tightest constraint we have on almost all boards. As long as the code is understandable, and documented well, I believe that we would take whatever will save the most resources no matter what that may be.

Gigahawk:
OK, sounds good, I'll revert the last commit, if it get's approved I'll go ahead and squash.

klardotsh (approved):
:ship:  this is awesome, thank you!

klardotsh:
Thanks, `Squash and Merge` button!

Issue #91, opened 2018-11-25
GUI keyboard layout editor


kdb424:
As simple as it sounds, a gui tool to make things easier. Still trying to decide if it would be a native app or web app.

klardotsh:
"As simple as it sounds"

Tools like this are anything but easy with all the keyboards that exist out there.

dzervas:
Wouldn't a simple converter from QMK to KMK make more sense? It will attract more people (no idea what this actually means in code effort)

klardotsh:
That'd be doable with a C AST parser (or I guess, C preprocessor parser, since QMK's user-facing stuff is almost entirely implemented with preprocessor macros), or by figuring out a way to convert QMK Configurator's `keymap.json` output into Python source (via AST or via plain old `print()`) with as many keycodes preserved as possible, yeah

dzervas:
well using the C preprocessor is fairly easy - abusing the AST is definitely not. I'll dig in deeper when the time comes (months?) and report back

pullenrc:
It's not in the gui state yet, but I did start on a translator for the qmk configurator json output, which prints a kmk keymap.
https://github.com/pullenrc/kmk_firmware/tree/master/qmk_parser

kdb424:
When KMK was started, we were told by the core devs that JSON support was coming, but we didn't built it into the core. Maybe it would be reasonable to build in support for us to bring this functionality to core through a module/extension once we merge in this branch?

https://github.com/KMKfw/kmk_firmware/tree/topic-merge-keyboard-and-state-rebased

Open to thoughts on this.

klardotsh:
We finally have a for-realsies tracking PR for `vNext` (as some would call it), in #207. Let's definitely get that merged before taking this on, just for sake of staying focused and having one target to build against.

Since my track record on FOSS stuff for the past ~18 months has been absolutely shit, I don't want to overpromise - I have a bunch of other hobbies and just general _things in life_ going on right now - but I would love to either provide a helping set of eyes or hands if needed, or at least a review on a PR, for this functionality - it would be a huge boon to folks making the jump to KMK.

[...]

PR #203, opened 2021-06-14
encoder functionality added, oled integration started

pullenrc:
This commit adds a new board (teensy 4.1), brings online encoder functionality through the Keyboard class, as well as my start of oled integration.


kdb424:
Thanks for moving to the new branch. Read through the fork notes, and those seem like reasonable integrations. OLED is a bit tricky to put lines on what is "supported" vs "a feature", but the branch makes it super easy to plug and play, as well as base works off of others. I'll try to take the time here soon to push that branch to push it into master soon.

README.md
```diff
@@ -1,3 +1,22 @@
+# This fork is working to incorporate encoders, oled displays, and haptic feedback
```
klardotsh, line 1: This blurb makes context in your own repo, but doesn't make sense to be upstreamed. However, this information is still valuable to maintain - maybe it should be moved somewhere in `/doc`? Open to ideas

kmk/encoder.py
```diff
@@ -0,0 +1,151 @@
...
+            return None
+
+    # checks encoder pins, reports encoder data
+    # logic is from https://github.com/nstansby/rpi-rotary-encoder-python/blob/master/encoder.py
+    def report(self):
+        encoder_pad_a = int(self.pad_a.value)
+        encoder_pad_b = int(self.pad_b.value)
+        new_encoder_state = "{}{}".format(encoder_pad_a, encoder_pad_b)
+
+        if self.encoder_state == "11":  # Resting position
```
klardotsh, line 59: While there's some examples of "stringly-typed" code in the system still that we haven't bothered to rip out, I broadly find that [enums are a bit cleaner to work with](https://github.com/KMKfw/kmk_firmware/blob/6baaf5e5d430c82f6f01cde41875a943b247550e/kmk/consts.py#L7), and while I'm not sure entirely, I have a hunch that they're more performant on microcontrollers than strings, too

kmk/internal_state.py
```diff
@@ -12,13 +12,124 @@ class InternalState:
...
+    key_log = []
+    prev_key_string = ""
+    prev_key_log = []
+    # mode badge handling
+    prev_active_layer = []
+    mode_badge = []
+    prev_mode_badge = []
+    prev_logo = []
+
+    alphas = {
```
klardotsh, line 30: I haven't finished reading the whole PR at the time of writing this comment, so maybe it's answered elsewhere, but on a first read: why this instead of reusing what's in `kmk/keys.py`?

kmk/internal_state.py
```diff
@@ -29,20 +140,20 @@ def __init__(self, config):
...
-            'hid_pending={} '
-            'leader_mode_history={} '
-            'active_layers={} '
-            'start_time={} '
-            'timeouts={} '
-            'tapping={} '
-            'tap_dance_counts={} '
-            'tap_side_effects={}'
-            ')'
+            "InternalState("
```
klardotsh, line 143: This will fail the linter (actually, basically all strings in this PR seem to) because [we use single quotes](https://github.com/KMKfw/kmk_firmware/blob/master/setup.cfg#L6) as our lone deviation from the [Black style guide](https://github.com/psf/black)

There's maybe some argument to be made to stop deviating from Black's decisions, since Black, like `go fmt`, is meant to be a "one true way" solution, but the current status quo dates back to 2018 when I have more time to care about silly things like quoting style ¯\\_(ツ)_/¯

klardotsh (commented):
Okay, I see, it looks like the `KC` stuff is related to the QMK JSON importer work (refs #91) that I didn't realize was in here until I'd read through it all. Can we find another home for that so it doesn't eat up RAM on the MCU until/unless necessary?

kmk/encoder.py
kdb424: They are, especially when used with const types. Enums should be preferred, and set to const where possible.

kmk/internal_state.py
kdb424: `make black` should also just fix these sorts of things IIRC.

kdb424:
The other thing I have noticed with this branch is it's a mix between adding features like oled, and also QMK integration. I believe that these should be separate PR's if possible.

kmk/encoder.py
klardotsh: to be more constructive here, a potential alternative implementation here then would be:

```python
class EncoderPadState:
    OFF = False
    ON = True

# Convenience for reverse lookup
EncoderPadLookup = {
    False: EncoderPadState.OFF,
    True: EncoderPadState.ON,
}

# ...

new_encoder_state = (EncoderPadLookup[self.pad_a.value], EncoderPadLookup[self.pad_b.value])
```

This gives you a tuple of (A, B) with, ultimately, True/False values, but more human readably so, because we can then compare `new_encoder_state[0]` against `EncoderPadState.OFF`, which gives a bit clearer of a picture of what's going on to a reader.

There's further renditions of this that involve [namedtuples](https://circuitpython.readthedocs.io/en/latest/docs/library/collections.html#collections.namedtuple) to make things even more readable, but that's both optional, and not guaranteed to be portable across MCUs since `collections` is stripped from some CPy builds

pullenrc:
I have implemented what you have detailed, thanks for the input!

________________________________

pullenrc:
I will get the oled and qmk  parser separated.  I have created a separate repo for the qmk parser, and am working on improving it.

________________________________

pullenrc:
I am pretty new to black, the command 'make black' will fix the double quotes?  I couldn't find the double quoted item below; I have reformatted all " to '.

________________________________

pullenrc:
I created another repo for the json importer, and removed project pics and such, I didn't realize the entire diff was being merged and not just the contents I was working on.

________________________________

PR #208, opened 2021-06-20
The CI Error Rate Is Too Damn High (and this meme is ancient)

klardotsh:
Yeah, so anyway, these four commits get at least pre-merge CI working again, as well as `make test` locally assuming the dev has pipenv set up correctly


kdb424 (approved):
Looks clean to me. Approved.

PR #211, opened 2021-06-30
Encoder module!

pullenrc:
Here is my stab at a module for encoders.  It is tested a and working on an Atreus62 with a single encoder.  I really like the modular design of this new branch, this was pretty easy to do once I got my head wrapped around it.

### DIFF
--- a/boards/atreus62/README.md
+++ b/boards/atreus62/README.md
@@ -1,0 +1,18 @@
+# TG4X
+
+![Atreus62](https://assets.bigcartel.com/product_images/189335282/BIlqCtd.jpg?auto=format&fit=max&w=1200)
+
+Atreus62 is a 60% column staggered keyboard pinky stagger 
+
+kb.py is designed to work with the Teensy 4.1
+
+Retailers (USA)  
+[Atreus62](https://shop.profetkeyboards.com/product/atreus62-keyboard)  
+
+Extentions enabled by default  
+- [Layers](https://github.com/KMKfw/kmk_firmware/tree/master/docs/layers.md) Need more keys than switches? Use layers.
+- [RGB](https://github.com/KMKfw/kmk_firmware/tree/master/docs/rgb.md) Light it up
+- [Encoder](https://github.com/KMKfw/kmk_firmware/tree/master/docs/encoder.md) Twist control for all the things
+
+Common Extentions
+-
--- a/boards/atreus62/README.md
+++ b/boards/atreus62/README.md
@@ -1,4 +1,4 @@
-# TG4X
+# Atreus62
 
 ![Atreus62](https://assets.bigcartel.com/product_images/189335282/BIlqCtd.jpg?auto=format&fit=max&w=1200)
 
--- a/boards/atreus62/README.md
+++ b/boards/atreus62/README.md
@@ -14,5 +14,3 @@ Extentions enabled by default
 - [RGB](https://github.com/KMKfw/kmk_firmware/tree/master/docs/rgb.md) Light it up
 - [Encoder](https://github.com/KMKfw/kmk_firmware/tree/master/docs/encoder.md) Twist control for all the things
 
-Common Extentions
--
--- a/docs/encoder.md
+++ b/docs/encoder.md
@@ -4,13 +4,16 @@ Add twist control to your keyboard! Volume, zoom, anything you want.
 ## Enabling the extension
 The constructor takes a minimun of 3 arguments: a list of pad_a pins, a list of pad_b pins, 
 and an encoder_map.  The encoder_map is modeled after the keymap and works the
-same way. It should have as many layers as your keymap, and use filler keys for 
+same way. It should have as many layers as your keymap, and use KC.NO keys for 
 layers that you don't require any action.  The encoder supports a velocity mode
 if you desire to make something for video or sound editing. The direction of 
 increment/decrement can be changed to make sense for the direction the knob is 
 turning by setting the is_inverted flag.
 
 ## Configuration
+
+There is a complete example in the Atreus62 main.py
+
 Create your special keys:
 ```python
 Zoom_in = KC.LCTRL(KC.EQUAL)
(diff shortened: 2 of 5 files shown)


boards/atreus62/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Retailers (USA)  
+[Atreus62](https://shop.profetkeyboards.com/product/atreus62-keyboard)  
+
+Extentions enabled by default  
+- [Layers](https://github.com/KMKfw/kmk_firmware/tree/master/docs/layers.md) Need more keys than switches? Use layers.
+- [RGB](https://github.com/KMKfw/kmk_firmware/tree/master/docs/rgb.md) Light it up
+- [Encoder](https://github.com/KMKfw/kmk_firmware/tree/master/docs/encoder.md) Twist control for all the things
+
+Common Extentions
```
kdb424, line 17: If this section is not needed, it can be removed.

docs/encoder.md
```diff
@@ -0,0 +1,48 @@
+# Encoder
+Add twist control to your keyboard! Volume, zoom, anything you want.
+
+## Enabling the extension
+The constructor takes a minimun of 3 arguments: a list of pad_a pins, a list of pad_b pins, 
+and an encoder_map.  The encoder_map is modeled after the keymap and works the
+same way. It shoul have as many layers as your keymap, and use filler keys for
```
kdb424, line 7: `shoul` should be `should`. Typo

docs/encoder.md
```diff
@@ -0,0 +1,48 @@
...
+Add twist control to your keyboard! Volume, zoom, anything you want.
+
+## Enabling the extension
+The constructor takes a minimun of 3 arguments: a list of pad_a pins, a list of pad_b pins, 
+and an encoder_map.  The encoder_map is modeled after the keymap and works the
+same way. It shoul have as many layers as your keymap, and use filler keys for 
+layers that you don't require any action.  The encoder supports a velocity mode
+if you desire to make something for video or sound editing. The direction of 
+increment/decrement can be changed to make sense for the direction the knob is 
+turning by setting the is_inverted flag.
```
kdb424, line 11: Maybe add `is_inverted` into a code block to make that clear it means something in code.

docs/encoder.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Configuration
+Create your special keys:
+```
```
kdb424, line 15: Any python code blocks should start with 
\`\`\`python

in order to get syntax highlighting.

kdb424 (changes requested):
Once those cleanups are made, and I didn't mark repeats, I'll re-review. Good over all though, and debug should be finished or removed before final review as well. Overall, looks great.

kmk/modules/encoder.py
```diff
@@ -0,0 +1,221 @@
...
+            )
+        return keyboard
+
+    def get_reports(self, keyboard):
+        for idx in range(self.encoder_count):
+            if self.debug_enabled: # not working as inttended, do not use for now
+                print(self.encoders[idx].__repr__(idx))
+            encoder_key = self.encoders[idx].report()
+            if encoder_key is not None:
+                k = self.send_encoder_keys(keyboard, encoder_key, idx)
```
kdb424, line 220: To make this cleaner, why not just return, as opposed to storing?

kmk/modules/encoder.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def __init__(self, pad_a, pad_b, encoder_map):
+        self.pad_a = pad_a
+        self.pad_b = pad_b
+        self.encoder_count = len(self.pad_a)
+        self.encoder_map = encoder_map
+        self.make_encoders()
+
+    def on_runtime_enable(self, keyboard):
+        raise NotImplementedError
```
kdb424, line 159: These should be blank `return` as opposed to raising an error if they are intentionally not used. This is to catch you if you flat out forget to include them.

kmk/modules/encoder.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def on_runtime_disable(self, keyboard):
+        raise NotImplementedError
```
kdb424, line 162: Here as well

kmk/modules/encoder.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def during_bootup(self, keyboard):
+        raise NotImplementedError
```
kdb424, line 165: And here

kmk/modules/encoder.py
```diff
@@ -0,0 +1,221 @@
...
+
+        modified_keyboard = self.get_reports(keyboard)
+
+        return modified_keyboard
+
+    def after_matrix_scan(self, keyboard):
+        '''
+        Return value will be replace matrix update if supplied
+        '''
+        raise NotImplementedError
```
kdb424, line 180: And the rest. I won't mark them all. You get it.

kmk/modules/encoder.py
```diff
@@ -0,0 +1,221 @@
...
+    # checks encoder pins, reports encoder data
+    def report(self):
+        new_encoder_state = (
+            self.encoder_pad_lookup[int(self.pad_a.value)],
+            self.encoder_pad_lookup[int(self.pad_b.value)],
+        )
+
+        if self.encoder_state == (self.eps.ON, self.eps.ON):  # Resting position
+            if new_encoder_state == (self.eps.ON, self.eps.OFF):  # Turned right 1
+                self.encoder_direction = 'R'
```
kdb424, line 81: Could we use a bool for this, and document which is which? That should take less resources, even if only by a bit.

kdb424: Could also do something like this with a `const`. That will be easier to read, and low memory. `from micropython import const`

kdb424:
Look like a `make fix-formatting fix-isort` is in order as well.

docs/encoder.md
kdb424: `minimun` to `minimum` as well.

docs/encoder.md
pullenrc: Fixed, I think.

docs/encoder.md
pullenrc: Added to all blocks of python codee

pullenrc:
Working to get this all fixed this week, thanks for the feedback.

kdb424 (approved):
Approving this. I'll tag @klardotsh for final code review, but looks great to me.

kdb424:
@pullenrc https://github.com/KMKfw/kmk_firmware/pull/211/checks?check_run_id=3059690312 Can you check the liner errors here, and resolve those. I'll get a merge in as soon as that passes checks.

PR #230, opened 2021-08-30
Added stargazers to readme

kdb424:


klardotsh:
Not personally a fan - stars are interesting as an abstract measurement of how many people might be interested in using KMK, but IMO is not really a metric worth putting so front-and-center: it's not actionable, it doesn't help keyboard builders in any way, and honestly just kinda looks egoey to be in a README?

PR #232, opened 2021-09-12
Convert to supervisor ticks

### LINKED ISSUE #215: use supervisor.ticks_ms if available
CircuitPython adds [`supervisor.ticks_ms()`](https://circuitpython.readthedocs.io/en/latest/shared-bindings/supervisor/index.html#supervisor.ticks_ms) as a wrapping time source.  It would be nice if kmk used this in lieu of its own ticks_ms, but the substitution isn't 1:1 because `supervisor.ticks_ms()` wraps every 2**29ms (but it never loses precision).  There are routines suggested in the docs for comparing tick values and adding durations to tick values (to implement deadlines/timeouts); these will someday (soon?) be packaged into a new `adafruit_ticks` library (with compatibility code for 6.3) so it may also make sense to wait for that to pop up.

kdb424:
Resolves #215

### DIFF
--- a/kmk/handlers/stock.py
+++ b/kmk/handlers/stock.py
@@ -1,4 +1,4 @@
-from supervisor import ticks_ms
+from time import sleep
 
 
 def passthrough(key, keyboard, *args, **kwargs):
@@ -102,7 +102,7 @@ def bkdl_released(key, keyboard, KC, *args, **kwargs):
 
 
 def sleep_pressed(key, keyboard, KC, *args, **kwargs):
-    ticks_ms(key.meta.ms)
+    sleep(key.meta.ms / 1000)
     return keyboard
 
 
--- a/kmk/handlers/stock.py
+++ b/kmk/handlers/stock.py
@@ -1,4 +1,4 @@
-from supervisoor import ticks_ms
+from supervisor import ticks_ms
 
 
 def passthrough(key, keyboard, *args, **kwargs):
--- a/kmk/handlers/stock.py
+++ b/kmk/handlers/stock.py
@@ -1,4 +1,4 @@
-from kmk.kmktime import sleep_ms
+from supervisoor import ticks_ms
 
 
 def passthrough(key, keyboard, *args, **kwargs):
@@ -102,7 +102,7 @@ def bkdl_released(key, keyboard, KC, *args, **kwargs):
 
 
 def sleep_pressed(key, keyboard, KC, *args, **kwargs):
-    sleep_ms(key.meta.ms)
+    ticks_ms(key.meta.ms)
     return keyboard
 
 
--- a/kmk/kmktime.py
+++ b/kmk/kmktime.py
@@ -1,23 +1,15 @@
-import time
+import const
+
+_TICKS_PERIOD = const(1 << 29)
+_TICKS_MAX = const(_TICKS_PERIOD - 1)
+_TICKS_HALFPERIOD = const(_TICKS_PERIOD // 2)
 
 
-def sleep_ms(ms):
-    return time.sleep(ms / 1000)
+def ticks_diff(new, start):
+    diff = (new - start) & _TICKS_MAX
+    diff = ((diff + _TICKS_HALFPERIOD) & _TICKS_MAX) - _TICKS_HALFPERIOD
+    return diff
 
 
-def ticks_ms():
-    '''Has .25s granularity, but is cheap'''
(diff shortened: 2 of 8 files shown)


kmk/kmktime.py
```diff
@@ -1,23 +1,22 @@
...
+
 
 def sleep_ms(ms):
     return time.sleep(ms / 1000)
 
 
 def ticks_ms():
-    '''Has .25s granularity, but is cheap'''
-    return time.monotonic() * 1000
+    return supervisor.ticks_ms() * 1000
```
jepler, line 16: The `*1000` seems wrong, since the units are already ms.  Do you want to eliminate this function and instead supply it with `from supervisor import ticks_ms` so no wrapper is needed?

kmk/kmktime.py
```diff
@@ -1,23 +1,22 @@
...
 def ticks_diff(new, old):
-    return new - old
-
-
-def accurate_ticks():
-    '''Is more expensive, but good for time critical things'''
-    return time.monotonic_ns()
-
-
-def accurate_ticks_diff(new, old, ms):
```
jepler, line 22: `ticks_diff` always returned a signed difference, but the way I read it `accurate_ticks_diff` returned a bool indicating before/after a deadline (and needs 3 args, not two). The way you're using it below, it looks like you expect a boolean result.

I'd have explained the old `accurate_ticks_diff` as: Return `True` iff the difference between `new` and `old` is less than `ms`. This becomes something like
```
def check_deadline(new, old, ms):
    """Return True iff less than `ms` milliseconds have passed between `old` and `new`"""
    return ticks_diff(now, start) < ms
```

jepler (changes requested):
Thanks! It's nice to see someone is trying to use these routines I added to CP.

I have some comments.  I didn't do any testing, but I think it probably doesn't work properly as currently written.

kmk/kmktime.py
kdb424: Good call.

kmk/kmktime.py
kdb424: Added this. Much appreciated 👍

kmk/handlers/stock.py
```diff
@@ -102,7 +102,7 @@ def bkdl_released(key, keyboard, KC, *args, **kwargs):
 
 
 def sleep_pressed(key, keyboard, KC, *args, **kwargs):
-    sleep_ms(key.meta.ms)
+    ticks_ms(key.meta.ms)
```
jepler, line 105: this seems wrong, doesn't it still need to be sleep_ms (or `time.sleep(key.meta.ms/1000)`)?

kmk/kmktime.py
```diff
@@ -1,23 +1,15 @@
-import time
+import const
```
jepler, line 1: needs to be `from micropython import const`.

kmk/handlers/stock.py
```diff
@@ -1,4 +1,4 @@
-from kmk.kmktime import sleep_ms
+from supervisoor import ticks_ms
```
jepler, line 1: typo, should be `supervisor`.

kmk/kmktime.py
jepler: also missing `from supervisor import ticks_ms`

jepler (changes requested):
Spotted some more problems. with the changes in https://gist.github.com/b155a9afe222fda215fb2be5367d4092 I tested that KT_LT worked, so I think that TT should work too (but didn't test it). I did not test the power saving or usb/ble rescanning stuff.

kdb424:
Coffee before code. Got it. Thanks a ton. Think everything is resolved.

kmk/handlers/stock.py
jepler, line 105: did this actually get fixed? My testing didn't exercise this area of the code, but I think calling ticks_ms() with an argument is incorrect.

kdb424: It was, in fact broken. This solves it.

jepler (approved):
Thank you!  I didn't re-test but you got to everything I had spotted, and what I did test worked.

kdb424:
Gonna pass this off to @klardotsh for a final review. Thanks for all of the help @jepler !

klardotsh (approved):
Rad. Thanks @jepler and @kdb424! More ancient KMK relics sent to the shredder :recycle:

PR #233, opened 2021-09-13
Fix send_string (#206)

### LINKED ISSUE #206: send_string in master giving error
When using send_string to send 'int' this happend:
`Traceback (most recent call last):
      File "main.py", line 50, in <module>
      File "kmk/handlers/sequences.py", line 45, in send_string
KeyError: t`

offending line in main:
`INT = send_string('int')`

This works fine in the current main branch, portingmy files over to the rebased branch is when I am having the errors

Gigahawk:
Fixes #206

### DIFF
--- a/kmk/handlers/sequences.py
+++ b/kmk/handlers/sequences.py
@@ -42,7 +42,7 @@ def send_string(message):
     seq = []
 
     for char in message:
-        kc = KC[char]
+        kc = getattr(KC, char.upper())
 
         if char.isupper():
             kc = KC.LSHIFT(kc)


kmk/handlers/sequences.py
```diff
@@ -42,7 +42,7 @@ def send_string(message):
     seq = []
 
     for char in message:
-        kc = KC[char]
+        kc = KC.__getattr__(char.upper())
```
Gigahawk, line 45: Somewhat of a lazy fix, I made an attempt to get `KeyAttrDict` to also create missing key entries when accessed via `__getitem__` but I couldn't figure it out so this will have to do unless someone else has a better idea.

Seems to work fine on my end testing on a Feather RP2040

klardotsh: Preference to using `getattr(KC, char.upper())` here instead of directly calling the dunder (which is a bit of an implementation detail) but otherwise this looks good :+1: 

Thanks for doing this!

klardotsh (changes requested):

klardotsh (approved):

PR #234, opened 2021-09-14
Add QMK equivalent attributes of DiodeOrientation

Gigahawk:
IMO the existing `DiodeOrientation` names are somewhat confusing, the names used in QMK are unambiguous and likely to be more familiar with anyone coming from QMK.

Support for the old definitions remains for compatibility, but docs have been updated to point to the more familiar naming.

### DIFF
--- a/docs/Getting_Started.md
+++ b/docs/Getting_Started.md
@@ -33,7 +33,7 @@ keyboard = KMKKeyboard()
 
 keyboard.col_pins = (board.GP0,)
 keyboard.row_pins = (board.GP1,)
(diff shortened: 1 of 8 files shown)


kdb424:
I like this change, but please hit every instance in the repo if you could of both options. Merging as is breaks all keyboards and keymaps that define the old way.

klardotsh (approved):
@kdb424 no it doesn't, it's fully backwards compatible by simply setting an alias and updating the docs to use the new aliases rather than the underlying `COLUMNS` and `ROWS` names

PR #238, opened 2021-09-15
added ADNS9800 and mouse keys/mouse buttons/mouse wheel support

kbjunky:
ADNS9800 - module for interfacing with ADNS9800 optical sensor
mouse_keys - module for using mouse buttons, mouse wheel as well as moving mouse cursor with keyboard keys


kdb424:
Looks really clean. I'll let @klardotsh do a full review, but once we get tests passing, looks good by me.

kdb424 (approved):

kmk/modules/adns9800.py
```diff
@@ -0,0 +1,3313 @@
+import busio
+import digitalio
+import microcontroller
+
+import time
+
+from kmk.modules import Module
+from kmk.modules.mouse_keys import PointingDevice
+
+firmware = (
```
klardotsh, line 10: Does this blob have a source we can link (or, ideally, build ourselves)? Who/what is the copyright owner, and what is the license? As a GPLv3 project we have to be somewhat careful about what blobs we put in the code, and my spidey sense is somewhat going off here as a "I'm not sure how to navigate this one at a glance and will need to re-read with more context" matter.

klardotsh: Additionally, I'd be inclined to move this to its own file (`adns9800_firmware.py`) if we do retain it, and import it here so that the "logic" file can be short and concise, and the blobs are over in the corner somewhere.

klardotsh (changes requested):
marking req change to block merge pending firmware discussion, but the actual logic powering this seems to be really clean (not that I know much about optical sensors), so thanks for opening this and hopefully we can get it merged!

kmk/modules/adns9800.py
kbjunky: This is proprietary firmware from Avago Technologies (now Broadcom). It can be obtained from them. Afaik there's no open source alternative nor there's a way to compile it from source (8051 Asm probably).

kdb424: Curious to where you found this firmware, and what the licence is on it.

kbjunky: Just like the sensor itself, China.

kdb424: https://github.com/mrjohnk/ADNS-9800 Guessing this is where you got the firmware code from?

kbjunky: Just picked one from collection. Might be the same one or one of them. Is that a problem? If so this one file can be dropped I suppose?

kdb424: KMK is GPL licensed, so we have to be careful of what code comes in. We just need to track down the license, and make sure it's GPL comparable to be sure we aren't in violation of anything. It doesn't have to be dropped depending on the licence.

kbjunky: I'll try to find out then.

klardotsh: Broadly speaking, after reflecting on it a bit, I'd rather not take binary blobs as part of KMK's source. Instructions on how to pull such a blob yourself and use it in the module is fine, much like how MicroPython and CircuitPython do/used to handle it.

That said, if these firmwares can be built from source and are on license terms we can deal with, there could be a conversation about whether KMK should build them in CI or such. Let's cross that bridge when/if we get there, once we know more about this firmware.

kbjunky: OK. I've pushed modified src/docs to reflect your decision. I am still in the process of figuring out legal aspects of attaching this firmware here but for now this is the best solution. Even if it turns out it's OK to place is inside this repo I doubt there will ever be source code released for it so one way or another it won't end up here anyway.

klardotsh (approved):
Thanks for this! If you're able to figure out the licensing situation for (or find the source of) the firmware, I'd love to revisit that discussion. For now, merging the functionality + docs pointing to firmware instructions. Cheers!

kbjunky:
Thank you.

PR #241, opened 2021-09-17
fix doc typos and switched description for keycodes NUBS and NUHS

honboubao:
Just noticed the word extension is misspelled on several occasions. Also the description for the two ISO-layout keys are switched


kdb424:
I can't spell apparently. Thank you.

kdb424 (approved):

PR #242, opened 2021-09-17
reduce Key objects memory footprint

honboubao:
I've run into the problem of running out of memory on my Adafruit ItsyBitsy nRF52840 Express with having a few hundred Key objects floating around. The empty lists for the event handlers make up about 60% of the entire Key object, it adds up quite a bit if you have many Keys defined. They are also seldom or never used, so I changed the code for the lists to only be defined when they are really needed.

### DIFF
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -401,10 +401,6 @@ class Key:
         self.no_press = bool(no_press)
         self.no_release = bool(no_press)
 
-        self._pre_press_handlers = []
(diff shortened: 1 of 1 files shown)


klardotsh (approved):
Oh _wow_, I don't know if those lines have been touched at all since their first hackish implementation way back when. Thanks a bunch, this is going to be a huge help for small MCUs (M4, etc.) as well!

PR #243, opened 2021-09-17
always release the same key as was pressed

honboubao:
I had a problem of a layer being stuck with `MO`.

So if matrix position x has a `MO(1)` defined on layer 0 but a different key, e.g. `KC.A`, on layer 1, then the `is_pressed == True` matrix scan event will call `MO(1).on_press` and set layer 1 to be active. The `is_pressed == False` matrix scan event then will get `KC.A` from `_find_key_in_map` and call `KC.A.on_release` and there is no way to get back to layer 0.

I have changed to code to always remember the keys being pressed down and reference those same keys when the physical key is being release.

If you think about it, is there ever a case where we want to release a different keycode than was pressed?

### DIFF
--- a/kmk/handlers/stock.py
+++ b/kmk/handlers/stock.py
@@ -8,9 +8,6 @@ def passthrough(key, keyboard, *args, **kwargs):
 def default_pressed(key, keyboard, KC, coord_int=None, coord_raw=None, *args, **kwargs):
     keyboard.hid_pending = True
 
-    if coord_int is not None:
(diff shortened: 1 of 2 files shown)


kdb424:
We *used* to have this in there, but it must have gotten lost over time. Thanks 👍

klardotsh (approved):
I haven't tested this on hardware, but I'll trust that @honboubao has and merge. I'm spinning up a new KMK board this week so I guess I'll find out pretty quickly if anything's wonky.

PR #245, opened 2021-09-18
New encoder module

elric91:
Another shot around encoders, as existing one was not managing the button presses correctly (or I did not get how to use it) and had lots of options that did not seem to be used in the code or in the examples. I also modified the code to be able to define the encoders parameters _after_loading the module. Cleaning the code also mean I get rid for the time being of the velocity part (could easily be reinserted as soon as I get what it is used for)


kdb424:
I'll give it a read once it's passing tests. Run `make test lint` and resolve any errors you see or it won't pass the check that's currently failing. Thanks for taking a shot at this. The other obvious blocker before I dig into code is we'll need docs to go with it so others know how to use it.

kdb424 (approved):
Looks clean to me. @klardotsh will have a final review and merge rights.

klardotsh:
I'd strongly prefer to land on one encoders module: while I accept that KMK is probably going to end up with feature creep because folks will want X, Y, or Z from other firmwares, once we take on a feature, there should generally be one upstream "blessed" way to do things IMO, and alternative implementations can stay in forks.

That said, I'd be up for trying to merge these two implementations as and if possible, because I agree that this reads fairly simply (there's some technical remarks I could make about efficiency here or there, but that's really optimization we can handle once we've converged on a common ground between this encoder implementation and that submitted by @pullenrc)

Perhaps you two could hash out how to merge these two sets of ideas, if they should be merged at all?

elric91:
It should be fairly easy as I started from @pullenrc code, where I did 2 main things
- simplified functions which use was not obvious to me (mostly velocity)
- adapted the code so that we do not have to setup the encoder prior to loading module

if @pullenrc is OK,  I can try to merge the missing functions, but what is more important is : "is it better to define encoder prior to loading the module or after ?" (regarding to the rest of the config file) as it is the main difference between the two approaches

elric91:
I'll try a comparison to explain what I did try to achieve

Current encoder approach (just kept lines to highlight declaration steps order)
```python3
# 1. Import
from kmk.modules.encoder import EncoderHandler

# 2. Encoder Map declaration
encoder_map = [
    [ ( KC.VOLU, KC.VOLD, 2, ),  ], #  Only 1 encoder is being used, so only one tuple per layer is required
    [ (Zoom_in, Zoom_out, 1),],
    [ (_______, _______, 1),],  #  no action taken by the encoder on this layer
]

# 3. Hardware definition
encoder_ext = EncoderHandler([board.D40], [board.D41], encoder_map)
encoder_ext.encoders[0].is_inverted = True

# 4. Module loading
keyboard.modules = [layers_ext, encoder_ext]

# 5. Keymap definition
keyboard.keymap = ...
```

Proposed declaration approach
```python3
# 1. Import
from kmk.modules.new_encoder import EncoderHandler

# 2. Module loading
encoder_handler = EncoderHandler()
keyboard.modules = [layers, encoder_handler]

# 3. Hardware definition
encoder_handler.pins = ((board.GP17, board.GP15, board.GP14, False),)

# 4./5 Maps declaration 
keyboard.keymap = ...

encoder_handler.map = ( ((KC.UP, KC.DOWN, KC.MUTE),), # Standard
                        ((KC.VOLD, KC.VOLU, KC.MUTE),), # Extra
                        ((KC.A, KC.Z, KC.N1),), # NumPad not yet properly configured
                        ((KC.A, KC.Z, KC.N1),), # Gaming not yet properly configured
                        )
```

I tried to keep the same declaration order as the main keymap definition. I think that it will simplify the use of special keys defined through other modules / extensions that are declared post module load (i.e. if you want the encoder to emulate a mousewheel)
I also reintegrated the inverted property directly in encoder.pins declaration (last optional parameter)

Does it make sense ?

pullenrc:
I am glad you moved the map into the class, I have been meaning to do that.  My button is wired into my matrix, so I have been putting off handling the button press as well.  Looks good to me, merge away.

klardotsh:
If you two have reached a consensus on the design, let's go ahead and merge into `encoder.py` so there's one source of truth module for this - thanks for hashing that out!

elric91:
Almost done (done, but I want to test / improve misstep handling when you rotate the encoder very quickly). Will push an updated version coming WE

elric91:
Merge is done
Tests are OK on my side. I did update atreus/main.py as well but was not able to test it (as I do not have relevant hardware)

PR #246, opened 2021-09-19
Improve hold tap (MT, LT, TT) behaviour

honboubao:
I am experimenting with KMK for a keyboard with home row mods and layer keys. Several unexpected thing came up with the current implementation of ModTap.

With switch x as MT(KC.LSFT, KC.A) and switch y as KC.S:

1 - can't activate mod by holding (for stuff like shift click or ctrl click)

x down, wait for tap time  
expected result: shift is down  
actual result: nothing happens

mt_pressed adds the mod key to keys pressed but without any further interaction hid_pending is never set and the mod key is not reported to the OS

2 - rolling key strokes

x down, y down, x up, y up all within tap time  
expected result: as  
actual result: Sa  

Both, hold behaviour (uppercase S) and tap behaviour (a), are triggered, when it should be either one or the other. Also activation of a gets delayed.

3 - stuck mod key

x down, wait for tap time, y down, y up, x up all within tap time  
expected result: S  
actual result: S and shift is stuck down  

x down adds shift to set of pressed keys. y down reports shift down to OS. x up removes shift from pressed keys, but shift up doesn't get reported to the OS because hid_pending is not set after tap time is expired.

---

I have implemented a solution that - for now - behaves  even a little bit more tap favoured than ZMK's tap-preffered hold-tap flavor (https://zmk.dev/docs/behaviors/hold-tap/). 

This means that hold behaviour is only activated if the MT key is being held for tap time and without any interruptions. Releasing the MT key wihin tap time will trigger the tap behaviour, as will pressing a different key within tap time. This applies to MT and LT. In ZMK's [diagram](https://zmk.dev/assets/images/comparison-72cfad239e451bf5e3f8bd50cec77d3a.png) it would mean, that all four examples would result in fj.

TT behaves a little bit differently in that interruptions by other keys will lead to the hold behaviour, i.e. switching layers momentarily. I find it to be more appropriate to toggle layer (tap behaviour) state only for explicit, uninterrupted taps of the TT key.

I plan to build upon this to implement more configurable options like in ZMK.


kmk/modules/layers.py
```diff
@@ -47,8 +38,8 @@ def __init__(self):
         make_argumented_key(
             validator=layer_key_validator,
             names=('LT',),
-            on_press=self._lt_pressed,
-            on_release=self._lt_released,
+            on_press=curry(self.ht_pressed, key_type='LT'),
```
klardotsh, line 41: Preference to restoring the const values or at least storing these in variables somewhere: magic strings are prone to mistakes.

The const values that previously existed are a fair bit RAM-lighter than these strings will be, which doesn't so much matter on some of the beefier NRF and RP2040 boards of today, but does still matter on older/smaller boards running on M4 or STM32.

kmk/modules/modtap.py
```diff
@@ -21,6 +22,17 @@ def before_matrix_scan(self, keyboard):
         return
 
     def after_matrix_scan(self, keyboard):
+        '''Before other key down decide to send tap kc down.'''
+        if self.matrix_detected_press(keyboard):
+            for key, state in self.key_states.items():
+                if not state.activated:
+                    # press tap because interrupted by other key
+                    self.key_states[key].activated = 'interrupt'
```
klardotsh, line 30: again preference to making these consts or part of an Enum-esque class, and not magic strings

kmk/modules/modtap.py
```diff
@@ -35,23 +47,90 @@ def on_powersave_enable(self, keyboard):
...
-            keyboard.hid_pending = True
-            keyboard.tap_key(key.meta.kc)
-
-        self._mod_tap_timer = None
+    def ht_released(self, key, keyboard, *args, **kwargs):
+        '''On keyup, release mod or tap key.'''
+        if key in self.key_states:
+            state = self.key_states[key]
+            keyboard.cancel_timeout(state.timeout_key)
+            if state.activated == 'hold':
```
klardotsh, line 70: ditto (for every flavor of this in this file)

klardotsh (changes requested):
I'm not totally set up to test this since I don't actually use the behavior. Once the MR feedback is addressed and the merge/rebase conflicts are handled, is there anyone who can actually test this out there, or do I resort to a YOLO merge?

klardotsh:
@elric91 has graciously offered in Matrix to test this since I don't use this behavior, but I can't assign them to be a reviewer since they aren't part of the KMKfw org - so I'll just wait for a thumbs up/down comment and/or remarks from them whenever they can provide it

elric91:
Based on my quick tests, I can confirm that MT behaves as described. I need to do futher tests on LT as it acts more as a layer change than a temporary layer change (same behaviour whithout applying the patch : I may misunderstand the use of LT)

honboubao:
Thank you, @elric91  and @klardotsh for testing and checking the code.

I have now replaced the magic strings and updated to master. Additionally, I have made 2 small changes:

1) I have moved the HoldTap class to another file
2) I have modified TT slightly. The TT hold behaviour now momentarily toggles the layer instead of always setting it to active. So when the layer has been activated before by a single tap, now holding the key will deactivate the layer momentarily. When the layer was not active, holding the key will activate the layer momentarily.

klardotsh (approved):
Thanks a bunch for this - this seems to be a huge step up as far as implementation goes!

Issue #150, opened 2019-12-30
Tapdance no longer works.


kdb424:
```
Traceback (most recent call last):
  File "main.py", line 73, in <module>
  File "kmk/kmk_keyboard.py", line 300, in go
  File "kmk/kmk_keyboard.py", line 146, in _handle_matrix_report
  File "kmk/internal_state.py", line 130, in matrix_changed
  File "kmk/internal_state.py", line 139, in process_key
  File "kmk/keys.py", line 84, in _on_release
  File "kmk/handlers/stock.py", line 134, in td_released
  File "kmk/internal_state.py", line 194, in _process_tap_dance
AttributeError: 'Key' object has no attribute 'codes'

HELLA_TD = KC.TD(
    KC.A,
    KC.B,
    send_string('macros in a tap dance? I think yes'),
    KC.TG(1),
)
```

kdb424:
We need to see if this is even still a problem.

pullenrc:
This just happened to me, I'm not sure if I should have posted here first, at any rate I didn't see this until after the new issue was open.

klardotsh:
Reopening; closing #202 as dupe

klardotsh:
I need to get an actual KMK dev board back up to test this - I basically have my production board that hasn't been touched in years, and... somewhere in the pile of maker hardware are the MCUs and breadboards I used to test features on :)

Doing some heavy electrical and plumbing lifts on my sailboat this weekend so can't promise I'll have time to look into this before, say, mid week, but I'll try to at least get a repro so I can figure out what's going on.

If you haven't already, see if you can spin anything up on the "new master" branch, https://github.com/KMKfw/kmk_firmware/tree/topic-merge-keyboard-and-state-rebased. That's something like 95% complete and works on non-split boards (the discussion about what it'll take to get that across the finish line has been ongoing on the Matrix/Discord chat, drop in there any time if you need more real time help!)

pullenrc:
I will give the rebased branch a shot, sounds interesting.

pullenrc:
On the rebased branch TD is still giving this error, exactly as above.  I'll dig around, however, I am still familiarizing myself with the new branch.

xs5871:
I think this issue can be closed.

kdb424:
It should have auto closed if tagged. Thanks for the heads up!

Issue #205, opened 2021-06-16
Consider static type checking


klardotsh:
Probably something like https://github.com/Microsoft/pyright (or maybe https://pyre-check.org) would help us sure up a lot of the insanity that happens at our interface boundaries.

I don't think Micro/Circuit Python support PEP-526, so we'd be leaning on PEP-484 and whatever subset of pyright's features can fit PEP-484 syntax, but that's alright. Alternatively, we could check to see if there's a Python 3.10 to 3.5 transpiler out there that would strip, say, type aliases or rewrite pattern matches, at compile time (given that my day job is TypeScript, I'm innately familiar with this concept, but I don't know if it's made its way into Python-land yet. It could be an interesting project to write if it doesn't exist, though...)

Another thing worth noting is that we'd need to ensure that https://github.com/python/typeshed (or a similar collection we and/or Adafruit maintains - this might be something to chat with @tannewt about) has type stubs for CircuitPython's stdlib, which would be a fairly non-trivial undertaking. We only use a small subset of CircuitPython's API, though, so doing a _partial_ type stub writeup probably wouldn't be nightmarish.

klardotsh:
Other rationale: we basically don't have automated tests (very early on I started down the road of trying to do functional testing, but mocking out the MicroPython APIs proved tricky, so eventually they were removed from the tree), so a type system would at least ensure all the proverbial lego bricks were the right shape, even if not validating their behaviors precisely.

tannewt:
The typeshed equivalent for CircuitPython is the `circuitpython-stubs` package: https://pypi.org/project/circuitpython-stubs/#history

We just started releasing it automatically.

sofubi:
@klardotsh @kdb424 as we've discussed over in Matrix I'm going to try to take this on.

I do have a few questions that I feel are worth keeping in here just so we can track them.

1. Is there a target Python version I should be working for?
2. As stated above by @klardotsh `"I don't think Micro/Circuit Python support PEP-526, so we'd be leaning on PEP-484 and whatever subset of pyright's features can fit PEP-484 syntax"` I'll work on getting the codebase typed and then use the `circuitpython-stubs` package when needed.
3. I believe `hid.py` was a file that maybe I should avoid- are there any other places worth avoiding?
4. Any really great places to start?

I'll stay in touch via Matrix but if anything important comes up pertaining to this thread I'll be sure to move conversation here.

Thanks!

kdb424:
[...]

PR #247, opened 2021-09-24
fix tapdancing

xs5871:
Addresses issue #150.


elric91:
Quick test done. Applied on top of MT patch (246).
Before patch : TD not working
After patch : TD working. No regression identified after 10 min use

klardotsh (approved):
ahhhh more silly nonsense that would be caught by a type system and/or unit tests (eg #205 / #226)

thanks for fixing this!

PR #249, opened 2021-09-27
Fix LT() doesn't work

atsuyuki:
Fixed a problem where layer switching was not working in LT().


klardotsh (approved):
yep, that's a silly copy-pasta mistake :+1: 

thank you!!

PR #251, opened 2021-10-01
fix BLE HID not working with consumer keys

honboubao:


klardotsh (commented):
since I don't use bluetooth, has anyone done any regression testing here? either you, or maybe @kdb424 would be able to poke at it and make sure the basics of the rest of BLE still work well? (broadly speaking this looks pretty clean so I'm not _particularly_ concerned, but it's worth looking at I guess)

kmk/hid.py
```diff
@@ -259,72 +257,59 @@ def post_init(self):
...
-
     @property
     def devices(self):
         '''Search through the provided list of devices to find the ones with the
         send_report attribute.'''
         if not self.ble.connected:
-            return []
-
-        result = []
-        # Security issue:
```
kdb424, line 272: Is this security issue removed? If not, the warning should not be removed.

kdb424:
I don't have a bluetooth board to test any more. Looks like a clean patch. I'll ask in the matrix server if anyone can test, and if we get a report back from someone else that it works as intended, I'll call it a pass other than the one note that I left.

klardotsh (approved):

PR #253, opened 2021-10-04
Ported keymap_jp from QMK

atsuyuki:
Ported keymap_jp from QMK. With this, KMK will work as a JIS keyboard.

### DIFF
--- a/kmk/extensions/keymap_extras/keymap_jp.py
+++ b/kmk/extensions/keymap_extras/keymap_jp.py
@@ -1,0 +1,158 @@
+# What's this?　
+# This is a keycode conversion script. With this, KMK will work as a JIS keyboard.
+
+# Usage
+# ```python
+# from kmk.extensions.keymap_extras.keymap_jp import JP
+# keyboard.keymap = [ ... JP.AT ... ]
+# ```
+
+# Credit
+#  Proted from keymap_jp.h on QMK
+#  https://github.com/qmk/qmk_firmware/blob/master/quantum/keymap_extras/keymap_jp.h
+
+'''
+/* Copyright 2016 h-youhei
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http:#www.gnu.org/licenses/>.
+ *
+ * JP106-layout (Japanese Standard)
+ *
+ * For more information, see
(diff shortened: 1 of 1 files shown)


kmk/extensions/keymap_extras/keymap_jp.py
```diff
@@ -0,0 +1,157 @@
...
+     * ├──────┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴────┤
+     * │        │ Z │ X │ C │ V │ B │ N │ M │ , │ . │ / │ \\ │      │
+     * ├─────┬──┴┬──┴──┬┴───┴┬──┴───┴──┬┴───┴┬──┴┬──┴┬──┴┬──┴┬─────┤
+     * │     │   │     │Muhen│         │ Hen │K↔H│   │   │   │     │
+     * └─────┴───┴─────┴─────┴─────────┴─────┴───┴───┴───┴───┴─────┘
+     */
+    '''
+
+    # Row 1
+    ZKHK = KC.GRV  # Zenkaku ↔︎ Hankaku ↔ Kanji (半角 ↔ 全角 ↔ 漢字)
```
kdb424, line 58: Is it possible that we add this into core KMK as another name for the GRV key instead of remapping it?

kmk/extensions/keymap_extras/keymap_jp.py
```diff
@@ -0,0 +1,157 @@
...
+    N3 = KC.N3  # 3
+    N4 = KC.N4  # 4
+    N5 = KC.N5  # 5
+    N6 = KC.N6  # 6
+    N7 = KC.N7  # 7
+    N8 = KC.N8  # 8
+    N9 = KC.N9  # 9
+    N0 = KC.N0  # 0
+    MINS = KC.MINS  # -
+    CIRC = KC.EQL  # ^
```
kdb424, line 70: Same with the rest of these. We shouldn't need to double up on memory to rename everything if we can have them all properly named. I won't list this comment on every one, but you get the idea.

kmk/extensions/keymap_extras/keymap_jp.py
```diff
@@ -0,0 +1,157 @@
...
+     * ├──────┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴────┤
+     * │        │   │   │   │   │   │   │   │ < │ > │ ? │ _ │      │
+     * ├─────┬──┴┬──┴──┬┴───┴┬──┴───┴──┬┴───┴┬──┴┬──┴┬──┴┬──┴┬─────┤
+     * │     │   │     │     │         │     │   │   │   │   │     │
+     * └─────┴───┴─────┴─────┴─────────┴─────┴───┴───┴───┴───┴─────┘
+     */
+    '''
+
+    # Row 1
+    EXLM = KC.LSFT(N1)  # !
```
kdb424, line 133: KMK actually offers shifted keys directly, so mapping a shifted key shouldn't need done manually.

kdb424 (changes requested):
I like what the intent is making these easier to find, but I think these should be merged into the key names themself, and into the docs, as opposed to adding an extension for this. It will cause less bloat, run faster, and require less memory.

PR #258, opened 2021-10-16
Add user_keymap for pyKey60

dgriswo:
Repo for hardware design files: https://github.com/jpconstantineau/PyKey60
This keyboard uses the RP2040, so CircuitPython 7.0.0 is required.


kdb424:
Really clean. I'll give it a pass on splitting out the kb.py like other boards as there is no microcontroller swaps as it's integrated unless @klardotsh disagrees for some reason. Thanks a ton for the port!

PR #261, opened 2021-10-17
fix tapdance with momentary modifiers; tap-timeout per tap.

xs5871:
Improves handling of momentary modifiers (i.e. different actions on hold or press/release) as tapdance targets.
tap_time is reset for each consecutive single tap, instead of being applied to the entire tap sequence.


kdb424:
I don't, but can't imagine it causing issues. Ship it.

Issue #237, opened 2021-09-15
Migrate to CircuitPython 7 Keypad APIs


klardotsh:
Our current matrix scanner is, well, what it is. It's pure Python, it bakes in quite a few assumptions about how your keyboard looks and works (eg #192), and it cuts some interesting corners in the name of performance. Thankfully, the lovely folks at Adafruit have taken it upon themselves to support various types of matrix and non-matrix key-based input mechanisms in CircuitPython itself, with efficient and ergonomic C-backed APIs. Let's use them.

We should decide whether to retain the old matrix scanner wholesale (@Gigahawk has alluded in Matrix that this is preferable for those using GPIO expanders, as some of them expose DigitalInOut-like objects that the C HAL can't talk to), use the existing MatrixScanner class as a stub/abstraction over the CircuitPython library (I'm not entirely sure why, other than for backwards compatibility, though the MatrixScanner isn't particularly publicly-used anyway so there should be no third-party consumers), or completely scrap the file.

This trivializes #188 to the point that I'm just going to absorb it into this issue: when we migrate to Keypad, we should expose the `KeyMatrix`'s `interval` to KMK users in a reasonably ergonomic way.

XScorpion2:
[CircuitPython Keypad](https://github.com/adafruit/circuitpython/tree/main/shared-module/keypad) does not debounce the signal and the comments do not match what the code is actually doing. The `interval` parameter controls how frequently it gets the switches current state using the `common_hal_digitalio_digitalinout_get_value` function. That function returns the current raw state of the pin with no filtering or debouncing added to the return. 

So with the default implementation, that code reads the pin values every 20ms, and queues up a key event if the current state does not match the previous state. Cherry MX switches per spec are rated for 5ms of debounce, so if you close that switch 1ms right before the 20ms scan rate, that switch will be 4ms away from being considered stable per spec. So the code could read any value at that point, closed or open. So if it reads that it is still open, and the key is released before the next scan rate, then the key will never be picked up that it changed state. If you shorten this interval to accommodate this short of a press, then you risk getting multiple press and release events. 

So to reiterate, the Keypad does not implement any debounce logic, it just runs slower than debounce spec for common switches and thus hides the issue by not responding to fast keypress and release cycles.

klardotsh:
[...]

PR #262, opened 2021-10-18
Use keypad.Keys module from CP7.0 to support boards with key per GPIO

eltariel:
These changes are intended to support the [Pimoroni Keybow2040](https://shop.pimoroni.com/products/keybow-2040) (same as #193) - a 4x4 grid with an RGB LED per key. Rather than directly modifying `KMKKeyboard` to accept a different matrix, I subclassed it to `CustomMatrixKeyboard` and overrode everything that works directly with the `MatrixScanner` class.

The new scanner implements as little as possible - it's basically just a key mapping / event translation layer over the top of `keypad.Keys`. Implementing a proper matrix should be a fairly trivial variation on this, using `keypad.KeyMatrix` instead (I don't have HW to test with although I could throw something together basic pretty easily).

My next step from here probably would be to refactor `KMKKeyboard` to pull everything but the `MatrixScanner` init code out into a common base class that we can use for the existing matrix as well as the CP7.0 keypad stuff. An extension on this would be a scanner that combines other scanners, so you could mix and match key-per-gpio and a regular matrix on a single board.

Hell, we could even have a scanner that reads encoder events, or one that reads UART or i2c events, or... etc 😛 

(EDIT: this implements #237, despite it not being what I set out to do originally...)

### DIFF
--- a/boards/boardsource/microdox/kb.py
+++ b/boards/boardsource/microdox/kb.py
@@ -2,7 +2,6 @@ import board
 
 from kmk.kmk_keyboard import KMKKeyboard as _KMKKeyboard
 from kmk.matrix import DiodeOrientation
-from kmk.matrix import intify_coordinate as ic
 
 
 class KMKKeyboard(_KMKKeyboard):
@@ -16,14 +15,10 @@ class KMKKeyboard(_KMKKeyboard):
     powersave_pin = board.P0_13
 
     # NOQA
-    coord_mapping = []
-    coord_mapping.extend(ic(0, x, 5) for x in range(5))
-    coord_mapping.extend(ic(4, x, 5) for x in range(5))
-    coord_mapping.extend(ic(1, x, 5) for x in range(5))
-    coord_mapping.extend(ic(5, x, 5) for x in range(5))
-    coord_mapping.extend(ic(2, x, 5) for x in range(5))
-    coord_mapping.extend(ic(6, x, 5) for x in range(5))
-
-    # And now, to handle R3, which at this point is down to just six keys
-    coord_mapping.extend(ic(3, x, 5) for x in range(2, 5))
-    coord_mapping.extend(ic(7, x, 5) for x in range(0, 3))
+    # flake8: noqa
+    coord_mapping = [
+     0,  1,  2,  3,  4,  20, 21, 22, 23, 24,
+     5,  6,  7,  8,  9,  25, 26, 27, 28, 29,
+    10, 11, 12, 13, 14,  30, 31, 32, 33, 34,
+            17, 18, 19,  35, 36, 37,
+    ]
--- a/boards/crkbd/kb.py
+++ b/boards/crkbd/kb.py
@@ -2,7 +2,6 @@ import board
 
 from kmk.kmk_keyboard import KMKKeyboard as _KMKKeyboard
 from kmk.matrix import DiodeOrientation
-from kmk.matrix import intify_coordinate as ic
 
 
 class KMKKeyboard(_KMKKeyboard):
@@ -21,14 +20,11 @@ class KMKKeyboard(_KMKKeyboard):
     i2c = board.I2C
     powersave_pin = board.P0_13
 
-    coord_mapping = []
-    coord_mapping.extend(ic(0, x, 6) for x in range(6))
-    coord_mapping.extend(ic(4, x, 6) for x in range(6))
-    coord_mapping.extend(ic(1, x, 6) for x in range(6))
-    coord_mapping.extend(ic(5, x, 6) for x in range(6))
-    coord_mapping.extend(ic(2, x, 6) for x in range(6))
-    coord_mapping.extend(ic(6, x, 6) for x in range(6))
+    # flake8: noqa
+    coord_mapping = [
+     0,  1,  2,  3,  4,  5,  29, 28, 27, 26, 25, 24,
+     6,  7,  8,  9, 10, 11,  35, 34, 33, 32, 31, 30,
+    12, 13, 14, 15, 16, 17,  41, 40, 39, 38, 37, 36,
+                21, 22, 23,  47, 46, 45,
+    ]
 
-    # And now, to handle R3, which at this point is down to just six keys
-    coord_mapping.extend(ic(3, x, 6) for x in range(3, 6))
-    coord_mapping.extend(ic(7, x, 6) for x in range(0, 3))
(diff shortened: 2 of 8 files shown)


xs5871:
Sorry there hasn't been an update in a while.
I did some additional work:
- updates to the scanner base class
- preliminary work to integrate with split keyboards. I don't have the hardware on hand to test it and I currently don't have the capacity to build it. I'd be surprised if it would work ootb.
- implementation of an aggregate scanner and scratched the plain scanner list.

`coordinate_mapping` hurts my brain, because its somewhat backwards. As it is now, it maps virtual keys to hardware keys, that is virtual keys are the indices and hardware keys are the value -- but it's supposed to translate hardware keys to virtual ones. It's possible to map multiple virtual keys to the same hardware key, which makes conceptually no sense (although only the first one would be used). I wish it was the other way around,... but anyway.
That ties into `intifiy_coordinates`. I don't see a way to re-implement it to play nicely with continuous `key_number`s without passing it row and column count, and at that point something like `row*num_colums+column` is seems like the less convoluted approach. Personally it'd bite the bullet and break backwards compatibility at this point. I'm open for discourse.

@eltariel: how's it going on your end. I'd still rather collaborate than maintain my own fork, though I really don't want to carry over the matrix-centric bytearray report.

FoamyGuy:
@eltariel or @xs5871 Is there any further update on this effort? I am very interested in seeing the support for `keypad` finished up and merged in. 

If you are having less time to dedicate towards it now I may be able to pick it up and try to get it wrapped up, though there is still a fair amount of things I need to get my head wrapped around before I could really dive in to this specifically. I currently have somewhat little experience with KMK (set up a 2 key "hello world" type example to learn the basics). But am looking to get more involved and most of the hardware that would be easiest for me to tinker with while learning are individual IO pins per key rather than matrices

xs5871:
There haven't been any updates; what's still missing (at least from my implementation):
* documentation
* test with new release of circuitpython, as there was a bug in the key-hashing at the time
* test with split keyboards (can't do that myself)

It's not that I don't have the time to finish this. I was hoping to hear from @eltarial again, as this is their PR and we didn't merge our implementations.

FoamyGuy:
@xs5871 Thank you. Can you link me to the branch your work is in? I assume here: https://github.com/xs5871/kmk_firmware/tree/circuitpython-7-keypad would be great to confirm though so I can be sure I'm starting from the right spot. 

I'd be willing to help with documentation.

I can also test with the newest release of CircuitPython

I can try to test with a split keyboard, but am a bit less confident in this one than the other two. I don't have a split keyboard ready to go that I can test with so I'd need to build something. But I may be able to make some sort of minimal split device on a breadboard for baseline testing.

We'll see what eltariel says. If it makes sense to merge together the work from the branch in this PR with yours we could do that. Or if not and they do not have the time to dedicate any more maybe we can make a new PR from your branch after the above gets resolved.

xs5871:
@FoamyGuy that is the correct and up-to-date branch.

Checked with cp 7.1.0-beta.1 and now instead of always getting the same hash for different objects, we get different hashes every time, even if the events should compare equal according to the documentation... Oh, well.
That doesn't impact functionality though in any way whatsoever.

> I can try to test with a split keyboard, but am a bit less confident in this one than the other two. I don't have a split keyboard ready to go that I can test with so I'd need to build something. But I may be able to make some sort of minimal split device on a breadboard for baseline testing.

That would be awesome.
I would hold off on documentation until we decided which branch we're using.

> We'll see what eltariel says. If it makes sense to merge together the work from the branch in this PR with yours we could do that. Or if not and they do not have the time to dedicate any more maybe we can make a new PR from your branch after the above gets resolved.

Sure, let's do that. Looking forward to getting this issue on the way.

jpconstantineau:
@FoamyGuy I have lots of split keyboards and a few more getting ready...
Ping me when you are ready to test the split portion.  I can test wireless and wired. It would be great to be able to get both Serial and I2C working as some boards use one vs the other...

I'll be able to test with the nrf52840 and the rp2040.

jpconstantineau:
@FoamyGuy I also have splits with matrices, and other simple keys (Keys vs KeyMatrix).
Most splits are symmetrical (Let's Split and all derivatives). Some may not be (TKL with separate numpad).  Taking care of the symmetrical ones with encoders on both sides will take care of 95% of all cases.

eltariel:
Hey all, sorry for the radio silence on this. I've stalled a bit since my main dev machine died and I've only just resurrected it, and $work has been chaos as well so I haven't had a lot of time.

That said, I had a bunch of updates just about ready to go before the damn PC died. I'd cleaned up a bunch of things including killing off `coordinate_mapping` altogether in favour of letting the `AggregateScanner` I wrote merge it's scanners into a single linear keymap. It was working locally with a keybow2040 split into two virtual matrices, and I was going to try to make two of them co-operate over serial or i2c next. I also have a split board and a macro pad on the way from @jpconstantineau so I'll be able to test this a bit more thoroughly when it shows up, but... I have to dig it out from an old ssd first 😖

I've got some time off coming up soon, in theory I'll get some more done on this soon. I want to dig into @xs5871 's branch too, it sounds like we were converging on a similar solution and it's far more productive to collaborate than compete 😄

xs5871:
@eltariel: any progress on your front? We have pending PR for the scanner interface to use the CP `KeyEvents`. Separating the Matrix code and allowing different scanners should be fairly trivial without breaking functionality.

eltariel:
Last couple of commits are ready to go. A few things are broken in the boards I've added, but that's unrelated to the keypad changes.

This PR has the following changes:
- Scanner classes
  - new `Scanner` base class based on the interface for the existing `Matrix`
  - `Matrix` implements `Scanner`
  - Added `NativeKeypadScanner` with factory methods to create scanners based on `keypad.Keys` and `keypad.KeyMatrix`
- Added check for `None` before creating the scanner in `KMKKeyboard`, so you can inject an alternate scanner
- Added `pixels` parameter to the `RGB` extension to allow using something other than Neopixels - works well so far with APA102 (see keybow for an example)
- Two new boards, both with key-per-GPIO and unusual RGB LEDs:
  - Pimoroni Keybow 2040: 4x4 with an RGB LED per key. Custom RP2040 board with an IS31FL3731 LED controller (currently LEDs aren't working).
  - Pimoroni Keybow: 4x3 with an APA102 RGB LED per key in a Raspberry Pi hat. Doesn't work properly on the rpi circuitpython port, but functions perfectly with an itsybitsy via an adapter board, I'm using it with an nRF52840, but should work with minimal changes for any of the itsybitsy boards.

eltariel:
Splitting this up into separate PRs for the scanner, the RGB changes, and the pimoroni boards.

PR #265, opened 2021-10-19
WIP: Adding VColMX44 and VColChoc44 (Atreus with 44 keys and RGB)

jpconstantineau:
I would like your comments before merging in.
I'll be adding another board which uses a different GPIO addressing but has the same matrix layout.

One board uses a chip directly soldered on the PCB while the other uses a Pi Pico.  The target UF2 for CP is different, one uses the Pico while the other is designed to re-use the PyKey60 UF2 directly.

| Board | MCU | CP UF2 |
| --- | --- | --- |
| VColMX44 | Raspberry Pi Pico | Pico UF2 |
| VColChoc44 | RP2040 | PyKey60 UF2 |

The matrix configuration and mapping between RGB and keys is the same. 

Should the hardware and KB configs be separated so that the keymaps could be used for both?
Alternately, since each can only have 1 hardware configuration, should both be merged as the example for PyKey60?

I feel that having hardware and layout merged might be simpler for those starting out as there aren't any processor options for each board (unlike the ErgoTravel for example, where a NiceNano, a RP2040 Pro Micro or a BlueMicro840 could all be used with CP and KMK to power the board)

### DIFF
--- a/user_keymaps/jpconstantineau/vcolchoc44_colemak_dh.py
+++ b/user_keymaps/jpconstantineau/vcolchoc44_colemak_dh.py
@@ -17,7 +17,7 @@ keyboard.extensions.append(rgb_ext)
 FUN = KC.MO(1)
 UPPER = KC.MO(2)
 XXXXXXX = KC.TRNS
-RGB_BR =  KC.RGB_MODE_BREATHE_RAINBOW
+RGB_BR = KC.RGB_MODE_BREATHE_RAINBOW
 RGB_P = KC.RGB_MODE_PLAIN
 RGB_B = KC.RGB_MODE_BREATHE
 RGB_R = KC.RGB_MODE_RAINBOW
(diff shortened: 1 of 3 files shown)


jpconstantineau:
I just read the comments in #254 and I think I'll create what's needed in boards since they are open sourced and will be put on sale.  I'll be making the changes...

jpconstantineau:
If you mean the keymap to be all in one column. It's probably my vscode setup...  Works great when checking code in for some projects but not this one...

kdb424:
We have that formatting also in `make fix-formatting` as well. Anything in our repo should be easily human readable, which means that keymaps ironically shouldn't be formatted to the same standards as normal code.

klardotsh:
The formatting things changed when I took a (required because of python versions) update to black and/or a Flake8 plugin. Our config files imply that the user keymaps directory should still be excluded, but I guess that behavior changed when the plugin got upstreamed into Flake8....

PR #274, opened 2021-10-29
Added missing keys on lily58

TellNoLies:
The lily58 keyboard main.py was missing the top row along with the center key on the thumb cluster. I added this in the layout but did not add any pins in kb.py. I was unsure what needed to be included.


kdb424:
I'll accept your keymap as well if @klardotsh is fine as you also fixed an issue with it. I'll let them merge at their leisure.

kdb424 (approved):

TellNoLies:
Lol. It was wrong. I had gotten carried away with a custom keymap that I
didn't realize that I was editing the wrong one. Plus I am not a programer,
so I am sure I have many mistakes in the config.

kdb424:
@klardotsh Looking for a final review before merging.

klardotsh:
no strong opinions here, so merging

TellNoLies:
So, I was looking at the traces on my lilly58 to see why my layout wasn't working when I updated my keylayout. It seems that the keys I added to both left and right sides, on the fourth row belongs to the fifth row. I had a look at the tracers on the board and discovered this.
Do I need to submit another pull request?

klardotsh:
you're welcome to push that change upstream in another PR yes!

PR #280, opened 2021-12-02
docs: README to reflect limited maintainership status

klardotsh:


kdb424 (approved):
Completely agree with this.

Issue #222, opened 2021-07-31
Add timeout/swallow API's


kdb424:
There are many keys that are dual use such as tap dance, mod tap, layer tap, ect. We need an API to allow this to become a standard to allow creating keys that can swallow/absorb keystrokes, and process them on timeouts, or even discard them completely. This allows us to restore tapdance, and Leader Mode easily, as well as add new features easily, such as combos.

I have started thinking about this problem, and this is what I have so far. This is open for discussion as no actual code is written, only the pseudo code below.

```python
def swallow(keycode, swallow_timeout=100, suppression=True)
    # keycode: Any valid keycode
    # swallow_timeout: Timeout in ms
    # suppression: allows the action to pass through or not
    timeout(keycode, time_in_ms)
   

def evict(keycode=None, empty=False)
    # keycode: Any valid keycode
    # empty: If all keys stored should be evicted
def revert(keycodesStored, suppression=False):
    # keycodesStored: list of all keycodes that were, or were not suppressed
    if not suppression:
        # Keycodes will need to be backspaced as they were not suppressed
        pass
    else:
        # Keys were suppressed and not typed. Just evict queue
        evict(keycode=None, empty=True)
    
```

## Leader mode

```python
if kc:
    leaderStored.ammend(swallow(keycode=kc, swallow_timeout=0, suppression=True))

if match in leaderStored:
    # do the leader macro or whatever
    revert(keycodesStored=leaderStored, suppression=True)
    
```

## Combos
```python
if kc:
    comboStored.ammend(swallow(keycode=kc, swallow_timeout=0, suppression=False))

if match in comboStored:
    # do the leader macro or whatever
    revert(keycodesStored=comboStored, suppression=True)

elif timeoutReached:
    # pop the oldest key off of the stack and reset timer
    pass

```

# Timeout

Timouts should store an end "timestamp" as well as a keycode.
A callback timer will automatically evict the keycode when the timeout
is reached if it has not already been evicted.

A list of touples should be usable, though other ideas are welcome.
```python
timeout_callbacks = [(KC_NO, timeout_expires), (KC_ESC, timeout_expires)]

def create_timeout(keycode=None, time_in_ms):
    # keycode: any valid keycode
    # time_in_ms: how long to wait
    # returns: timeout(keycode, expires_at_time)
```

xs5871:
would it help to extract the tap dance logic from `kmk_keyboard` and make it its own module?

PR #281, opened 2021-12-02
make TapDance a module

### LINKED ISSUE #279: using mod tap, holding two mods at the same time doesn't work
I use home row mods (so, mod tap), and I cannot hold two of them at the same time.
https://github.com/sadekbaroudi/kmk_firmware/blob/master/boards/ffkb/main.py#L93

For instance, if I want to press control alt delete, it just doesn't work.

What I do is press and hold my A and S keys at the same time (mod tap ctrl alt), and it outputs whichever key I pressed first, and then does nothing.

xs5871:
Probably a necessary or at least helpful step to unifying multi-purpose keys (TapDance, ModTap, etc...), see #222.
It doesn't yet implement the suggested API, but it adds a new module method `Module.process_key` that allows for intercepting and potentially modifying any `Key` before it is handled by `KMKKeyboard.process_key`.
It also became apparent that the `TapDance` code/logic would need a complete rewrite, ideally something similar to `HoldTap` with a proper state machine. Current code is difficult to follow and debug.

* fixes #279

### DIFF
--- a/docs/modtap.md
+++ b/docs/modtap.md
@@ -4,7 +4,10 @@ added to the modules list.
 
 ```python
 from kmk.modules.modtap import ModTap
-keyboard.modules.append(ModTap())
+modtap = ModTap()
+# optional: set a custom tap timeout in ms
+# modtap.tap_time = 300
+keyboard.modules.append(modtap)
 ```
 
 ## Keycodes
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -13,3 +13,4 @@ put on your keyboard
 when tapped, and modifier when held.
 - [Power](power.md): Power saving features. This is mostly useful when on battery power.
 - [Split](split_keyboards.md): Keyboards split in two. Seems ergonomic!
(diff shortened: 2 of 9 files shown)


klardotsh:
haven't had a chance to dive deeply into this but at a very quick glance this looks awesome - thank you for diving in! I'll try to get a proper review in as my wild schedule this week allows.

honboubao:
This is great! Just one thing, HoldTap still references keyboard.tap_time, which has been moved to the new TapDance Module

xs5871:
Huh, how did I miss that? Thanks.

pcharest2000:
Just to let you low I have been running this pull for over 3 weeks and it works great!   It considerably improve usage of the keyboard for modtap... This should really be incorporated in the main tree!!!!   Before this pool I could not use modtap like I did with QMK

kdb424 (approved):
I'm not seeing any major issues other than the failing check. Once that's set, I hope to merge this.

klardotsh:
thanks! this reads fine, and we have confirmed "tested and works" from the community ,so in it goes.

Issue #273, opened 2021-10-28
Support multiple backlight targets via multiple PWM outputs


HMarxen:
One feature I always wished for in QMK was having more than one key backlight zones. Since KMK compatible chips are generally more modern, many also have more PWM pins. So, this feature would make even more sense in KMK. So far, anything mor complex than one backlight zone would need an LED driver IC. With more supported individual PWM outputs, you could get away with a lot more experimentation before you need to add special ICs.

You could for example adjust the brightness of 7-segment displays independent from key BL. Or using an analogue volt meter, the kind with a moving needle, as a typing speed gauge. Or just having the wasd keys and other gaming relevant keys adjusted separately.

I would suggest the following behaviour:
Instead of one pin you can define an array of PWM pins.
KC.LED_INC, KC.LED_DEC and KC.LED_TOG etc. control all PWM pins and sets them back to the same values if they deviate.
KC.LED_INC(x) etc. controls individual PWM pins. X being the array position of the PWM pin.

HMarxen:
The RP2040 has 29 GPIO pins and 16 of those can do PWM. A numpad could have an individually adressable LED in each key. You could make a full light show without even needing Transistors.

TobiasWegener:
> 

I did try it and the RB Pico RP2040 supports sufficient current to support all 16 LED of my NumPad in full brightness, without any additional hardware (apart from 470 Ohm resistors).
But I had to use the CircuitPython internal library  (digitalio) to drive the Leds because I could not get the KMK implementation running.
Maybe I am to stupid to understand the documentation [LED](https://github.com/KMKfw/kmk_firmware/blob/master/docs/led.md) but I find it pretty limited and can't get it to work.

Is there a simple and complete code example for mono colored Led, that works out of the box for a Pico with Circuitpython 7.0 and up-to-date kmk_firmware? Thanks a lot.

xs5871:
I opened a merge PR (#283) for this.
@TobiasWegener we need to migrate the PWM interface from pulsio to pwmio to support Circuitpython 7.0. That is also addressed in the PR.

xs5871:
This issue can be closed.

PR #283, opened 2021-12-04
Multiple led pwm targets (see #273)

xs5871:
LED module argument `led_pin` accepts pin objects or lists of pin objects (free backwards compatibility).
Toggle, increase, and decrease are promoted to argumented keys (sadly not backwards compatible). Didn't yet figure out if we can make a `maybe_argumented_key`.
Increasing/Decreasing brightness of all LEDs doesn't reset brightness levels, but steps each LED brightness by the provided step value.

Pending:
* add `KC.LED_SET(value, *leds)` key
* documentation

usage example:
```python
# single LED
rgb_ext = RGB(
    led_pin = board.GP20
)
# multiple LEDs
rgb_ext = RGB(
    led_pin = [board.GP20, board.GP21, board.GP22]
)

# toggle all LEDs
LED_TOG_ALL = KC.LED_TOG()
# increase brightness of first LED
LED_INC_0 = KC.LED_INC(0)
# decrease brightness of second and third LED
LED_DEC_1_2 = KC.LED_DEC(1,2)
```

### DIFF
--- a/kmk/extensions/led.py
+++ b/kmk/extensions/led.py
@@ -9,6 +9,7 @@ from kmk.utils import clamp
 class LEDKeyMeta:
     def __init__(self, *leds):
         self.leds = leds
+        self.brightness = None
 
 
 class AnimationModes:
@@ -73,6 +74,11 @@ class LED(Extension):
             names=('LED_DEC',),
             validator=self._led_key_validator,
             on_press=self._key_led_dec,
+        )
+        make_argumented_key(
+            names=('LED_SET',),
+            validator=self._led_set_key_validator,
+            on_press=self._key_led_set,
         )
         make_key(names=('LED_ANI',), on_press=self._key_led_ani)
         make_key(names=('LED_AND',), on_press=self._key_led_and)
@@ -220,6 +226,11 @@ class LED(Extension):
 >>>>>>> c7bdb7c (apply suggested changes)
         return LEDKeyMeta(*leds)
 
+    def _led_set_key_validator(self, brightness, *leds):
+        meta = self._led_key_validator(*leds)
+        meta.brightness = brightness
+        return meta
+
     def _key_led_tog(self, *args, **kwargs):
         if self.animation_mode == AnimationModes.STATIC_STANDBY:
             self.animation_mode = AnimationModes.STATIC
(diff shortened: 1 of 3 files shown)


kmk/extensions/led.py
```diff
@@ -1,9 +1,18 @@
...
 
 from math import e, exp, pi, sin
 
 from kmk.extensions import Extension, InvalidExtensionEnvironment
-from kmk.keys import make_key
+from kmk.keys import make_argumented_key, make_key
+
+
+def _clamp(x):
+    return min(max(0, x), 100)
```
klardotsh, line 10: nit: the max of the clamp should be configurable and this moved to a `kmk/utils.py` for re-use elsewhere, this is the type of thing that's likely to be useful in other places too

```suggestion
def _clamp(x, top = 100):
    return min(max(0, x), top)
```

kmk/extensions/led.py
```diff
@@ -26,12 +35,16 @@ def __init__(
         user_animation=None,
         val=100,
     ):
+        self._leds = []
         try:
-            self._led = pulseio.PWMOut(led_pin)
+            if not isinstance(led_pin, list):
+                led_pin = [led_pin]
+            for pin in led_pin:
+                self._leds.append(pwmio.PWMOut(pin))
```
klardotsh, line 43: to be compatible with generators, this could use a try/except of trying to iterate over `led_pin`, and on failure, assign the iterator to `[led_pin]`, a-la https://stackoverflow.com/a/1952655

```suggestion
            try:
                pins_iter = iter(led_pin)
            except TypeError:
                pins_iter = [led_pin]
                
            for pin in pins_iter:
                self._leds.append(pwmio.PWMOut(pin))
```

kmk/extensions/led.py
```diff
@@ -191,17 +212,23 @@ def animate(self):
         else:
             self.off()
 
+    def _led_key_validator(self, *leds):
+        if leds is not ():
```
klardotsh, line 216:
```suggestion
        if leds:
```

klardotsh (commented):
this is a really good start, I like this.

kdb424:
Looks like there is a linter error, but otherwise things look correct.

xs5871:
If there's no other objection, this should be ready to merge.

kdb424:
Looks like there's an error in the testing, otherwise, this looks good to merge.

kdb424 (approved):

Issue #286, opened 2021-12-05
Mouse keys do not work on PICO


pcharest2000:
KMK:  Latest pull from 05-12-2021
Circuit Python: 7.0
Hardware: Pi Pico

The mouse keys do not work at all, debug log:

KeyResolution(key=Key(code=1006, has_modifiers=None))
Failed to run pre hid function in module:  Buffer incorrect size. Should be 8 bytes. <MouseKeys object >
KMKKeyboard(debug_enabled=True diode_orientation=0 matrix_scanner=<class 'MatrixScanner'> unicode_mode=)
Failed to run pre hid function in module:  Buffer incorrect size. Should be 8 bytes. <MouseKeys object >
Failed to run pre hid function in module:  Buffer incorrect size. Should be 8 bytes. <MouseKeys object >

Does not seem to be a Circuit Python issue since the following example works:
https://tutorial.cytron.io/2021/07/06/diy-usb-keyboard-mouse-using-maker-pi-pico/

klardotsh:
should be fixed in master thanks to #289, feel free to reopen if it's still an issue

PR #289, opened 2021-12-05
fix hid report type disambiguition

xs5871:
addresses #286.

`hid_send` was always looking at the keyboards report when figuring out the device, instead of looking at the report it's supposed to send.
Tested with USB HID, needs to be verified with BLE before merging.

### DIFF
--- a/kmk/hid.py
+++ b/kmk/hid.py
@@ -227,7 +227,7 @@ class USBHID(AbstractHID):
 
     def hid_send(self, evt):
         # int, can be looked up in HIDReportTypes
-        reporting_device_const = self.report_device[0]
+        reporting_device_const = evt[0]
 
         return self.devices[reporting_device_const].send_report(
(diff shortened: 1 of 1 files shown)


klardotsh (approved):
if someone has BLE boards around that can test this, I'm cool with a merge after a quick verification. maybe @honboubao if time allows?

pcharest2000:
> if someone has BLE boards around that can test this, I'm cool with a merge after a quick verification. maybe @honboubao if time allows?

no ble board on my side but confirmed resolving mouse keys !

klardotsh:
gonna yolo merge. if someone hollers in matrix we'll know BLE got broken but I think this is probably for the greater good. according to the blame history this line predates BLE so we're at least not *explicitly* regressing.

honboubao:
for what it's worth, I did do a quick test with BLE just now, and mouse keys seem to work fine with the fix

PR #291, opened 2021-12-06
Mousekeys

pcharest2000:
Added clarification in keys.md regarding returning a boolean to the attached handler/callback


kdb424:
I'm a bit confused at what this PR is at this point with the title. Can we get more clarification as I see several files that seem unrelated touched.

pcharest2000:
> 

I apologize I am not very familiar with the git hub pull process and believe I have screwed up.    It seems they are all combined into one.  One pull was for mouse acceleration implementation....  issue 285 is already in the master tree...  these are all minor pull thus u can just ignore them!   Again sorry.

kdb424:
So git works on a branch system, so in order to take this PR, I'd have to take all changes in this branch. It will need cleaned up before a merge can happen.

kdb424 (approved):

PR #296, opened 2021-12-22
tweaks to the Getting_started page.

iraytrace:
Minor re-wording of Getting_Started.md page code example to provide hint a type of modification to make for two example boards.  Show picture of pins for those two boards to help novice user find them properly.


klardotsh (approved):
thank you!

PR #297, opened 2021-12-31
Fix two way uart

wildesteven:
These are the changes I made to use two way UART on my RP2040 Corne.  I added a check on `uart_flip` to support crossing TX and RX in hardware.  I added a header byte to the front of each UART message to confirm that the pin is not floating.  I only apply the `split_offset` when sending from right to left.

### DIFF
--- a/kmk/modules/split.py
+++ b/kmk/modules/split.py
@@ -53,6 +53,7 @@ class Split(Module):
         self._uart = None
         self._uart_interval = uart_interval
         self._debug_enabled = debug_enabled
+        self.uart_header = bytearray([0xB2])  # Any non-zero byte should work
         if self.split_type == SplitType.BLE:
             try:
                 from adafruit_ble import BLERadio
@@ -289,6 +290,8 @@ class Split(Module):
                 update[1] -= self.split_offset
 
         if self._uart is not None:
+            # Add a header byte the data
+            self._uart.write(self.uart_header)
             self._uart.write(update)
 
     def _receive_uart(self, keyboard):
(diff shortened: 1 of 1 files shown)


klardotsh (changes requested):
almost all of this appears to be sane, however, this appears to regress the case where the right side of a split is the primary - eg the side plugged into the actual PC. this was previously handled (not sure if correctly, but _handled_, in theory) by `self._is_target`. is it possible to refactor this to preserve EEHands functionality?

wildesteven:
I'm not sure what you mean.  It seems to work fine for me with either side as the primary, with or without EEHands.

Tonasz:
What test case do you @klardotsh mean?
I have 2x Pico connected through USB (for serial port debug), and with GND and RX -> TX, TX -> RX connected and tested these situations:
1) both sides
`Split(split_side=SplitSide.RIGHT,data_pin=board.GP1,data_pin2=board.GP0,uart_flip=False,split_target_left=False)`
2) both sides
`Split(split_side=SplitSide.RIGHT,data_pin=board.GP1,data_pin2=board.GP0,uart_flip=False,split_target_left=True)`
(i checked that the working half really changed by checking encoder input - only main side can produce it currently)
3) `Split(split_side=SplitSide.LEFT,data_pin=board.GP1,data_pin2=board.GP0,uart_flip=True,split_target_left=True)` + `Split(split_side=SplitSide.RIGHT,data_pin=board.GP0,data_pin2=board.GP1,uart_flip=True,split_target_left=True)`
4) `Split(split_side=SplitSide.LEFT,data_pin=board.GP0,data_pin2=board.GP1,uart_flip=True,split_target_left=False)` + 
`Split(split_side=SplitSide.RIGHT,data_pin=board.GP1,data_pin2=board.GP0,uart_flip=True,split_target_left=False)`
And everything worked.
Bonus tests on my fork with custom UART that allowed me connecting RX -> RX, TX -> TX on RP2040 and I could make two-way communication in analogous situations (just adjusted passed pins).

kdb424:
Closing in favor of #301. Thank you for your contribution!

PR #298, opened 2022-01-09
7 segment display

daysgobye:


kdb424 (approved):
Leaving to @klardotsh for final review if he wants to do so, otherwise I'll merge if tests are clean.

kdb424:
`make fix-isort` I think will resolve those issues.

daysgobye:
sooo Now there are lots of changes but should be good

kdb424:
It was a bit overreaching, but seems like they needed done anyways, and our makefile is overreaching. Yolo if it passes. For future tests `make test` will show you issues before a commit.

klardotsh:
I'd love for this to be a generic driver/extension and not hard-coded to be layers-related, but I don't have the brain space to help design up such an API right now :shrug:

kdb424:
Maybe worth changing the name to indicate what it does specifically? Having modules be specific can be a good thing to keep things simple, especially for things as simple as a seven segment display, but have different modules to load based on what you want it to do?

kdb424:
Closed as requested.

PR #300, opened 2022-01-16
Remove serial log errors when declaring encoders without a button

Tonasz:
I tried using encoders on my keyboard, which don't have their button pins connected separable, but are part of matrix. So I initialize them with `pin_button=None`. 
When I was doing that change, I thought that encoder is not working without it, because it was spamming the serial console with: `Failed to run pre matrix function in module:  'NoneType' object has no attribute 'get_value' <EncoderHandler object at 0x20009730>`.
I now realized it was working all the time, but this change remove the errors from log.


klardotsh (approved):
thank you!

PR #301, opened 2022-01-18
Split code fix/cleanup.

patrickmcquay:
…, as well as make it work nicely with ble.

@wildesteven I've taken your changes a bit further, and (I think) fixed some issues with the way that _is_target is decided in the split module, with help from @Tonasz. I would welcome a review, and hope I'm not stepping on too many toes.

I also added a bit more logging to kmk_keyboard to debug this, and considering that the rest of it does the same thing I think its relevant.

I don't know about using runtime.usb_connected to determine if usb is connected. The docs say that that is what it does but Tonasz said he thinks that it might be determining if the usb drive is enabled.

I also rearranged the order of the if block that decides what _is_target and split_side are, because starting an if block with a negative can be confusing to read. (was for me, I was confused about what order or precedence the types of connection happen in)

### DIFF
--- a/kmk/modules/split.py
+++ b/kmk/modules/split.py
@@ -92,7 +92,10 @@ class Split(Module):
             self._is_target = not bool(self.split_target_left)
         else:
             # Detect split side from name
-            if self.split_type == SplitType.UART or self.split_type == SplitType.ONEWIRE:
+            if (
+                self.split_type == SplitType.UART
+                or self.split_type == SplitType.ONEWIRE
+            ):
                 self._is_target = runtime.usb_connected
             elif self.split_type == SplitType.BLE:
                 self._is_target = bool(self.split_target_left)
--- a/kmk/modules/split.py
+++ b/kmk/modules/split.py
@@ -1,7 +1,7 @@
 '''Enables splitting keyboards wirelessly or wired'''
 import busio
 from micropython import const
-from supervisor import ticks_ms
+from supervisor import ticks_ms, runtime
 
 from storage import getmount
 
@@ -49,10 +49,11 @@ class Split(Module):
         self.data_pin2 = data_pin2
         self.target_left = target_left
         self.uart_flip = uart_flip
-        self._is_target = True
         self._uart = None
         self._uart_interval = uart_interval
         self._debug_enabled = debug_enabled
+        self.uart_header = bytearray([0xB2])  # Any non-zero byte should work
+
         if self.split_type == SplitType.BLE:
             try:
                 from adafruit_ble import BLERadio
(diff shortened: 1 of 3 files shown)


kdb424:
Supersedes #297 and cleans code up, along with preserving EE_HANDS functionality. Allowing @klardotsh to make final call, but I prefer this implementation.

kmk/modules/split.py
```diff
@@ -84,22 +85,25 @@ def during_bootup(self, keyboard):
...
             self._is_target = not bool(self.split_target_left)
+        else:
+            # Detect split side from name
+            if (
+                self.split_type == SplitType.UART
+                or self.split_type == SplitType.ONEWIRE
+            ):
+                self._is_target = runtime.usb_connected
+            elif self.split_type == SplitType.BLE:
+                self._is_target = bool(self.split_target_left)
```
Tonasz, line 101: I don't have any BLE controllers to test if it will work, but I see here a change which potentially could break BLE. If split_side is not declared, but recognized from drive name, both BLE sides will get the same _is_target value instead of pair of opposite values. Am I missing something?
```suggestion
                self._is_target = name.endswith('L') == self.split_target_left
```

Tonasz (commented):
I proposed using runtime.usb_connected, but was not sure if it will work in every situation. I checked and booted right side with `storage.disable_usb_drive()` and `usb_cdc.disable()` and it still worked, so I think we're good. And as you said, documentation confirms.
I like this improvement of @wildesteven PR and don't see major flaws (except one concern I commented).

kmk/modules/split.py
pmcquay: My code replicates the existing functionality. As far as I can see you arent missing anything. Im not not sure what to do in this case. Your code assumes that the ble user will want ee hands, but what happens if they dont have the drives labelled?

pmcquay: My code replicates the existing functionality. As far as I can see you arent missing anything. Im not not sure what to do in this case. Your code assumes that the ble user will want ee hands, but what happens if they dont have the drives labelled?

pmcquay: Honestly I think the better thing to do is test for these things when initializing and bail out with an error if nothing is set.

patrickmcquay: now that I can look at this on a bigger screen, Your suggestion makes sense. I just couldnt figure out what was going on there.

patrickmcquay:
seems reasonable, I haven't gotten errors like that though, so I'm not sure how to test it. I'll take a whack at it though.

xs5871:
For testing: you could send intentionally bad data.

patrickmcquay:
I've implemented a basic checksum here. I don't think it covers all cases, but it definitely ignores bad keys, (yup, half my keyboard wasn't working for a while, lol). If the checksum equals the header you might lose a couple keys instead of just one, for example but I think that's not unreasonable.

kdb424:
@patrickmcquay Is this ready for full review?

pmcquay:
It is. I've been daily driving this code since it was written and had no issues.

kdb424 (approved):

kdb424:
Just need the conflicts to be resolved now that we have merged other PR's in, and we'll get it merged.

pmcquay:
It was formatted incorrectly in another commit, I needed to fix it to pass tests. Probably dont need it now, but whatever.

pmcquay:
Awesome, ill get it rebased and sorted out later tonight.

patrickmcquay:
not sure how the problem with encoder.py got into master

kdb424:
Fixed and merged

PR #302, opened 2022-01-19
implement keyboard lock status reader API

xs5871:
This extension exposes host-side locks like caps or num lock.
example:
```python
import digitalio

from kmk.extensions.lock_status import LockStatus

locks = LockStatus()
keyboard.extensions.append(locks)

led_clock = digitalio.DigitalInOut(board.GP25) # change to pin of LED
led_clock.direction = digitalio.Direction.OUTPUT

# somewhere in a loop:
if locks.get_caps_lock():
    led_cap_lock.value = 1
else:
    led_cap_lock.value = 0
```

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -9,6 +9,7 @@ extensions are
 
 - [International](international.md): Adds international keycodes
 - [LED](led.md): Adds backlight support. This is for monocolor backlight, not RGB
+- [LockStatus](lock_status.md): Exposes host-side locks like caps or num lock.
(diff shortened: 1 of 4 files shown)


xs5871:
Why didn't the check fail for PR #300?

kdb424:
> Why didn't the check fail for PR #300?

I'm not quite sure. `make fix-formatting` will solve it automatically though.

klardotsh (approved):
yeah, this is awesome. thank you!

Issue #278, opened 2021-12-01
Permissive hold


sadekbaroudi:
First and foremost, great project! I'm very pleased with how quickly and easily you can write firmware!

I am a fairly fast typist, and I use home row mods. Without a feature like permissive hold it is very hard for me to use any keyboard using kmk.

If I set the tap_timeout too low (100 ms or below), I trigger the hold while typing. If I set it longer, I have to slow down my typing when using any mods or layers to a point of being very inconvenient.

When using QMK or ZMK, they have a mechanism that allows me to resolve this.

This works perfectly in QMK:
https://beta.docs.qmk.fm/using-qmk/software-features/tap_hold#permissive-hold

ZMK has a clear outline of how they select which:
https://zmk.dev/docs/behaviors/hold-tap/

Both of these allow you to configure it per key as well. For instance, I need the shift to be very fast, and the others can be a bit slower.

klardotsh:
I personally don't use this functionality in other firmwares so I likely won't be taking this up myself, but I'll leave the ticket open in case someone else is perusing through issues looking for inspiration some time and wants to submit a pull request.

xs5871:
_Just_ hold-preferred / tap-preferred, asignable per key, is a fairly straight-forward implementation.
We could extend the validator signature by a `behavior` stored in meta and do something like this for configuration:
```python
from modules.holdtap import InterruptBehavior
FASTSHIFT = KC.MT(KC.X, KC.LSFT, behavior=InterruptBehavior.HOLD)
```
Then, just launch the corresponding handler in `ht_activate_on_interrupt` based on the `InterruptBehavior` enum.
 
It would be nice to also have the other hold-tap flavors as layed out by the ZMK reference. Those are somewhat more involved.

xs5871:
The HoldTap enhancements solve this issue.

Issue #256, opened 2021-10-15
Firmware Hang Up at using ModTap on a split keyboard


yoichiro:
I'm a keyboard designer of the Lunakey Pico which is a split keyboard. When tapping a key with the ModTap setting on the right side which the USB cable is not connected, the firmware on the right side is hanged up.

For example, I set the following ModTap keycode on the right side (that is, I use the [main.py](https://github.com/KMKfw/kmk_firmware/blob/master/boards/lunakey_pico/main.py) ):

```
KC.MT(KC.MINS, KC.LSHIFT)
```

Then, when tapping the key, the `-` character is sent, but at the same time, the following error occurs:

```
Traceback (most recent call last):
  File "main.py", line 120, in <module>
  File "kmk/kmk_keyboard.py", line 497, in go
  File "kmk/kmk_keyboard.py", line 115, in _send_hid
  File "kmk/hid.py", line 128, in send
  File "kmk/hid.py", line 233, in hid_send
OSError: USB busy
```

I seems that the firmware tried sending hid report at tapping the key set as ModTap on the right side. However, the USB cable is not connected on the right side. Instead, the TRRS cable is connected on the right side to communicate with the  left side using an UART.

The photo of the Lunakey Pico is below:

![PXL_20211015_061214796](https://user-images.githubusercontent.com/261787/137440720-24b2097c-2737-4fad-b863-c6e82c170675.jpg)

In my keyboard, this issue occurs 100% on the right side. But, this issue never occur on the left side because the USB connection established.

yoichiro:
Feel free to let me know if you want to add a code for debug. I can try it in my keyboard.

PR #304, opened 2022-01-24
hold-tap-enhancements

xs5871:
This has been a requested feature (#278), and after using it for 2 days, rightfully so.
Just wanted you to know that this is being worked on.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -7,7 +7,7 @@ from kmk.modules.layers import Layers
 keyboard.modules.append(Layers())
 ```
 
- ## Keycodes
+## Keycodes
 
 |Key         |Description                                                                  |
 |-----------------|------------------------------------------------------------------------|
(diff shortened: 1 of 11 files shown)


kdb424:
I like where this is going! Let us know if you need anything specific, but thank you for looking into this and testing. I'll try to get it tested as well on my boards.

xs5871:
I think screwed up merging the master branch, sorry.
Anyway, last commit implements evaluation of hold-tap behavior on other key release, which seems to be the default in QMK/ZMK (for reference).
`prefer_hold` and `tap_interrupted` options aren't working with the layer module yet, as the interrupting key gets its keycode assigned before the layer switch is performed.

kdb424:
Let us know when you are ready for review, and we'll get it looked at.

xs5871:
The Layer module now also supports `prefer_hold` and `tap_interrupted`; more specifically layer tap/toggle (TT) uses `prefer_hold` as defaut instead of overloading `ht_(de)activate_on_interrupt`.
The `tap_interrupted` logic may be potentially controversial. It relies on suppressing keys and on release retroactively adding them to `keyboard._coordkeys_pressed`. That ensures consistent behavior even when the keys change identity after a layer change.

However: one of the commits ("fix permanent hid_send") currently regresses on #256/#257. I'd rather suggest a "more proper" fix in a separate PR addressing this issues than revert the commit. The fix is already implemented, I'll reference it here once published. That also means there's no need to review at this point in time.

Documentation will be updated next. Tapdance is predestined to use the same hold-tap backend, but is definitely going to be a separate feature branch.

xs5871:
Should be ready for review, finally.

kdb424 (approved):
I'll give @klardotsh a day or two to refute anything here, otherwise looks good to me.

PR #311, opened 2022-01-31
fix for dropping keys

patrickmcquay:
it is possible to get through both of those if statements without hitting either one, especially on a split keyboard. After that this routine will use the previous key (which means that both the next keystroke and the duplicate previous stroke get swallowed). My fix makes it look up the key in the map all of the time.

This may seem heavy handed (I believe the first if block is there as a performance thing, though I cant be sure.), but I believe its the right fix, because it makes the code much simpler to read and reason about, and it doesn't seem to have any real performance impact.


patrickmcquay:
looks like if both keys are pressed, and the current key doesn't get cleared out inbetween then this problem happens. The current key only gets cleared out once per cycle, so because handle_matrix_report gets called twice for split boards, this can happen. and the reason it only drops primary board keys is because of the order of calls to process the keys. Primary keys are processed last.

Another fix for this would be to clear out the current key before running this, but I think that my fix makes the code less confusing and likely to break in other ways. I dont think we need keyboard.current_key anywhere else, so this could be made local and would have avoided the problem entirely. LMK if this makes sense and I can do another commit here for that.

xs5871:
The lookup of _released_ keys is important in case the keymap changes between a press and release, which is likely to happen as soon as you use more than one layer for example. However, a try-except-debug pattern for the lookup would still be better than crashing the entire keyboard.
The `current_key`, once set, shouldn't be changed during a cycle. I agree that resetting it wouldn't be preferrable.

What about sending the matrix updates through a fifo buffer before processing?
```python
if secondary_matrix_update:
    updates.append(secondary_matrix_update)
if matrix_update:
    updates.append(matrix_update)
handle_matrix_report(updates.pop(0))
```

patrickmcquay:
ahhh, that makes sense, I thought it was a performance thing.

Your change is probably more elegant than mine. I really dislike the if not is pressed and then if nothing was set pattern it used before though, its just asking for bugs. Could we move to a if is pressed with an else block there? that way would ensure that at least one of the blocks executes. I also like the fifo idea. Let me put something together.

patrickmcquay:
something like this:

`if is_pressed:
            self.current_key = self._find_key_in_map(int_coord, row, col)
        else:
            # key is released, need to look up from previous keys
            # pressed because it could be different due to layers
            self.current_key = self._coordkeys_pressed[int_coord]
            if self.debug_enabled:
                print('PressedKeyResolution(key={})'.format(self.current_key))`

patrickmcquay:
yeah so that doesnt work either, double releases get through there and break things. I've had a couple shots at this now and my solution (flawed as it is) so far is the best I can see. If I implement a fifo buffer I get doubled keypresses and releases, I do not know whats going on there.

What I do know is that this is the place where the problem is, so I'll keep working on it. This PR just isnt the solution.

What is the reasoning for only handling one press per cycle? I'm sure there is one, but it doesnt seem necessary to me.

xs5871:
Is there a specific scenario when this happens, mashing for example?
Have you verified that it's not the scanners that report duplicate events?

There's code that handles timeouts and delayed actions, which is executed once a cycle. More than one key event pre cycle could concievably introduce some ugly race conditions.

patrickmcquay:
yeah, these scenarios are all easily reproducible by mashing, but in real typing they happen as well, just not nearly as often. I dont know what the problem is with a buffer, it certainly could be an interaction like the one that started this, or it could be keyswitch bounce. I'm going to look at it more later.

patrickmcquay:
that was fun. Turns out because we reuse things here, clearing them out and setting them to other things affects the queue that they are put in. That causes all sorts of issues with things.

Making copies was the solution and works correctly. I've also reverted the other change because as long as we are dealing with one key at a time its probably better

xs5871:
Tested on single and split matrix, doesn't break anything, couldn't provoke key event drops.

kdb424:
Will give @klardotsh a chance to review, but looks clean to me.

kdb424 (approved):

klardotsh (approved):
seems sane to me and the thread seems in agreement that this works as designed, so merging. thank you!

klardotsh:
er, not merging, because conflicts, though GitHub doesn't say (on my end at least) what they are. whenever those are resolved, one of us will merge

patrickmcquay:
this branch is pretty mangled, and I didnt understand that kmk uses rebase instead of merge. I'll just trash it and recreate.

patrickmcquay:
continued in #317. Please trash this PR

PR #288, opened 2021-12-05
unit tests for execution in desktop dev environment

honboubao:
I have added a few unit tests that can be executed without having a board at hand. There're still many features that aren't covered yet, but I guess it's a start.

PR #312, opened 2022-01-31
Scanner interface rework

xs5871:
This replaces the current matrix scanners update object.
Instead of the implementation specific bytearray, there's a `KeyEvent` object inspired by circuitpythons version for their `keypad` library. Main difference to the current implementation is that keys are identified by a single number, instead of row and column.
Outstanding efforts to provide more/different scanners would benefit from this abstraction.

The code has been tested with a monolithic keyboard, and a split using UART. I don't expect there to be issues with BLE, but someone with a wireless split has to confirm.

The only breaking change is in the signature of `intifiy_coordinates`:
```python
intifiy_coordinates(row, column, number_of_columns)
```
I updated the only appearance of ic in the documentation, but I don't own a corne, so I'm half-guessing.
This is also somthing I'd appreciate help verifying.

For reference: this is more or less complimentary to #262. It's been a while since the last update though, and my previous branch is now stale, messy and broke split communication. I understand that this is a big, singular commit, but it's almost impossible to make it more atomic.

### DIFF
--- a/boards/boardsource/microdox/kb.py
+++ b/boards/boardsource/microdox/kb.py
@@ -15,10 +15,15 @@ class KMKKeyboard(_KMKKeyboard):
     i2c = board.I2C
     powersave_pin = board.P0_13
 
+    # NOQA
     coord_mapping = []
-    coord_mapping.extend(ic(0, x) for x in range(10))
-    coord_mapping.extend(ic(1, x) for x in range(10))
-    coord_mapping.extend(ic(2, x) for x in range(10))
+    coord_mapping.extend(ic(0, x, 5) for x in range(5))
+    coord_mapping.extend(ic(4, x, 5) for x in range(5))
+    coord_mapping.extend(ic(1, x, 5) for x in range(5))
+    coord_mapping.extend(ic(5, x, 5) for x in range(5))
+    coord_mapping.extend(ic(2, x, 5) for x in range(5))
+    coord_mapping.extend(ic(6, x, 5) for x in range(5))
 
     # And now, to handle R3, which at this point is down to just six keys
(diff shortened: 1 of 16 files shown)


klardotsh:
if @honboubao (BLE) and @kdb424 (corne) are able to provide manual testing of this, rad, otherwise, this passes my style/flow review and is what I'd consider "fine to merge when tested by at least one other user, or (now that #288 just merged) tests are added to prove it works" (which would be a @xs5871 thing, I imagine)

klardotsh:
@kdb424 has merge rights so can green-check and merge this without my further input if they so desire

kdb424:
Not only will the docs have to be updated, but every supported board needs to as well for this to be merged, or anyone using boards like this (lily58, corne, Iris, and more), will be completely broken. Once all of the boards with a non standard matrix are reworked, I will review this.

xs5871:
That's a very good point. Is there a way to change this into a draft for the time being? It'll take me a while to go through all of this.

xs5871:
All boards are ported. There was a mismatch between physical keyboard / keymap layout and `coord_mapping` on the number of keys with two boards. I guessed the correct `coord_mapping` and added a NOQA comment.
I also added a flat `coord_mapping` to the documentation that may be easier to read and understand. The BR translation is computer generated, sorry.

Now to figure out the unit test...

xs5871:
unit test already coming in handy

xs5871:
Verified as working with BLE split with gratitious help from @Tonasz.

kdb424 (approved):
Thank you both @xs5871 and @Tonasz for your work and testing out code <3

PR #313, opened 2022-02-02
fix bytearray.startswith not available in CircuitPython

honboubao:
So sorry for introducing an operation breaking bug with #288 

This will hopefully fix the issue.

@xs5871 thanks for bringing it up. Could you verify, if this fixes things and whether you notice any other problems?

I currently don't have hardware set up to test properly, which is why I added the tests in the first place. And probably relied a bit too much on those

### DIFF
--- a/kmk/hid.py
+++ b/kmk/hid.py
@@ -126,8 +126,7 @@ class AbstractHID:
         pass
 
     def send(self):
-        changed = not self._evt.startswith(self._prev_evt)
-        if changed:
+        if self._evt != self._prev_evt:
             self._prev_evt[:] = self._evt
             self.hid_send(self._evt)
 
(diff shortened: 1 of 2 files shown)


xs5871:
yes, KMK no longer crashes.
Out of curiosity: why introduce that check in the first place? It doesn't appear to be necessary for the unit test suite.

honboubao:
thanks for checking!

If you change it to if True, the tests won't pass anymore. The tests check the HID messages being sent and assume that HID messages are only sent whenever there are changes to be reported.

Contrary to that, #257 introduced a change which (when using modtap) causes a HID message to be sent with each main loop iteration, even if the message is the same as in the previous iteration.

The check is intended to prevent these extra messages and restore the behaviour to something more like before #257.

xs5871:
That is actually very good to know. I removed that constant hid_pending = True in the hold-tap-enhancements draft PR.
Thanks for the heads-up.

kdb424 (approved):
Thanks for unbreaking. We should have noticed before merging, but these things happen.

PR #316, opened 2022-02-03
fix lock_status on secondary split

xs5871:
The lock status extionsion is throwing exceptions on split secondaries.
Turns out you can't recieve and read HID reports if there's no HID. May need further refactoring with bi-directional split communication, once implemented.


klardotsh (approved):
seems reasonable to me. thanks yet again!

PR #318, opened 2022-02-05
Fixed multiple encoders not working

FakeAnimeBoy:
Multiple encoders didn't work as a result of an unbound value of 'idx', the loop counter in the during_bootup method. Binding this value has fixed the issue. This has been tested on hardware (6 encoders) and is confirmed working.


klardotsh (approved):
just taking your testing at face value since I don't use _any_ encoders, let alone multiple. thank you, and congrats on your first kmk pr merged!

Issue #198, opened 2021-04-30
Change UART pins definition


Max585t:
This would allow for better split keyboard support. Currently, when using Raspberry Pi Picos in split keyboards, there is no way to have either side be the target because of the way Rx and Tx pins are set. 

The code would change from
```python
def init_uart(self, pin, timeout=20):
        if self.is_target:
            return busio.UART(tx=None, rx=pin, timeout=timeout)
        else:
            return busio.UART(tx=pin, rx=None, timeout=timeout)
```
to
```python
def init_uart(self, pin, timeout=20):
        if self.is_target:
            return busio.UART(tx=None, rx=pin[1], timeout=timeout)
        else:
            return busio.UART(tx=pin[0], rx=None, timeout=timeout)
```

The users keyboard definition, you would have to define the uart pins as such:
```python
keyboard.uart_pin = (board.TX, board.RX)
```

kdb424:
If this was tested on actual hardware, feel free to PR this. We can validate against other hardware to ensure it won't break other boards.

Max585t:
I have tested this on a pi pico. I would like to make a PR but I cannot publish my branch

kdb424:
I'll see about getting a PR then based on this. I have 2 Pro Micro 2040's coming to test this now as well. Thank you for solving this.

Tonasz:
Isn't this issue solved as a byproduct of https://github.com/KMKfw/kmk_firmware/pull/301? It changed target assignment and added `uart_flip` parameter (disables unnecessary pins swap when False)

PR #320, opened 2022-02-06
Add RP2040 PIO UART implementation for split keyboards

Tonasz:
This change adds alternative PIO implementation of UART. The story behind it is that I learned the hard way that RP2040 is not able to use default UART implementation (I don't know if it is circuit python implementation fault or general issue) on other pins than TX/RX and these cannot be reversed. PIO subprogram goes around these issues.

I understand concern that this is putting RP2040 specific workaround into general repository. On the other hand, as far ias I know, this is also RP2040 specific problem, so solution IMHO can also be RP2040 specific.

Example use cases (mostly for pre-designed PCB):
- UART connected to different pins than TX/RX
- UART connection not crossed (RX->TX / TX->RX)
- one pin connection (like mine); KMK doesn't have onewire based split, so I can at least have one way UART connection


xs5871:
That's a great start for a one-wire protocol.

Here's my issue with that topic:
Just to clarify, and that relates to #198, UART is working perfectly fine, you can route it to any GPIO that the MCU supports hardware UART for (I don't know of any MCU were this would be different, but then again, I don't know every MCU).
If you're trying to connect TX->TX and RX->RX, that will never work. That's not how UART works. In that sense the default implementation in KMK is the wrong way around (see #198).
I acknowledge that a one- or two-way protocol over the same physical pin is useful, but that's not a UART, and we should make that distinction to avoid confusion.

Tonasz:
As I commented on #198 issue, I think it's obsolete because of #301 PR.
Because some of discussion was on discord, I would like to paste what @kdb424 said, so it's not lost in limbo:

_"Any other chip (NRF52840) And M4 chips on Adafruit controllers would let me put TX and RX almost anywhere without a care in the world. RP2040 is the exception, not the rule.
The reason it flipped the TX and RX pins, is because that's how boards designed for QMK used to work when they used a 2 wire split, and therefor, we wrote what code we had to for the hardware that existed (with adapter boards)"_

So I don't see a reason why not to just keep it as it, as use PIO as RP2040-specific solution for RP2040-specific problems. I think that changing names to be more specification correct, would make them counterintuitive, but I'm not attached to them.

Tonasz:
Anyway, I see possible improvement here, but **I'm not sure if it should be part of this PR or other**. I could get rid of all the `if self.split_type == SplitType.X` cluttering the code and `use_pio` flag by replicating the pattern from `kmk_keyboard.py` and `hid.py` files (`KMKKeyboard` assign right `USBHID` implementation class to `_hid_helper` class on init)

```
AbstractSplitProtocol
└─ UartProtocol
│   └─ UartPioProtocol
└─ BleProtocol

xs5871:
Seperating the transport implementation from the split module is something we should look into, we can do that in another feature branch.
On the same note: the pio code doesn't really fit into kmk/handlers -- those are handlers for keycodes. I'd rather see it in a new folder, like "transports".

xs5871:
Is there a necessity to implement the missing timeout handling? We're reading only once there are enough packets buffered to potentially constitute a full message, so I guess not. Correct me if I'm missing something.

Tonasz:
I moved `pio_uart.py` to `kmk/transports`. I felt it could be in wrong place, so I'll go with suggestion :)

I think that because timeout situation is highly unlikely because, as you mentioned, we're waiting for buffered packets. I don't know how this work on lower level and if it is possible to get "stuck between bytes" while reading. Theoretically it should be implemented in PIO handler, but because I kinda shamelessly remixed existing examples, I'm not sure how this should work (something in assembler layer or just resetting PIO by python handler).

xs5871:
Hm. The PIO can definitely get stuck if it doesn't get the data it's expecting (from recent experience), but I don't think this would happen with this specific implementation. It's all based on timing and delays, so worst case it should give the wrong data.
Transmission consistency is checked a layer above, we should be fine.
To answer your question: you can't do the timeout logic in PIO assembly. And doing it in the python handler is kind of pointless, because it would basically be the same check as we're already doing in the split code.

kdb424 (approved):
Will trust the other reviewer. Will merge on test completion.

PR #321, opened 2022-02-06
Pimoroni trackball

Tonasz:
I couldn't find Pimoroni Trackball implementation so I made one. Clicking the ball acts like a left mouse button and movement translates to - depending on selected mode (`TB_MODE` keycode switches the mode on hold):

- mouse movement with light acceleration
- vertical scroll movement

Module is small and working, but there could be some remaining concerns:

1. I receive a lot of "non-lethal" errors when using it through `bitbangio` and sometimes keyboard-hanging errors when using `busio`. They occur only while moving the ball. I thought it was my fault, but @xs5871 tried my code, confirmed it worked with `busio` and AFAIK he got zero errors - we suspect I got some electrical issues on bus, which I'm not able to solve right now, but they are plausible as I had some wiring issues in the past. So that's something to confirm by other users.
2. I couldn't pass larger mouse movement in hid message than 127. It's definitely too small on 1440p screen. I don't know where the limit is set (report_x and report_y has only 1 byte each and HID_REPORT_SIZES cannot be increased), or if I'm passing this value to HID wrong. Or if I will have to split larger movement to few, consecutive, smaller messages.
3. I guess current hid implementation also doesn't have an option for horizontal scroll movements. Any tips how to achieve that?

### DIFF
--- a/kmk/modules/pimoroni_trackball.py
+++ b/kmk/modules/pimoroni_trackball.py
@@ -1,0 +1,205 @@
+'''
+Extension handles usage of Trackball Breakout by Pimoroni
+Product page: https://shop.pimoroni.com/products/trackball-breakout
+'''
+
+import struct
+import math
+from micropython import const
+from kmk.modules.mouse_keys import PointingDevice
+from kmk.keys import make_key
+
+I2C_ADDRESS = 0x0A
+I2C_ADDRESS_ALTERNATIVE = 0x0B
+
+CHIP_ID = 0xBA11
+VERSION = 1
+
+REG_LED_RED = 0x00
+REG_LED_GRN = 0x01
+REG_LED_BLU = 0x02
+REG_LED_WHT = 0x03
+
+REG_LEFT = 0x04
+REG_RIGHT = 0x05
+REG_UP = 0x06
+REG_DOWN = 0x07
+REG_SWITCH = 0x08
+MSK_SWITCH_STATE = 0b10000000
+
+REG_USER_FLASH = 0xD0
+REG_FLASH_PAGE = 0xF0
+REG_INT = 0xF9
+MSK_INT_TRIGGERED = 0b00000001
+MSK_INT_OUT_EN = 0b00000010
+REG_CHIP_ID_L = 0xFA
+RED_CHIP_ID_H = 0xFB
+REG_VERSION = 0xFC
+REG_I2C_ADDR = 0xFD
+REG_CTRL = 0xFE
+MSK_CTRL_SLEEP = 0b00000001
+MSK_CTRL_RESET = 0b00000010
+MSK_CTRL_FREAD = 0b00000100
+MSK_CTRL_FWRITE = 0b00001000
+
+ANGLE_OFFSET = 0
+
+
+class TrackballMode:
+    '''Behaviour mode of trackball: mouse movement or vertical scroll'''
+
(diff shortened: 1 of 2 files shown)


xs5871:
couple of notes:
* The `Trackball`class should inherit from `kmk.modules.Module`
* There should be a polling rate limitation, i.e. read data every n milliseconds. The I2C polling is blocking and takes quite a long time. In addition the current implementation calculates acceleration purely based on distance, not velocity. The rate at which the controller polls isn't fixed and varies between controllers and configuration/load. A rate limit will give you more consistant acceleration.
* alternatively: implement read on interrupt. May be a future enhancement.
* The calculation of acceleration needs tuning, and personally I would use math that's easier on the MCU. Is the angle correction really necessary? If so, instead of throwing trigonometric functions at it, use a precalculated rotation matrix. Also not a big fan of rational exponentiation. This is of course debatable.
* set I2C frequency to a higher value. 400kHz is working fine, didn't try to go higher. Someone should read the datasheet of the processor on the trackball.

On your concerns:
1. No bus errors here.
2. That's the USB protocol, you have to basically "send" data more often. I got much better results with a rate limit of ~50ms or more (see last paragraph) for some reason. Not sure why, because that's a _very_ long time and it should have the opposite effect.
3. You'd have to look up on the respective USB HID Report. Looks like you could extend the mouse report to 5 bytes and use the last one for horizontal scrolling. No garantuee that this is correct.

kdb424 (approved):

kdb424:
I'll hold back from merging until @xs5871 and @Tonasz are happy with the implementation fully, but this looks great thus far. If you feel it's "good enough" for now, and a follow up PR should be done to push it further, that is also reasonable.

kmk/modules/pimoroni_trackball.py
```diff
@@ -0,0 +1,205 @@
...
+        )
+
+    def during_bootup(self, keyboard):
+        return
+
+    def before_matrix_scan(self, keyboard):
+        '''
+        Return value will be injected as an extra matrix update
+        '''
+        up, down, left, right, switch, state = self._read_raw_state()
```
xs5871, line 87:
```suggestion
        now = ticks_ms()
        if now - self.last_ticks < self.polling_interval:
            return
        self.last_ticks = now
        
        up, down, left, right, switch, state = self._read_raw_state()
```

xs5871 (commented):
suggestion for rate limiting

xs5871:
now that I'm looking at it, does the control flow ever reach the `finally` clause in '_i2c_rdwr'?
We can set the other optimisations I mentioned aside for an enhancement PR. Especially the acceleration would need some time to fine tune.

Tonasz:
I implemented your suggestions, thank you very much!

I had to retest them and noticed one major flaw which was highlighted by that polling - pointing_device report was not cleared after sending hid. That's maybe why you thought that bigger polling interval improved something - pointer started moving faster because it kept it's velocity some for 20ms, but it lost most of its accuracy.

"does the control flow ever reach the finally clause in _i2c_rdwr?" - finally is resolved just before return:
>If the try statement reaches a break, continue or return statement, the finally clause will execute just prior to the break, continue or return statement’s execution.
> -- https://docs.python.org/3/tutorial/errors.html

xs5871:
Very nice. From my perspective that's ready to merge.

PR #322, opened 2022-02-07
I2c encoder support

giovanniborella:
This PR adds support for the Adafruit I2C QT Rotary Encoder with NeoPixel break-out board over i2c. 

Only basic functionality like the stock encoder extension:
* invertable
* on_move, on_press keypress.

Still needs work for implementing NeoPixel-capabilities


kdb424:
Try running `make fix-formatting fix-isort test` and see if that returns a clean test. We should be able to review once that test passes.

kdb424:
isort is still not pleased, and looks like there are some other changes needed. We can get to those after you intend to come out of draft state, but the line I showed before should sort that, and has been tested on Mac and Linux. I'm not sure if anyone has tested Windows.

giovanniborella:
@kdb424 I'm trying to get make up and running on WSL, will pull out of draft when test is successfull

kdb424:
Looks generally great. Don't mind us nitpicking the small details here and there. Thanks so much for adding support for this, and we'll get it merged as soon as we can!

Tonasz:
A lot of code is duplicated with `encoder.py` file. Most of code could be reused. For example there could be `AbstractEncoder` and inheriting from him `GPIOEncoder` (I don't know what's the right name) and `I2CEncoder` and they would differ only where they need. What do you think about it?
Also, I see that doc example use `from kmk.modules.i2c_encoder import i2cEncoderHandler` but that's `EncoderHandler` in `i2c_encoder.py` file.

giovanniborella:
Good point @Tonasz! - ill get on merging the two

giovanniborella:
I dont have a GPIO encoder setup, so i'm afraid that part is untested

kmk/modules/encoder.py
```diff
@@ -76,26 +75,42 @@ def update_state(self):
...
-                if self.on_button_do is not None:
-                    self.on_button_do(self.get_state())
-
-    # returnd knob velocity as milliseconds between position changes (detents)
+    def button_event(self):
+        new_button_state = self.pin_button.get_value()
+        if new_button_state != self._button_state:
+            self._button_state = new_button_state
+            if self.on_button_do is not None:
+                self.on_button_do(self.get_state())
```
Tonasz, line 94:
```suggestion
        new_button_state = self.pin_button.get_value()
        if new_button_state != self._button_state:
            self._button_state = new_button_state
            if self.on_button_do is not None:
                self.on_button_do(self.get_state())
```

Tonasz (changes requested):
It looks a lot cleaner now :)
Despite reintroducing 2 small, but already fixed in the past issues, **it looks good to me**.
I don't have I2C encoders to test it. I got one classic encoder on my keyboard and it works fine. Somebody with multiple encoders could confirm it working, but if mentioned issue is fixed, it should be fine.

giovanniborella:
I merged @Tonasz' suggestions

kdb424:
I'll leave the final call to Tonasz before I check for any final merge blockers.

Tonasz (approved):
I'm not able right now to test how it will behave with multiple encoders (but it should be alright after latest changes) but this looks good to me :) I don't know other I2C encoders and how compatible they are with each other, so if there's gonna be a need, we will have to do some additional changes.

Tonasz (approved):
LGTM :) (see above comment)

kdb424 (approved):

PR #327, opened 2022-02-16
Add more detail for UART split docs

zyxwars:


xs5871 (commented):
More documentation is always appreciated. Two things:
- `keyboard.data_pinx` doesn't exist anymore, it's all in the split module.
- the heuristic to determine data pin order has been obsoleted with `uart_flip` by #301.

zyxwars:
I rewrote the example, removing keyboard.data_pin

kdb424:
I'll leave final approval up to @xs5871 when they are happy. I'll merge when and if they approve.

Tonasz:
I like this example a lot more after xs5871 suggestions. I'm wondering, maybe the topic of crossed cables (current example, with `uart_flip`) and not-crossed cables (default), should be more extended with example for each type of connection?

> keyboard.data_pinx doesn't exist anymore, it's all in the split module.

But split currently still do a fallback to `keyboard.data_pin` and EE hands below has example with `Split()` so using `keyboard.data_pin`. I also kinda like having `data_pin` included in `kb.py` files, so maybe it should not be hidden inside documentation?

PR #328, opened 2022-02-18
Split BLE fixes

Tonasz:
I couldn't make split BLE working so I started diving in it. Unfortunately I encountered many dead ends (like issues with PC Bluetooth adapter) and I couldn't recognize the simplest issue.
First commit is one line change which fixes a bug a make the split BLE useable for me.

Second commit are changes which were made around some other issues I encountered while testing:
- advertising was done till `self._connection_count < 2` condition is solved and it won't happen if you you are using Split BLE with USB HID (don't know if anybody ever tries it but...)
- while target (left) half was looking for the second (right) one, it was hanging for few seconds - theoretically you won't use only one half, but on the other - why not? I imagine left split gaming with right half saving the battery. I tried moving the connection check to the background
- the check of `self._connection_count == 1` doesn't tell if we're connected to PC and not other half or vice versa. I tried implementing a check, but the only thing I could find was `bleio_connection.discover_remote_services()` which I'm not sure if is reliable method
- I added `BLE_REFRESH` button code, because HID advertising was becoming sometimes stuck for me; I haven't noticed any BLE keybindings existing and they could be could idea (but for example, such refresh should probably not only restart advertising, but also connection)

Creating as draft, as we really need to talk about it design wise and check on multiple configuration.

### DIFF
--- a/kmk/handlers/stock.py
+++ b/kmk/handlers/stock.py
@@ -112,3 +112,14 @@ def hid_switch(key, keyboard, *args, **kwargs):
     )
     keyboard._init_hid()
     return keyboard
+
+
+def ble_refresh(key, keyboard, *args, **kwargs):
+    from kmk.hid import HIDModes
+
+    if keyboard.hid_type != HIDModes.BLE:
+        return keyboard
+
+    keyboard._hid_helper.stop_advertising()
+    keyboard._hid_helper.start_advertising()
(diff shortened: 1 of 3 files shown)


Tonasz:
I close this in favor of #358, which is based upon this one, but fix more stuff.

PR #330, opened 2022-02-20
Override LED creation in the RGB extension.

eltariel:
Adds an optional `pixels=` parameter to the RGB extension to allow an alternate LED driver to be used. Includes documentation with example code, but no implementation (yet).

This is split out from #262 for simplicity's sake. The Pimoroni boards will be in a future PR once these changes land, as they depend on both this PR and #329.

### DIFF
--- a/docs/rgb.md
+++ b/docs/rgb.md
@@ -147,3 +147,29 @@ Make sure that your board supports LED backlight by checking for a line with
 `PIXEL_PIN`. If it does not, you can add it to your keymap.  If you added the
 LED's yourself, you will also need to set `num_pixels` to the number of
 installed LED's in total.
+
+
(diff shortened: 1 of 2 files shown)


kdb424:
This has been tested on boards that already contain Adafruit LED's I assume?

eltariel:
> This has been tested on boards that already contain Adafruit LED's I assume?

Yep, the big blob PR included a board with the APA102 LEDs (Pimoroni Keybow). I literally lifted the example from the little board I have running right in front of me 😁 

The other board I have here that needs this is a Keybow 2040. It's got LEDs driven by an IS<something> chip which will need an adapter class to work like this.

kdb424 (approved):
Sounds great. Let's get this in!

Issue #336, opened 2022-02-22
Beginner-friendly example code for each type of feature supported


idesignstuff:
As one who is fairly new to custom/diy/mechanical keebs, and less than a beginner at python, some example code in the docs for the core features would be much appreciated.

Namely:

- [x] assigning simple and combo keycodes
- [x] layer do's and don'ts
- [ ] RGB control
- [x] macros

I'm excited & grateful to have found KMK, and hopeful that some maintainers will pick it back up.

kdb424:
It's still in active development, and by no means abandoned. We require additional help, especially as there is little time for myself or klardotsh to commit code directly, but many members are actively pushing code, and getting merged into mainline. RGB control should be handled completely in the RGB docs here https://github.com/KMKfw/kmk_firmware/blob/master/docs/rgb.md and what you call "macros" should be covered here. https://github.com/KMKfw/kmk_firmware/blob/master/docs/sequences.md

idesignstuff:
Thanks!
It's really encouraging that you are actively working on kmk. I'm grateful
for your effort!

What I meant with my issue was that, while RGB may be completely described,
how to call things isn't spelled out for beginners, even in the 'getting
started' section.

Thanks for the links!
I'll read them and see how far I get.


Christian

Tonasz:
What do you mean by  "completely described but things isn't spelled out for beginners"?
I'm genuinely curious, because RGB has one of the best doc among other undocumented stuff - it has:
- code snippet adding extension in `Enabling the extension` section
- keycodes and configuration variables listed
- public interface functions described for advanced modding and fiddling
- `Troubleshooting` section with most common issues

If you're looking for "complete examples" there is whole folder of configuration which can be copied to board and run right away - https://github.com/KMKfw/kmk_firmware/tree/master/boards - maybe it's not featured enough?

idesignstuff:
Tonasz, wow!

Thanks for the tip about the running configs in the boards folder!

I'm a beginner at writing code & custom keebs. Since I started with a
handwired custom board, it never occurred to me to look there. I assumed
the boards folder would be only matrix and pin configs.

And, yes the RGB section of the docs really spells out the features and
functions available, but NOT how to roll those into other commands (for
instance: press a layer switch key = switch layers AND change color of the
RGB).


Christian

idesignstuff:
So, I've had a few sessions to try some of this stuff out.
[...]

PR #338, opened 2022-02-24
Add doc index

Tonasz:
I don't know if this is right approach, but an index was something I missed during early KMK adoption. I know that list of files is an index on itself, but it's not categorized.
Inspired by #336 issue, which was looking for already existing documentation.


kdb424:
I'll have to actually take a look at this again when I'm awake, but generally speaking, I like this. Just want to make sure things aren't missing if we are going to push to get them out there front page!

kdb424 (approved):

PR #340, opened 2022-02-25
animation speed based on system timer

xs5871:
Ties RGB animation speed to supervisor ticks, instead of main loop execution frequency.


kdb424 (approved):
That's likely what I should have done. Thanks!

PR #345, opened 2022-02-28
Update keys.md

idesignstuff:
Added example for use of .after_keys_handler() to change RGB settings and add serial print.


kdb424:
Appreciate!

kdb424 (approved):

PR #346, opened 2022-03-01
Add kmk.scanners.bidirectional

s-ol:
This is for matrices that use diodes in both directions to double the number of keys without using extra pins.

>     A Scanner for Keyboard Matrices with Diodes in both directions.
>     In a bidirectional matrix, each (col, row) crossing can be used twice -
>     once with a ROW2COL diode ("A"), and once with a COL2ROW diode ("B").
>     The raw key numbers returned by this scanner are based on this layout ("up_down"):
>         C1  C2  C3
>       +-----------
>     R1| A0  A1  A2
>     R2| A3  A4  A5
>       +-----------
>     R1| B6  B7  B8
>     R1| B9 B10 B11

Here's an example schematic that would use the `mapping_left_right` layout:

image

I'll keep this PR a draft until my board prototype arrives and I can verify everything works properly. It would also be cool to cover the scanner with tests, but I don't have much experience with python testing or circuitpython so if someone wants to help out with that that would be great :smiley:

### DIFF
--- a/kmk/scanners/bidirectional.py
+++ b/kmk/scanners/bidirectional.py
@@ -1,0 +1,220 @@
+import digitalio
+
+from kmk.matrix import KeyEvent
+from kmk.scanners import Scanner
+
+
+def range_rev(start, stop):
+    return range(stop - 1, start - 1, -1)
+
+
+def mapping_up_down(cols, rows):
+    '''
+    Maps keys according to the following diagram:
+
+       1 2 3
+     +------
+    1| A A A
+    2| A A A
+     |
+    1| B B B
+    2| B B B
+    '''
+
+    coord_mapping = list(range(2 * cols * rows))
+    return coord_mapping
+
+
+def mapping_up_down_mirrored(cols, rows):
+    '''
+    Maps keys according to the following diagram:
+
+       1 2 3
+     +------
+    1| A A A
+    2| A A A
+     |
+    2| B B B
+    1| B B B
+    '''
+
+    coord_mapping = []
+    coord_mapping.extend(range(cols * rows))
+    coord_mapping.extend(range_rev(cols * rows, 2 * cols * rows))
+    return coord_mapping
+
+
+def mapping_left_right(cols, rows):
+    '''
+    Maps keys according to the following diagram:
+
+       1 2 3  1 2 3
+     +-------------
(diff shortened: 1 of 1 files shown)


kmk/scanners/bidirectional.py
```diff
@@ -0,0 +1,229 @@
...
+    1| A A A
+    2| A A A
+     |
+    1| B B B
+    2| B B B
+    '''
+
+    coord_mapping = []
+    for y in range(2 * rows):
+        coord_mapping.extend(ic(y, x, cols) for x in range(cols))
```
xs5871, line 22: A heads-up:
`intify_coordinate` should be regarded as legacy for backwards compatibility . It's easier to understand what's going on if you do:
```python
coord_mapping = list(range(2*cols*rows))
```

xs5871 (commented):
Looks fine so far, but I'll wait until this is out of draft for a proper review.

kdb424:
@s-ol Curious to if you are continuing to work on this. I'm closing stale issues and PR's, but don't want to close this by mistake if you are intending to continue work on this.

s-ol:
@kdb424 no worries, I will work on this but I'm still waiting to receive my MCUs unfortunately

kdb424:
Fantastic! Glad to hear it's still being worked on. No rush.

kmk/scanners/bidirectional.py
```diff
@@ -0,0 +1,220 @@
+import digitalio
+
+from kmk.matrix import KeyEvent
```
xs5871, line 3:
```suggestion
from keypad import Event as KeyEvent
```

xs5871 (changes requested):
Looks sane to me. I won't build custom hardware specifically to test it, but I trust that it's in the authors own interest that it works.
Use the CP implementation of the KeyEvent and we're good.

s-ol:
Sooo playing around with this a bit I figured out *why* this isn't done more commonly:  
It has ghosting problems. The version of the code here has them already with certain combinations of two keys, but it can be improved by tri-stating the unused output lines while scanning. At that point you have ghosting only with four keys pressed, but it's still unsuitable for my application.

Since my board is probably the only one in existence right now made that way, and soon it won't be anymore, I'll close this PR :wink:

PR #350, opened 2022-03-03
Layer example code

HBBisenieks:
Inspired by [this post](https://www.reddit.com/r/circuitpython/comments/szjcl2/kmk_firmware_advice_for_handwired_macropad/) in the r/CircuitPython subreddit to add example code for all layer keycodes complete with explanation of their function and any caveats similar to the layer documentation in QMK.


kdb424:
Seems a little verbose, but we also do attract a different crowd than the traditional firmwares. Seems reasonable if this is helpful. Would love for some input from other reviewers though.

HBBisenieks:
Honestly, I think it is a little verbose, and I'm not opposed to changing it now or in the future to cut down on that. I do think that KMK is seeing interest from folks who maybe haven't used something like QMK before, so the example code stuff might be overboard. I'd love to see the KMK community put together a style-guide for docs (and also adopt a code of conduct), but that's neither here nor there for this PR.

idesignstuff:
Hey, verbose is great for the newbs (like me) to really have some concepts spelled out. I'd say maybe the docs should be concise, but the verbose examples need a home: maybe their own folder, or as an addendum to the 'getting started' info.

xs5871:
I don't mind the verbosity as much, it's the redundancy with the keycode table above. There's value in showcasing a couple of the layer keys, especially with argumented keys, but we really don't need an `EXAMPLE = KC.FOO` for each and every one of them.
As for the layout: I have to scroll to match the example code with the example description, and if we're having that discussion anyway, that's not the most helpful design.

If you'd press me for constructive criticism, I'd suggest:
* merging all the code parts (module init, layer definition, layer keys) into one Example section, 
* reduce the number of layers to two and 
* pick like three layer keys to showcase, 
* work in an example for custom modtap behavior,
* then have one somewhat verbose text block below giving context or insights to the code above

HBBisenieks:
word, I'll see about reworking it more along those lines, @xs5871

xs5871:
I'm happy with that.

kdb424:
Gave it a read, and that's much cleaner, but seems to make most of the points that you had tried to make originally. Thanks for the contribution!

PR #356, opened 2022-03-09
implement combo/chord/sequence module

xs5871:
I'd like to submit this for discussion.
It's not quite as broad as the swallow API mentioned in #222, but I think it's generalized enough to be able to handle most multi-key functionality. Commit includes documentation, so I won't go into details about features.

I don't mind keeping this PR open for a while. I stumbled upon some scaling and consistency issues with our current timeout API that I started working on that on the side.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,6 +22,7 @@
 
 ## [Modules](modules.md)
 
+- [Combos](combos.md): Adds chords and sequences
 - [Layers](layers.md): Adds layer support (Fn key) to allow many more keys to be put on your keyboard
 - [ModTap](modtap.md): Adds support for augmented modifier keys to act as one key when tapped, and modifier when held.
 - [Mouse keys](mouse_keys.md): Adds mouse keycodes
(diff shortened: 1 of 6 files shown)


kdb424:
I love what I see. I'd love to see if this API can be used for layers, mod tap, tap dance, ect, and if that can all be ported, I'd love for this to go in, and unify all of the time based keys as the standardized API, and get an example of how to make use of this in some of the most basic ways possible for each major feature, so it's easy to adapt any ideas to.

xs5871:
It'll be interesting to see how much of the functionality of both systems could be combined without getting too unwieldy. Of course `HoldTap` is currently much more efficient at dealing with single key repeats. One of my next objectives is to de-jank the `tapdance` code and my brain needs a break from combos.

kdb424:
I don't need that merged in this commit to get it in, but I do want to have that at least in mind before merging. Let me know if that's reasonable, and I'll do a final review.

xs5871:
You can go ahead with the review. I'll work on further convergence piece by piece.

kdb424 (approved):

Issue #80, opened 2018-10-22
Support "standalone keys"


klardotsh:
This could be the push button on a rotary encoder (such as Adafruit's p/n 377), a kick pedal, whatever (though kick pedal support is a totally different type of beast than the button on the Adafruit 377). Support both of these cases as "secondary" matrices, probably scanned before (but in addition to, rather than instead of, as @kdb424 does for split keyboards) the primary matrices.

This came up in the CircuitPython Weekly for 22 Oct 2018 as something someone is interested in (though the comment was not in a KMK context, it's still a good idea)

klardotsh:
I have some sketches locally of what a full-on multi-matrix KMK might look like, which is really the core of what this idea was getting at. Going to leave the issue open, but it'll probably evolve a bit to encompass a refactor of split keyboard handling as well (in preparation for some mythical v1.5/v2).

pullenrc:
I implemented a rotary encoder class(pull request just sent today), how the rotation is being handled could easily be applied to stand alone keys not in a matrix, as it's checking regular I/O lines for changes.

klardotsh:
Mostly superceded by #237 - `keypad` will provide us easy mechanisms to handle this. Not sure whether I want to leave this open or not, since I guess it's not a strictly related to moving to `keypad`, just something we'll get cheaper/free with that migration.

xs5871:
resolved by `keypad` and support for multiple scanners.

PR #357, opened 2022-03-10
Support for rotary encoders using `rotaryio`

mbellitti:
I have created a module `RotaryIOEncoder` that wraps the functionality of [rotaryio](https://docs.circuitpython.org/en/latest/shared-bindings/rotaryio/index.html#module-rotaryio). 

The current KMK module does not use `rotaryio`, so it does not have support for `divisor` (essential when the number of retents is not the same as the number of cycles per revolution). It also has some other minor issues, like dropping some keypresses if they happen too fast, which is solved by using `rotaryio`.

The module in this pull request allows the user to add multiple rotary encoders to a project, with layers support. It does not support press buttons on the encoder because `rotaryio` does not, they should be added to the matrix or included using `keypad` in a future version of KMK.

I have tested this module with [this](https://www.adafruit.com/product/5001) scroll wheel.

### DIFF
--- a/docs/rotaryio_encoders.md
+++ b/docs/rotaryio_encoders.md
@@ -1,0 +1,50 @@
+The module `RotaryIOEncoder` is a wrapper around `rotaryio`  (see [here](https://docs.circuitpython.org/en/latest/shared-bindings/rotaryio/index.html#module-rotaryio)). `rotaryio` has a quirk: encoders must use pins that are consecutive (for example GP16 and GP17 on a Raspberry Pi pico). 
+
+`RotaryIOEncoder` supports using multiple encoders at the same time, each with its own keymap (multiple layers are allowed).
+
+Press buttons are not supported, you should include them in your matrix or wait for the `keypad` module (available in CircuitPython 7.0.0) to be integrated in KMK.
+
+The following is a complete example (you can use it as your `code.py`) of how to use `RotaryIOEncoder`. The example uses `"divisor" = 2` because it was tested using [this](https://www.adafruit.com/product/5001) scrollwheel, but for most cases `"divisor": 4` should work.
+
+```
+import board
+
+from kmk.kmk_keyboard import KMKKeyboard
+from kmk.matrix import DiodeOrientation
+from kmk.keys import KC
+from kmk.modules.rotaryio_encoder import RotaryIOEncoder
+from kmk.modules.mouse_keys import MouseKeys
+
+keyboard = KMKKeyboard()
+
+# We want to use the encoder to scroll so we need mouse keys
+keyboard.modules.append(MouseKeys())
+
+# In this example we use only one encoder, connected to the pin 16 and 16
+# of a raspberry pi pico.
+enc1 = {"pin_a": board.GP17,
+            "pin_b": board.GP16,
+            "divisor": 2,
+            "keymap": [[KC.MW_UP, KC.MW_DN],]
+            }
+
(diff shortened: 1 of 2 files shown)


Tonasz:
I haven't got time yet to test it, but I'm curious what do you (submitter and others) think about merging that with existing Encoder module. It already got the old implementation and I2C variant, so maybe adding the third, rotario would be more tidy?
Afaik, no one remember why current implementation has comment that it couldn't use rotaryio, so maybe this could be even the default implementation, and the previous one would the additional one for a while?

xs5871:
I agree with Tonasz, I'd much prefer rotaryio as an additional option in the already existing encoder module.
Regarding the current implementation: Rotaryio wasn't used explicitly because it required consecutive pins. Might be another case of allowing compatibility to existing hardware.

btw: KMK already supports the `keypad` API.

mbellitti:
> I haven't got time yet to test it, but I'm curious what do you (submitter and others) think about merging that with existing Encoder module. 

That could work. We could include three classes in the rotary encoder module:
1. using `rotaryio`
2. I2C
3. "by hand" implementation

It sounds like the best option is to combine the two modules and give the user the option of which class to use. I would be happy to start working on it but I don't have the hardware to run tests right now.

> Might be another case of allowing compatibility to existing hardware.

Makes sense. 

> btw: KMK already supports the keypad API.

I see. I mentioned that in the docs for this module because [this issue](https://github.com/KMKfw/kmk_firmware/issues/80) is still open, and having extra keys on a rotary encoder is exactly that kind of problem.

johnskiller:
> I agree with Tonasz, I'd much prefer rotaryio as an additional option in the already existing encoder module. Regarding the current implementation: Rotaryio wasn't used explicitly because it required consecutive pins. Might be another case of allowing compatibility to existing hardware.
> 
> btw: KMK already supports the `keypad` API.

I'm not sure about what "consecutive pins" means, is it must be like GP0,GP1, so GP0, GP8 not work?

but from https://docs.circuitpython.org/en/latest/shared-bindings/rotaryio/index.html the rotaryio can use any two pins:
class rotaryio.IncrementalEncoder(pin_a: [microcontroller.Pin](https://docs.circuitpython.org/en/latest/shared-bindings/microcontroller/index.html#microcontroller.Pin), pin_b: [microcontroller.Pin](https://docs.circuitpython.org/en/latest/shared-bindings/microcontroller/index.html#microcontroller.Pin), divisor: [int](https://docs.circuitpython.org/en/latest/docs/library/builtins.html#int) = 4)

I have't test it yet since my keyboard using GP13 and GP14, maybe use another test bed.

xs5871:
> but from https://docs.circuitpython.org/en/latest/shared-bindings/rotaryio/index.html the rotaryio can use any two pins: class rotaryio.IncrementalEncoder(pin_a: [microcontroller.Pin](https://docs.circuitpython.org/en/latest/shared-bindings/microcontroller/index.html#microcontroller.Pin), pin_b: [microcontroller.Pin](https://docs.circuitpython.org/en/latest/shared-bindings/microcontroller/index.html#microcontroller.Pin), divisor: [int](https://docs.circuitpython.org/en/latest/docs/library/builtins.html#int) = 4)

It's not mentioned in the CP docs, but the RP2040 implementation is (for some reason) done with PIO and because of that _requires_ the pins to be sequential.

kdb424:
Closing this for now as it seems to have been resolved in a different PR.

PR #359, opened 2022-03-11
Add boot.py docs

zyxwars:


docs/boot.md
```diff
@@ -0,0 +1,48 @@
...
+
+```python
+import supervisor
+import board
+import digitalio
+import storage
+import usb_cdc
+import usb_hid
+
+supervisor.set_next_stack_limit(4096 + 4096)
```
xs5871, line 32: A comment to this line would be nice, like "this is the circuit python default" or "refer to your board specific boot.py". I don't know if this is board specific.

docs/boot.md
```diff
@@ -0,0 +1,48 @@
...
+
+# If this key is held during boot, don't run the code which hides the storage and disables serial
```
xs5871, line 34: Add something along the lines: The exact pins are dependend on your hardware wiring.

xs5871 (commented):
I have a suggestion mainly targeted at newcomers.

docs/boot.md
jpconstantineau: @tannewt provided an example of his boot.py in his [repo](https://github.com/tannewt/keebloop/blob/main/boot.py).  He doesn't use KMK but he loads his board configuration in boot.py and uses it to detect if there are any keys pressed while booting.  If so,  then the different USB devices are not disabled.

I suspect a similar approach could be used if kb.py is loaded.

For the PyKey series of boards, they all have the same row/column pin names as part of the boards and it's easy to hardcode the pin names to first row/first column.  See [here](https://github.com/jpconstantineau/pykey/blob/main/pykey_boards_production/pykey60/boot.py) for what I use.  I suspect it's almost identical.

However, for any other boards, we would need to be able to pick up dynamically what pins are used.

As a first pass, this example is good but comments on what needs to be changed would be useful. Especially why the output/input config and why it's in a specific way. (diodes are pointing which way...)

jpconstantineau: I wonder if you do have compile issues in kb.py if you really get stuck or if it simply bails and goes into safe mode with the USB drive enabled with the error messages in boot_out.txt.  Even if it doesn't go in safe mode, wouldn't it not run the USB disabling code that prevents you to go in programming mode?

Give me some code I can try and I can see what really happens.  If my boards do get locked out, I do have a jlink to erase and reflash everything from the bootloader.  

The only way I could potentially see why you would get locked out is in cases where your GPIOs are valid (not a compile error) but not connected to switches or set with the wrong diode direction.  It would never be able to detect when you press the switch and thus always automatically disable USB.  Perhaps, the documentation should refer to the method of how getting in safe mode for those trying to configure their boot.py but incorrectly do so and lock themselves out.

klardotsh:
Hey look I don't have a PR review here, but that's actually why I'm commenting: I wanted to thank you all.

- @zyxwars for showing up and contributing your first pul request - welcome, and thank you!
- the rest of you, for handling this review with thoughtful and respectful, productive feedback.

I originally built this firmware almost as a joke, and actually quite literally as a mistake (I bought the wrong microcontrollers from a local shop in Vancouver BC for the keyboard I was trying to build, and QMK didn't support them, so I had to DIY something that would work... and so it began). It warms my heart to see people taking care of this thing so many years later. Thank you.

We're working on promoting some folks to being proper bona-fide maintainers of KMK now that I've mostly stepped down (I just pay for the domain name and host the matrix room, really) and @kdb424 is busy doing life, but in the meantime: ping me when you all have a resolution on this with your group approval, and I'll merge whatever you all agree on. Cheers, and good night!

docs/boot.md
xs5871: That is probably right. Let's take the discussion from this PR into an issue.

docs/boot.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+### Disable serial comms
+
+```python
+# Equivalent to usb_cdc.enable(console=False, data=False)
+usb_cdc.disable()
+```
+
+### A fully working example, with a
```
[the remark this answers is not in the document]
zyxwars: Oops, fixed it.

docs/boot.md
zyxwars: Comment added.

zyxwars:
Okay so I added some more comments explaining the code.

xs5871 (approved):
Awesome, thanks.

kdb424 (approved):

PR #366, opened 2022-03-24
fix: after_matrix_scan when board is the target

Boot-Error:


kdb424:
Thanks for the PR. Clean and simple fix! Really appreciate the follow up on that.

kdb424 (approved):

Issue #224, opened 2021-08-13
Convert RGB to use adafruit_fancyled


kdb424:
https://learn.adafruit.com/fancyled-library-for-circuitpython/install-software
This would offload a lot of HSV calculations into C, which would be much less intensive on the controller, as well as making new effects easier.

xs5871:
I conducted a scientifically questionable speed comparison: The current KMK implementation for hsv/rgb conversion is about 5.5 times faster than adafruit_fancyled, and 3 times faster than adafruit_fancyled converted to integer arithmetic.
example:
```python
for i in range(1000):
    CRGB(CHSV(i, 255, 255))
```

Some of that slowness is due to the fact that adafruit_fancyled creates new objects for every conversion. If the CHSV object is not re-initialized, and only converted to CRGB, KMK is only 2 times faster than adafruit_fancyled.
example:
```python
c = CHSV(0, 255, 255)
for i in range(1000):
    c.hue = i
    CRGB(c)
```

however... if i take my dodgy fancyled-but-integer-only conversion, and use it in a conversion function instead of instantiating a class, it turns out to be twice as fast the current KMK one. It still has the occasional of-by-one rounding error, but that should be fixable.

PR #367, opened 2022-03-24
Enhancement rgb

xs5871:
relates to #224. Implements:
- use common clamping function in incrementing/decrementing methods
- faster HSV to RGB conversion
- make neopixel lib only necessery for default pixelbuffer

HSV domain has been changed from [0-360, 0-100, 0-100] to [0-255, 0-255, 0-255]. Makes the math easier and faster for 16 bit fixed point calculations.

### DIFF
--- a/kmk/extensions/rgb.py
+++ b/kmk/extensions/rgb.py
@@ -6,6 +6,8 @@ from kmk.extensions import Extension
 from kmk.handlers.stock import passthrough as handler_passthrough
 from kmk.keys import make_key
 from kmk.kmktime import PeriodicTimer
+from kmk.utils import clamp
+
 
 rgb_config = {}
 
@@ -321,10 +323,7 @@ class RGB(Extension):
         if step is None:
             step = self.sat_step
 
-        if self.sat + step >= 100:
-            self.sat = 100
-        else:
-            self.sat += step
+        self.sat = clamp(self.sat + step, 0, 100)
 
         if self._check_update():
             self._do_update()
@@ -337,10 +336,7 @@ class RGB(Extension):
         if step is None:
             step = self.sat_step
 
-        if (self.sat - step) <= 0:
-            self.sat = 0
-        else:
-            self.sat -= step
+        self.sat = clamp(self.sat - step, 0, 100)
 
         if self._check_update():
             self._do_update()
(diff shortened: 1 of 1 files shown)


kdb424:
While I love most of this, I do wonder if we are going to lose any actually noticeable range of RGB by clamping to 255 instead of 360 that gets lost. I like the perf that we gain, but would like to see if anyone else notices, or misses that extra colour range (mostly in hue, not sat/val)

xs5871:
That is a fair point and something I considered.
The missing color resolution isn't really humanly noticeable for static colors, especially since common RGB LEDs aren't that color accurate anyway and we're not dealing with color accurate print media, for example. Jumps in brightness are much more noticable (from a perception point of view), and we'll have some rounding errors some way or another, if we don't want to use floats. Unless we use at least 24 bits for each axis, but I think that's overkill for our use. Value and Saturation have a higher impact than Hue on smooth color transitions as soon as they aren't at 100%.
The [255, 255, 255] range is standard/common for embedded systems, so that's what I went with.

kdb424:
Purely anecdotal, but without the PR, I noticed a wide range of blue, but the red range felt insanely shallow if that makes sense. I haven't tested this on hardware yet, but on the swirl animation, this was very noticeable. I'll be sure to test on real hardware and give in a look. I agree that resolution doesn't have to be hyper accurate, but seeing that some things already felt compressed, I had the jerk reaction. I'm sure it's something you tested though already, and you have proven to know what you are doing. Once I get a chance to test on hardware, code looks clean. I'll try to have it merged by end of weekend at the latest.

xs5871:
The "original" KMK implementation didn't use the full range of 255 values for each RGB channel, but capped out at 100. I only noticed it because of the refactor and assumed that was an oversight. The non-linearity of neopixels and absence of any kind of gamma correction may very well result in the effect you describe. The apparent brightness of the red and green channel of WS2818 is much, much less than the blue one; even more so if you turn down the brightness.

kdb424:
Well that explains a lot. I won't lie, I wrote that HSV to RGB after learning the math, noticed it mostly worked, and shipped it. Sounds like you didn't remove any functionality then over my flawed implementation. Code looks clean, and that checks out. Merging.

kdb424 (approved):

PR #374, opened 2022-03-31
Allow passing user-defined MPY_SOURCES and MPY_CROSS to make

xs5871:
It's not unlikely to have multiple mpy-cross binaries on a system, so it makes sense to let user choose one.
Customizable MPY_SOURCES let me compile board specific files or external libraries as well.


kdb424:
Good catch

PR #375, opened 2022-03-31
implement stricter and more consistent timeouts.

xs5871:
- stricter timeouts: Instead of adding an entire millisecond, use a list
  of timeouts that are supposed to be handled on the same tick. This
  reduces the delay between target tick and actual tick, especially for
  many timeouts issued to the same tick, i.e. combos.
- consistent timeouts: Timeouts are now guaranteed to be handled in the
  order they were issued. Timer rollover is handled properly.

caveat / change to the interface: the returned `timeout_key` is a tuple
of the target timeout tick and the index of the timeout at that tick.

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -3,6 +3,7 @@ from supervisor import ticks_ms
 from kmk.consts import KMK_RELEASE, UnicodeMode
 from kmk.hid import BLEHID, USBHID, AbstractHID, HIDModes
 from kmk.keys import KC
+from kmk.kmktime import ticks_add, ticks_diff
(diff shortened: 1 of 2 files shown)


kdb424 (approved):
Looks clean, and I like the functionality. Solid.

kdb424:
Looks like this broke the tests for combos in `tests/test_combos.py", line 34, in test_basic_kmk_keyboard`. I didn't catch that before merge, but running `make test` locally on master shows what's broken. Can we resolve that in a followup?

xs5871:
Yes, I already pin-pointed the issue. Will provide a fix ASAP.

PR #376, opened 2022-04-01
fix failing unit test.

xs5871:
During timeout processing, cancelations have to take immediate effect,
but new 0-delay timeouts have to be moved to the next cycle.


kdb424 (approved):
Thanks for the really swift fix!

PR #380, opened 2022-04-05
add `None` to account for no encoder button in encoder docs

funkfinger:
I had to dig through the code to find out what to do... adding it to the docs...


xs5871:
That checks out. Thanks.

xs5871 (approved):

kdb424 (approved):

PR #382, opened 2022-04-05
Fix WinCompose usage

sparksThefire:
WinCompose requires the `enter` key to be typed at the end of a special unicode character. In order to help print emojis and emoticons in a windows environment, I have added the enter key to the Windows compose unicode sequence method. I have tested this with my local keyboard.


klardotsh (approved):
awesome, thank you, glad this worked out!

PR #383, opened 2022-04-08
caps word implementation

masiullah:
This is a similar feature available in some of the forks of qmk allowing to break out of caps lock if you type keys that break the word
Please review and provide feedback if this is in the correct direction
This is my first contribution as I am learning Python and experimenting with the kb2040 I got recently

### DIFF
--- a/kmk/modules/capsword.py
+++ b/kmk/modules/capsword.py
@@ -1,0 +1,98 @@
+from kmk.keys import KC, ModifierKey, make_key
+from kmk.modules import Module
+
+
+class CapsWord(Module):
+    # default timeout is 8000
+    # alphabets, numbers and few more keys will not disable capsword
+    def __init__(self, timeout=8000):
+        self.alphabets = range(KC.A.code, KC.Z.code)
+        self.numbers = range(KC.N1.code, KC.N0.code)
+        self.keys_ignored = (
+            KC.MINS,
+            KC.BSPC,
+            KC.UNDS,
+        )
+        self._timeout_key = False
+        self._cw_active = False
+        self.timeout = timeout
+        make_key(
+            names=(
+                'CAPSWORD',
+                'CW',
+            ),
+        )
+
+    def during_bootup(self, keyboard):
+        return
+
+    def matrix_detected_press(self, keyboard):
+        return keyboard.matrix_update is None
+
+    def before_matrix_scan(self, keyboard):
+        return
+
+    def process_key(self, keyboard, key, is_pressed, int_coord):
+        if self._cw_active and key != KC.CW:
+            continue_cw = False
+            # capitalize alphabets
+            if key.code in self.alphabets:
+                continue_cw = True
+                keyboard.process_key(KC.LSFT, is_pressed)
+            elif (
+                key.code in self.numbers
+                or isinstance(key, ModifierKey)
+                or key in self.keys_ignored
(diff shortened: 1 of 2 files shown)


kdb424:
I like this functionality, and want to hear what one of our reviewers says about the code, but from a glance, I think this looks fairly good!

kmk/modules/capsword.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # enables/disables capsword
+        if key == KC.CW and is_pressed:
+            if not self._cw_active:
+                self._cw_active = True
+                self.discard_timeout(keyboard)
+                self.request_timeout(keyboard)
+            else:
+                self.discard_timeout(keyboard)
+                self.process_timeout()
```
xs5871, line 65: This should go in the `KC.CW.on_press` handler.

kmk/modules/capsword.py
```diff
@@ -0,0 +1,98 @@
...
+            continue_cw = False
+            # capitalize alphabets
+            if key.code in self.alphabets:
+                continue_cw = True
+                keyboard.process_key(KC.LSFT, is_pressed)
+            elif (
+                key.code in self.numbers
+                or isinstance(key, ModifierKey)
+                or key in self.keys_ignored
+                or key.code >= 1000  # user defined keys are also ignored
```
xs5871, line 46:
```suggestion
                or key.code >= FIRST_KMK_INTERNAL_KEY  # user defined keys are also ignored
```
for consistency: let's reference magic numbers by name.

kmk/modules/capsword.py
```diff
@@ -0,0 +1,98 @@
...
+                'CAPSWORD',
+                'CW',
+            ),
+        )
+
+    def during_bootup(self, keyboard):
+        return
+
+    def matrix_detected_press(self, keyboard):
+        return keyboard.matrix_update is None
```
xs5871, line 30: This method is never used.

xs5871 (commented):
I've got some comments, nothing major.
Docs would be nice at some point, of course.
I'm almost certain I could break it with the right unit-test, but that's a bridge we can cross when there are complaints, to be honest.

kmk/modules/capsword.py
masiullah: Sure, will definitely

kmk/modules/capsword.py
masiullah: I will remove this method

kmk/modules/capsword.py
masiullah: I will move this to the handler

masiullah:
> I've got some comments, nothing major. Docs would be nice at some point, of course. I'm almost certain I could break it with the right unit-test, but that's a bridge we can cross when there are complaints, to be honest.

I would also want to do unit test cases from the computer. Is there any doc or guidance available that can help me perform those UT cases

xs5871:
There's pretty much no documentation, and be warned, the unit test code is very bare bones.
Tests are under `tests/`. You can take a look at the `test_*.py` files for reference and run them with `make unit-tests`.
I wouldn't mind approving the PR without a unit test for now. You can always submit it later.

masiullah:
Please review and let me know your thoughts

xs5871 (approved):

kdb424 (approved):

kdb424:
Merging based on a passed review.

PR #384, opened 2022-04-08
swapping control with cmd

masiullah:
Another qmk inspired feature which allows swapping control with cmd and vice versa
Helpful when using the keyboard on mac and windows

### DIFF
--- a/docs/cg_swap.md
+++ b/docs/cg_swap.md
@@ -7,7 +7,7 @@ from kmk.module.cg_swap import CgSwap
 # cg_swap disabled on startup
 cg_swap = CgSwap()
 # cg_swap enabled on startup
-# cg_swap = CgSwap(true)
+# cg_swap = CgSwap(cg_swap_enable=True)
 keyboard.modules.append(cg_swap)
 
 ```
--- a/docs/cg_swap.md
+++ b/docs/cg_swap.md
@@ -1,0 +1,13 @@
+# Ctrl Gui Swap
+This module allows to swap Ctrl with Gui and vice versa. This will reset on restart to the default implementation
+
+## Enabling the module
+```python
+from kmk.module.cg_swap import CgSwap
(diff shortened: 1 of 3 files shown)


kdb424 (changes requested):
I like the functionality, but I'd like to see a change in the name of the module as it sounds too broad, and may collide with others thoughts on what it does. Functionality wise, I think it looks great.

masiullah:
> I like the functionality, but I'd like to see a change in the name of the module as it sounds too broad, and may collide with others thoughts on what it does. Functionality wise, I think it looks great.

Thanks. I agree the module name is too broad. I will change the name of the module to reflect exactly the feature it is performing at the moment
My thought was to add more keys in future to the same module which can swap other modifier keys and may be others could also add on to the same module if additional swap functionality are required

kdb424:
If you can come up with a valid use case other than Mac/Other swap keys, I'm open to ideas. Otherwise, I may suggest something like mac_swap or something similar. I also use my keyboard on Mac/Linux, but I do all the hacking OS level for reasons, so I didn't bother with this sort of keyboard hacks, but I understand not everyone wants to, or can, do that.

masiullah:
Honestly, I don't have any other valid use cases for swapping the modifiers or any other keys. Unless some one else can provide other valid cases modifying the module name will be the best option
I will rename the module to either mac_swap or cg_swap (control gui swap)

kdb424:
Sounds great to me! Really appreciate the contribution, and that's a great feature. I did almost forget. Needs docs to go with it. Don't know how that slipped my mind.

masiullah:
Sure, will also add docs for both the pull requests with at least basic information on usage

kdb424:
Perfect. I'll do what I can do get them reviewed and merged this weekend if mergable, or at least give reasonable feedback.

kdb424 (approved):

PR #373, opened 2022-03-31
Fix direct GPIO scanner for split keyboards

levpopov:
https://github.com/KMKfw/kmk_firmware/issues/372

PR #388, opened 2022-04-10
Enhancement default scanner

xs5871:
Makes `keypad.KeyMatrix` the default scanner; depends on #373 to work with splits.
The PR has to be rebased, once #373 is merged. I'll mark it ready for review once that's done.


xs5871:
We can leave this open for a bit, it's *technically* a breaking change.
I'd also appreciate confirmation from a split user that everything is in order.

kdb424 (approved):

PR #389, opened 2022-04-10
Levpopov direct gpio split fix

kdb424:
Rebased #373 and rebased to pass test.

### DIFF
--- a/kmk/scanners/native_keypad_scanner.py
+++ b/kmk/scanners/native_keypad_scanner.py
@@ -1,7 +1,6 @@
 import keypad
 
-from kmk.matrix import DiodeOrientation, KeyEvent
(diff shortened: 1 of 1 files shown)


xs5871 (approved):
Thank you very much, that was quick.

kmk/scanners/native_keypad_scanner.py
```diff
@@ -1,6 +1,7 @@
 import keypad
 
-from kmk.scanners import DiodeOrientation, Scanner
```
kdb424, line 3: Just wanted to make sure this is correct @xs5871

kdb424:
@xs5871 I don't think the ping worked in the line comment. Wanted to make sure you were notified of that concern before I merge.

PR #390, opened 2022-04-10
Split tap key and encoders working on splits

daysgobye:
I started working on this because I wanted encoders to work on splits. 

So after looking at how encoders worked I saw that they are not part of the matrix and just called "tap_key". I thought the most pragmatic way of doing this was be able to call tap_key from the offside. I think this not only fixes encoders but is a good thing to have in the code base in general. If we have a function to just send a key press we should also be able to use that on splits. 

Let's get into how I did this and what changes I made.

In the split module, I added a new uart header at the top of the file. This will be added to the messages sent to see if it should be handled like normal or as a " tap_key ". Next, I had to add the send_key_press function. This is the function you want to use outside of the split code. Right now it just calls the _send_uart_press function if uart is being used. But If this gets merged I will work on the BLE side.
Now we have a function we can call we need to send it to the other side. So I made "_serialize_keyCode" This takes in our keycode and gives us back bytes we can send over uart.

Ok offside done we need to handle it correctly.

Incomes "_deserialize_keyCode" takes those bytes and gives back a kmk keycode same as KC.ESC. Then in "_receive_uart" We had an if block that was checking the old uart header so already nothing was broken it was just not dealing with the new messages. So we simply add a new if block that checks to see if it is the "key press header" and handle it thereby calling"_deserialize_keyCode" to get our key and keyboard.tap_key(deserialize_keyCode)

### DIFF
--- a/docs/encoder.md
+++ b/docs/encoder.md
@@ -24,7 +24,7 @@ encoder_handler = EncoderHandler()
 keyboard.modules = [layers, modtap, encoder_handler]
 ```
 
-2. Define the pins for each encoder (pin_a, pin_b, pin_button, True for an inversed encoder)
+2. Define the pins for each encoder (pin_a, pin_b, pin_button [or `None` if the encoder's button is handled differently or not at all], True for an inversed encoder)
 ```python
 #GPIO Encoder
 encoder_handler.pins = ((board.GP17, board.GP15, board.GP14, False), (encoder 2 definition), etc. )
@@ -62,6 +62,8 @@ encoder_handler.map = [(( KC.VOLD, KC.VOLU, KC.MUTE),(encoder 2 definition), etc
 
 4. Encoder methods on_move_do and on_button_do can be overwritten for complex use cases
 
+
+
 ### Note on split keyboard usage
 If your second side has an encoder as well. Set it up the exact same but when you set up the module pass True in to tell the encoder module that this is a split side. Then when you turn the knob it will send it over to the main side. Example
 ```python
@@ -82,7 +84,7 @@ import board
 from kmk.kmk_keyboard import KMKKeyboard
 from kmk.consts import UnicodeMode
 from kmk.keys import KC
-from kmk.matrix import DiodeOrientation
+from kmk.scanners import DiodeOrientation
 from kmk.modules.layers import Layers
 from kmk.modules.encoder import EncoderHandler
 
@@ -175,3 +177,5 @@ encoder_handler.map = ( ((KC.UP, KC.DOWN, KC.MUTE),), # Standard
 
 if __name__ == "__main__":
     keyboard.go()
+
+
--- a/docs/encoder.md
+++ b/docs/encoder.md
@@ -24,7 +24,7 @@ encoder_handler = EncoderHandler()
 keyboard.modules = [layers, modtap, encoder_handler]
 ```
 
-2. Define the pins for each encoder (pin_a, pin_b, pin_button [or `None` if the encoder's button is handled differently or not at all], True for an inversed encoder)
+2. Define the pins for each encoder (pin_a, pin_b, pin_button, True for an inversed encoder)
 ```python
 #GPIO Encoder
 encoder_handler.pins = ((board.GP17, board.GP15, board.GP14, False), (encoder 2 definition), etc. )
@@ -62,6 +62,18 @@ encoder_handler.map = [(( KC.VOLD, KC.VOLU, KC.MUTE),(encoder 2 definition), etc
 
 4. Encoder methods on_move_do and on_button_do can be overwritten for complex use cases
 
+### Note on split keyboard usage
+If your second side has an encoder as well. Set it up the exact same but when you set up the module pass True in to tell the encoder module that this is a split side. Then when you turn the knob it will send it over to the main side. Example
+```python
+from kmk.modules.encoder import EncoderHandler
+encoder_handler = EncoderHandler(True)
+keyboard.modules = [layers, modtap, encoder_handler]
+```
+Now there are a few limatations at this time.
+* only works with UART ATM (ble should be easy)
+* It will always be on the base layer.
+
+
 ## Full example (with 1 encoder)
 
 ```python
@@ -70,7 +82,7 @@ import board
 from kmk.kmk_keyboard import KMKKeyboard
 from kmk.consts import UnicodeMode
 from kmk.keys import KC
-from kmk.scanners import DiodeOrientation
+from kmk.matrix import DiodeOrientation
 from kmk.modules.layers import Layers
 from kmk.modules.encoder import EncoderHandler
 
--- a/kmk/modules/encoder.py
+++ b/kmk/modules/encoder.py
@@ -5,6 +5,7 @@ import digitalio
 from supervisor import ticks_ms
 
 from kmk.modules import Module
+from kmk.modules.split import Split
 
 # NB : not using rotaryio as it requires the pins to be consecutive
 
@@ -206,10 +207,11 @@ class I2CEncoder(BaseEncoder):
 
 
 class EncoderHandler(Module):
-    def __init__(self):
+    def __init__(self, is_split=False):
         self.encoders = []
         self.pins = None
         self.map = None
+        self.is_split = is_split
 
     def on_runtime_enable(self, keyboard):
         return
@@ -219,6 +221,10 @@ class EncoderHandler(Module):
 
     def during_bootup(self, keyboard):
         if self.pins and self.map:
+            if self.is_split:
+                for module in keyboard.modules:
+                    if type(module) is Split:
+                        self.split_module = module
             for idx, pins in enumerate(self.pins):
                 try:
                     # Check for busio.I2C
@@ -253,7 +259,10 @@ class EncoderHandler(Module):
             else:
                 key_index = 1
             key = self.map[layer_id][encoder_id][key_index]
-            keyboard.tap_key(key)
+            if self.is_split:
+                self.split_module.send_key_press(key.code)
+            else:
+                keyboard.tap_key(key)
 
     def on_button_do(self, keyboard, encoder_id, state):
         if state['is_pressed'] is True:
--- a/kmk/modules/split.py
+++ b/kmk/modules/split.py
@@ -7,7 +7,7 @@ from keypad import Event as KeyEvent
 from storage import getmount
 
 from kmk.hid import HIDModes
-from kmk.keys import KC, make_key
+from kmk.keys import make_key
 from kmk.kmktime import check_deadline
 from kmk.modules import Module
 from kmk.scanners import intify_coordinate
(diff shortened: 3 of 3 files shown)


masiullah:
@daysgobye do you have a sample code for the split encoder declarations in the main.py?
 I am using the below but getting an error and pretty sure I am doing wrong here

encoder_handler = EncoderHandler(True)
encoder_handler.pins = ((keyboard.encoder_pin_0, keyboard.encoder_pin_1, None, False), (keyboard.encoder_pin_0, keyboard.encoder_pin_1, None, False),)

encoder_handler.map = (
    ((KC.VOLD, KC.VOLU), (KC.VOLD, KC.VOLU),),
    ((KC.VOLD, KC.VOLU), (KC.VOLD, KC.VOLU),),
)
keyboard.modules.append(encoder_handler)

xs5871:
This PR and some of the discussion is related with #342.

I'll try to keep this short and only address the things that need discussion and aren't purely formal. Don't be discouraged if this conversations goes back and forth a couple of times.

There are two big issues I see:
1. The Encoder module gets entagled with the Split. I'd prefer if the encoder module were kept agnostic of any split presence or logic. If we ever get one-wire communication, the split _has_ to coordinate which sides turn it is to send or receive, and modules can't just put data on the wire whenever they want. We have to be a bit careful not to hack us into a corner.
2. The use of `make_key` only works with keys that have no internal state and no additional handlers attached. There will be people reporting this as bug sooner or later.

How do we solve this?
2. This seems to be easy to fix. Instead of transmitting the resolved key code, transmit encoder ID and direction and feed that into the corresponding `on_move_do` (or a wrapper of that method). That'll also let you keep the internal state of the encoder synchronized.
1. Now to the hard part: how to get state data from the encoder to the split. That's something we discussed in #342. We can introduce another module/extension class method to get and or set state from split. The module doesn't have to know anything about split and the split stays in charge of orchestrating the communication.

daysgobye:
" 2. This seems to be easy to fix. Instead of transmitting the resolved key code, transmit encoder ID and direction and feed that into the corresponding on_move_do (or a wrapper of that method). That'll also let you keep the internal state of the encoder synchronized. "

 Over all I like this. for encoders and it seems like a good move. Because this will also let layers work on the other encoder. I don't know if I am missing the point or something But this would still require both sides encoder module to know about it being split. The main side would have to have the keymap for both. and the other one would not be able to call tap key. 
So I feel like this requires a more invasive change to be done. I was trying to make a simple addition to the systems that are in place and make something that still feels like kmk. I feel like I did that because the work I have done in splits don't care about encoders whatsoever. And instead, encoders care about sending a keypress and splits care about talking between the 2 sides.

That all being said I really do mind redoing this to work in the way you described. 

" The use of make_key only works with keys that have no internal state and no additional handlers attached. There will be people reporting this as bug sooner or later." 

fair. But better then not working at all maybe HAHAHA

"The Encoder module gets entagled with the Split. I'd prefer if the encoder module were kept agnostic of any split presence or logic. If we ever get one-wire communication, the split has to coordinate which sides turn it is to send or receive, and modules can't just put data on the wire whenever they want. We have to be a bit careful not to hack us into a corner."

So correct me if I am wrong but the only other way I can see of handling this is to leave encoders.py alone and make the changes to tap_key. And I can 100% see the logic in that. if the keyboard is split then it knows that it needs to tap a key and get that to the computer. And encoders can continue to just call tap_key because it does not care how that key gets to the computer.

"This PR and some of the discussion is related with https://github.com/KMKfw/kmk_firmware/pull/342."

I know, If that is coming in soon then cool I will redo the work to be in that pr. But it looks like it's kind of moving slowly. And I just wanted this for myself bad and wanted to give it back because I think kmk should have some kind of support for splits with encoders. 

Full disclosure I sell keyboards that are splits that have encoders. And I think kmk is better for users then qmk. So want to make sure the full kmk ecosystem is at least as good as qmks before I move to pushing kmk over qmk to our customers. 

P.S not very good at github so don't know how I am supposed to quote. and thank you so much for putting up with me I really appreciate that you take the time to look over my code and work on kmk.

daysgobye:
@[masiullah](https://github.com/masiullah) So your main side is set up like normal and working correct? 
You don't have to touch main.py on the main side. 
In main.py on the second side all you should have to do is pass True, to tell it its split. 
Then make sure your split is working over UART. BLE is not done yet.

xs5871:
> would still require both sides encoder module to know about it being split. The main side would have to have the keymap for both. and the other one would not be able to call tap key. 

Yes and no. We managed to have our matrix scanners be split "agnostic" by applying an offset (which is set by the split module), something similar could be done for the encoder ID.

> So I feel like this requires a more invasive change to be done.

That's fair, though not objectively quantifiable. Fact is, we have to change some things to get it working, let's find the best way to do it.
 
> fair. But better then not working at all maybe HAHAHA

Sometimes, sure, I don't disagree. But not if other people have to handle support and fix it, even more so if it's forseeable to cause issues.
One of the very cool "features" of KMK is its flexibilty, that's it's super easy to hack, and you can get it to do whatever you want, no matter how jank it is. However, for the sanity of our maintainers, we can't upstream every end-user solution.
The way your PR handles the communication part is generelly fine, it's only the part how you get and set data from the encoder. There will be more modules that need split synchronisation and we have to design a robust and scalable API, unless we want a mess of interdependance to cause us headaches down the line.
Of course that's not entirely up to you, and I'm more than happy to help you get to a point where we can agree on something sensible. It doesn't have to be perfect, just something were I know that I don't have to do the same work again a couple of amounts of time from now.

> So correct me if I am wrong but the only other way I can see of handling this is to leave encoders.py alone and make the changes to tap_key. And I can 100% see the logic in that. if the keyboard is split then it knows that it needs to tap a key and get that to the computer. And encoders can continue to just call tap_key because it does not care how that key gets to the computer.

I wouldn't recommend using tap_key for the aforementioned limitations and necessity to fix it later anyway, but you got the intention. 

> "This PR and some of the discussion is related with #342."
> I know, If that is coming in soon then cool I will redo the work [...]

No call to action. That's only so the other PR gets automatically pinged and references this PR.

> P.S not very good at github so don't know how I am supposed to quote. and thank you so much for putting up with me I really appreciate that you take the time to look over my code and work on kmk.

There's a "Quote Reply" option when you click on the ellipsis button in the comment header.

daysgobye:
> There's a "Quote Reply" option when you click on the ellipsis button in the comment header.

found it thanks

> No call to action. That's only so the other PR gets automatically pinged and references this PR.
> 
So I will assume this means continue for now.

> I wouldn't recommend using tap_key for the aforementioned limitations and necessity to fix it later anyway, but you got the intention.
> 
So bare with me. I want to find a path so I can go bang it out and come back with working code.

I don't have an idea of what you want it to look like. What I know right now is that overall the way I am sending data to from left to right is ok. But I don't think that you like how I get access to the split module. And the data I am sending is wrong (because it does not work for every key code, does not work with layers.....maybe more). 

what I would like your help with:
1. how should I get access to the split module?
2. If you have a way you want this done let me know. I'm shooting in the dark trying to incorporate what we have talked about.

my idea of how we do this:
both sides are told they are split encoders. The right side has no map and will still call do_move.In that function it will in some way get access to the split module > send the encoder: id and encoder direction. Then on the left side, the spilt module will now need to get access to the encoder module and call a function in there that takes an encoder id and direction. Then the left side split module will have 2  encoder maps. and this new function will work just like do_move but it takes the data from the split module to call the correct key from the split encoder map.

I think this will work, as I type this out tho it does not feel right. Maybe its just my coding style but this feels like a worse option because its a one-off thing just for encoders. I much prefer to make things that are more universal. But its not about me at the end of the day. If you feel like this is a better way for the project great. I think you will have a problem with this system because there is more dependence between the modules. I currently don't have an idea of how we can get around that. I don't truly see how the off set could help with this problem. But I don't know the code base as well as you do if you can point me to where to look for a better way of making this "agnostic" to splits I'm all ears.

kdb424:
I chimed in for the chat it Matrix/Discord, so I'll repeat what I said there, here. I believe that splits being cross pollinated with other modules *may* be inevitable. If @xs5871 has a better route, I'm inclined to go with their thoughts on this. If splits *do* need to be pollinated, we must create a standard, must have good reason to include these changes, each module *must* include a custom header, this all *must* be well documented somewhere in one place for every single reserved header for splits, and the code will be subject to a more stringent standard. I had to do something similar I believe for powersave, and I don't think this is a horrible way to do things. Again, if a better idea is proposed that @xs5871 finds more maintainable, I will trust their judgement here as they know the modern code base more so that I do.

xs5871:
> my idea of how we do this: [...]
I'd do the other way around, similar to the way we handle the split matrix. One design concept that we should at least try to work towards to, is that by default both sides run the same configuration. (or at least: it's not necessary to maintain two configurations)
Also: the secondary side _needs_ to execute the actions asociated with keys, in case the keys change the internal state of the keyboard.

Proposal: Both sides have the same encoder map with mappings for left and right. The primary side wouldn't need any adjustments at this point, the secondary side needs an offset to properly map its encoders. The offset may be set automatically by the split module, or manually in configuration.
I don't think you need to store the entire state of one side on the other, should be enough to apply changes through `on_x_do`. To be fair, the encoder module is probably not the best place to start. It's a bit convoluted and relies on too many special cases for my taste. I'd be nicer to have the refactor before the split sync, but I acknowledge that sync is much requested feature, clean code not so much.

I see two ways to shuffle data around that are reasonably consistent with the current design.
* Either there's a dedicated loop in the main loop (similar to `after_matrix_scan`, etc),
* or the split loops over `keyboard.modules` and transfers data if the module so desires.
The first option would elevate the split module to a special case, the second is less obvious in terms of data and control flow. I think I prefer the second option, given that we implement an API that makes it obvious.

Either way, the modules have to expose access methods for the synchronization mechanism to call. I'll cite myself from #342:
> Introduce two new attributes to modules/extensions:
> * The header byte identifying the module, ex. module.SplitID
> * A setter-getter method, ex:
> ```python
> def split_sync(self, keyboard, in_data: bytearray):
>    # de-serialize data from other split and interpret, apply, whatever (if not None).
>    # return serialized own data if there's any, else None.
>    return out_data
> ```
It should be the modules responsibility to know what data it wants synchronized, and how its serialized/deserialized.
Wether that's done in a single method, like in the citation, or in a getter/setter pair; and wether the header/SplitID should be defined in / managed by the module / split / centrally, that's up to whoever implements it. As long as it's consistent of course.

That's a rough draft, not many details for implementation.
By the way, thank you very much for your patience and for putting up with my "let's do it properly" rambling.

johnskiller:
> How do we solve this? 2. This seems to be easy to fix. Instead of transmitting the resolved key code, transmit encoder ID and direction and feed that into the corresponding `on_move_do` (or a wrapper of that method). That'll also let you keep the internal state of the encoder synchronized.


how about encoder only mapping to 3 new key and report coord_id to other side, let KMKkeyboard do the resolve and layer things

PR #393, opened 2022-04-12
Update Officially_Supported_Microcontrollers.md

xs5871:
remove nice!nano from the list of supported contollers.


xs5871:
This PR may be solved/obsoleted by https://github.com/boardsource/bs-python.

klardotsh:
given that my keyboard (keebio Iris) runs on a N!N I'm not 100% sure why we'd remove it from the list here, but I'll defer that decision to @kdb424 et al

kdb424:
KMKpython is old, broken, and not working, which took that board with it. That said, bs-python is adding proper support back @klardotsh

levpopov:
added #394 to address this

PR #396, opened 2022-04-14
CapsWord & CgSwap: Unit tests, document improvements and enabling these modules for TD, LT and MT keys

masiullah:
Below are the updates
Make it work with LT, MT and TD keys as it was a bit restricted
Added unit test cases for CapsWord and CgSwap
Updated documentation for CgSwap

I had to make some changes in layer, tapdance, holdtap and modtap modules to get this working.  Please let me know if this is acceptable. 

Also I saw there are some issues with auto time out of caps word when modtap and layer keys are used within next few seconds and I was not able to find the issue in code (may be I have not understood correctly how the core time outs work in KMK) and think this will work better to a greater degree without having the timeouts


masiullah:
On second thought i am thinking may be tap dance should not respond to capsword and only LT and MT is good enough. I use MT mostly for the home row mods and LT for space with layer switch and not having these keys react to caps word was making capsword ineffective

kdb424 (approved):
I'll leave final review to @xs5871 if they want, otherwise, I'll merge in 7 days if they don't object, or merge first.

xs5871:
It'll take me a couple of days to go through this.
Be warned: I see some inter-module dependence that we have to discuss. I'm certain there's a way around that, which would be preferable.

kdb424:
> It'll take me a couple of days to go through this. Be warned: I see some inter-module dependence that we have to discuss. I'm certain there's a way around that, which would be preferable.

No rush at all. Wanted to give you a chance to block merge, hence the 7 day window. Merging will be blocked until your review is final.

xs5871:
Let's start with a meta review before going into the actual code.
First off: there's a lot going on. For the future, please try to keep PRs within in reasonably reviewable size, both in diff size and functionally. There are a lot of changes, namely docs, refactoring, and (parts of) the unit test, that could be merged without much of a discussion. I leave it to you if you rather split the PR and get some things in sooner/easier (if not: I understand that it's harder to do that in retrospect).

Additionally, I have to guess what the problem is. "It's not working the way I think it should" is most of the time fine for smaller changes, and I can figure it out from the code. If I'm missing or misunderstanding something, let me know.

### CapsWord
That's the one I think I can mostly understand. Delayed keys aren't capitalized, and/or don't register as end-of-word?
First one can be solved by issueing `LSFT` once at the beginning and release it after the word ended, instead of press/releasing it for every key stroke.
The second one would suggest, that maybe instead of applying delayed keys directly, we may want to send them through the `process_key` pipeline.

### cgSwap
I _think_ this could also be solved by applying `process_key` to delayed keys, but to be honest, I don't really get the point. I thought it was to remap physical switches, but that's obviously not the case.

masiullah:
> Let's start with a meta review before going into the actual code. First off: there's a lot going on. For the future, please try to keep PRs within in reasonably reviewable size, both in diff size and functionally. There are a lot of changes, namely docs, refactoring, and (parts of) the unit test, that could be merged without much of a discussion. I leave it to you if you rather split the PR and get some things in sooner/easier (if not: I understand that it's harder to do that in retrospect).
> 
> Additionally, I have to guess what the problem is. "It's not working the way I think it should" is most of the time fine for smaller changes, and I can figure it out from the code. If I'm missing or misunderstanding something, let me know.
> 
> ### CapsWord
> That's the one I think I can mostly understand. Delayed keys aren't capitalized, and/or don't register as end-of-word? First one can be solved by issueing `LSFT` once at the beginning and release it after the word ended, instead of press/releasing it for every key stroke. The second one would suggest, that maybe instead of applying delayed keys directly, we may want to send them through the `process_key` pipeline.
> 
> ### cgSwap
> I _think_ this could also be solved by applying `process_key` to delayed keys, but to be honest, I don't really get the point. I thought it was to remap physical switches, but that's obviously not the case.

You are right on the money. if the other modules can apply the keys through process_key then this issue will be resolved and will not require these changes. If that is possible then I can split this PR to include only UT cases and docs update and there could be another PR to make the change to apply keys via process_key (I don't have much knowledge here otherwise I could definitely take this up)

Please let me know how you want to move with these changes

xs5871:
Split it into two PRs, one for capswords and cg_swap each. Leave all the passing UTs in and mention the UTs you'd like to see pass, but currently don't, in the PR description. I'll probably still request some minor changes, but it'll be more palatable and confined to one topic. We'll get there.

Regarding the `process_key` pipeline: I can help with that. My main concern is that we could potentially introduce infinite loops. I don't think that's very likely, catchable with the right unit test, and edge-cases at worst. Needs some carefull evaluation for sure. (I may also overestimate that, and it's not an issue at all.)

masiullah:
> Split it into two PRs, one for capswords and cg_swap each. Leave all the passing UTs in and mention the UTs you'd like to see pass, but currently don't, in the PR description. I'll probably still request some minor changes, but it'll be more palatable and confined to one topic. We'll get there.
> 
> Regarding the `process_key` pipeline: I can help with that. My main concern is that we could potentially introduce infinite loops. I don't think that's very likely, catchable with the right unit test, and edge-cases at worst. Needs some carefull evaluation for sure. (I may also overestimate that, and it's not an issue at all.)

Thank you. I will create new PR's in the next few days as you suggested and will close this one

PR #397, opened 2022-04-15
 added oleds 

daysgobye:
Hello, another pr on the docket for you lol. This is my oled extension, I have been spending a lot of time on this and I don't think I need to make any more changes in the near future so it's time to let others play with it. 

### peg prefix
you will see that it's prefixed with Peg. Peg is some software I am working on to let you configure your kmk board. The prefix is to denote that this is the OLED extension that peg interacts with. Anyone can use it and make changes to it but if there were 2 OLED extensions I wanted people to know that they are different. This does not seem to be a problem the last OLED extension never got pulled in. Being that not only users will interact with this but also other code some of my design decisions were about that. I have however tried to make this easy to use for the end-user.

### Docs
I kinda did something a little different in the docs by including images. But I wanted to try and make the best docs I could.

### Code
So one of the biggest things I focused on with this extension is to make it work on nrf52xx chips most chips people seem to be using have more rom&ram so if it can fit on the n!n it can work on anything. 

On that note, you will see my GC calls often I have tested the extension quite a bit to make sure it is stable. The next thing I wanted to make sure I did was to make this expandable, I want to add other features as time goes on. I feel like I have a solid base here that we can expand on.  You will see in "returnCurrectRenderText" my plan. I want to be able to do custom logic for each type of reaction and with this layout, I should be able to. WPM might need to do the math here or read a pin for battery percentage. But the user code stays the same in field 0 tell it what to react to > in field 1 gives it the data to display (images or text)

#### how we do images
Due to the memory constraints, I was not able to get the images working in the more traditional manner and have to load them right from the disk. This works well and saves us some ram.

### DIFF
--- a/docs/peg_oled_display.md
+++ b/docs/peg_oled_display.md
@@ -13,6 +13,13 @@ To use this you need to make some changes to your kb.py as well as you main.py I
 * react to battery percentage
 * react to WPM 
 
+## Required Libs
+you need these frozen into your circuitpython or in a lib folder at the root of your drive.
+* [Adafruit_CircuitPython_DisplayIO_SSD1306](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_SSD1306)
+* [Adafruit_CircuitPython_Display_Text](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text)
+* [download .mpy versions from here](https://github.com/adafruit/Adafruit_CircuitPython_Bundle/releases/download/20220415/adafruit-circuitpython-bundle-7.x-mpy-20220415.zip)
+
+
 ## kb.py
 Your chosen board may already have these changes done If not you will need to add them.
 
--- a/docs/peg_oled_display.md
+++ b/docs/peg_oled_display.md
@@ -14,16 +14,16 @@ To use this you need to make some changes to your kb.py as well as you main.py I
 * React to WPM 
 
 ## Required Libs
-you need these frozen into your circuitpython or in a lib folder at the root of your drive.
+You need these frozen into your circuitpython or in a lib folder at the root of your drive.
 * [Adafruit_CircuitPython_DisplayIO_SSD1306](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_SSD1306)
 * [Adafruit_CircuitPython_Display_Text](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text)
-* [download .mpy versions from here](https://github.com/adafruit/Adafruit_CircuitPython_Bundle/releases/download/20220415/adafruit-circuitpython-bundle-7.x-mpy-20220415.zip)
+* [Download .mpy versions from here](https://github.com/adafruit/Adafruit_CircuitPython_Bundle/releases/download/20220415/adafruit-circuitpython-bundle-7.x-mpy-20220415.zip)
 
 
 ## kb.py
 Your chosen board may already have these changes done If not you will need to add them.
 
-you need to add SCL and SDA to your keyboard. The pins the oled are connected to on your controller may not be called "SCL"and "SDA" they could be "GP21" and "GP13" for example. The best way to find out what they are is look at the boards pinout.
+You need to add SCL and SDA to your keyboard. The pins the oled are connected to on your controller may not be called "SCL"and "SDA" they could be "GP21" and "GP13" for example. The best way to find out what they are is look at the boards pinout.
 ```python
     SCL=board.SCL
     SDA=board.SDA
@@ -33,7 +33,7 @@ you need to add SCL and SDA to your keyboard. The pins the oled are connected to
 ## Main.py
 These are the changes that need to be made / added to your main.py
 ### Config
-no mater how you are going to use the oled you need this part
+No mater how you are going to use the oled you need this part
 ```python
     from kmk.extensions.peg_oled_Display import Oled,OledDisplayMode,OledReactionType,OledData
     keyboard = KMKKeyboard()
@@ -47,13 +47,13 @@ So the config for photos is quite simple. Getting the photos maybe not so much. 
 Oled takes 2-3 arguments 
 
 1.  OledData
-    * OledData can take image **or** corner_one,corner_two,corner_three and corner_four
+    * OledData can take image **or** corner_one, corner_two, corner_three and corner_four
     * Every item in OledData has 2 fields 
-    * 0: this is the reaction type right now it can be OledReactionType.LAYER or OledReactionType.STATIC
+    * 0: This is the reaction type right now it can be OledReactionType.LAYER or OledReactionType.STATIC
     * 1: An array of the items you want to show for the reaction. In this example 4 images to switch on the 4 layers
-2. toDisplay this takes a OledDisplayMode TXT or IMG.
-    * this tells the extension to load images or text.
-3. flip Boolean this will simply flip your display.
+2. To display called as "toDisplay=OledDisplayMode.TXT" this takes a OledDisplayMode TXT or IMG.
+    * This tells the extension to load images or text.
+3. Flip called as "flip= Boolean" this will simply flip your display.
 
 
 ```python
@@ -94,7 +94,7 @@ oled_ext = Oled(
 ```
 
 ### Note 
-your oled data can be a variable as shown below with images
+Your oled data can be a variable as shown below with images.
 
 ```python
 oled_display_data=OledData(image={0:OledReactionType.LAYER,1:["1.bmp","2.bmp","1.bmp","2.bmp"]})
--- a/docs/peg_oled_display.md
+++ b/docs/peg_oled_display.md
@@ -4,14 +4,14 @@ To use this you need to make some changes to your kb.py as well as you main.py I
 ### What you can and cant do
 
 #### Can do
-* display images
-* display text
-* set images or text to react to your layer
+* Display images
+* Display text
+* Set images or text to react to your layer
 
 
 #### Cant do yet / on the way
-* react to battery percentage
-* react to WPM 
+* React to battery percentage
+* React to WPM 
 
 ## Required Libs
 you need these frozen into your circuitpython or in a lib folder at the root of your drive.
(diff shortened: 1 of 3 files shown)


kdb424:
Unlike your other module, this is much less invasive, and OLED modules will likely never be so generic to work with absolutely everything. Once tests pass, I'll take a final look at code, but considering your use case, I'm fine having a semi-specialized OLED module like this in the repo.

docs/peg_oled_display.md
```diff
@@ -0,0 +1,97 @@
...
+    * 0: this is the reaction type right now it can be "LAYER" or "STATIC"
+    * 1: the items you want to show for the reaction. In this example 4 images to switch on the 4 layers
+3. toDisplay this takes a string "TXT" or "IMG".
+    * this tells the extension to load images or text.
+4. flip Boolean this will simply flip your display.
+
+![code layout](https://i.imgur.com/MrRYzo8.png)
+
+```python
+oled_ext = oled(keyboard, [{0:"LAYER",1:["1.bmp","2.bmp","3.bmp","4.bmp"]}],toDisplay='IMG',flip=True)
```
klardotsh, line 62: API change: this should not need to take `keyboard` as an init arg, because the Modules and Extensions APIs already pass the keyboard object reference as the first argument to `during_bootup` (https://github.com/KMKfw/kmk_firmware/blob/a210c2bf62c15c70e6af4c42d5b8bf6984bece5b/kmk/kmk_keyboard.py#L422-L424 and https://github.com/KMKfw/kmk_firmware/blob/a210c2bf62c15c70e6af4c42d5b8bf6984bece5b/kmk/kmk_keyboard.py#L429-L431). Initialization routines that require access to the overall keyboard object should be able to defer that access until `on_bootup` is called.

docs/peg_oled_display.md
```diff
@@ -0,0 +1,97 @@
...
+Ok now we get into something that looks a lot more complicated but we will get though it.
+
+Almost everything is the same we are still passing in the keyboard for pin config. We swap toDisplay to "TXT" lets get into that array.
+
+So each item in the array is a corner of the oled display
+
+0. top left
+1. top right
+2. bottom left
+3. bottom right
```
klardotsh, line 80: weird markdown nit: while this reads correctly in raw source, when this MD file is rendered to any other format, it will be interpreted as a numbered list and screw up the indexes (into 1-4). suggest formatting as

```
- element 0: top left
- element 1: top right
...
```

docs/peg_oled_display.md
```diff
@@ -0,0 +1,97 @@
...
+oled takes 3-4 arguments 
+1. keyboard
+    * this is where we get the config from
+2.  array of data
+    * this array can have a length of 1 or 4. For images we use a length of 1.
+    * Every item in the array has 2 fields 
+    * 0: this is the reaction type right now it can be "LAYER" or "STATIC"
+    * 1: the items you want to show for the reaction. In this example 4 images to switch on the 4 layers
+3. toDisplay this takes a string "TXT" or "IMG".
+    * this tells the extension to load images or text.
```
klardotsh, line 56: users should not have to pass magic strings to enable functionality, especially capitalized strings. this uses excess RAM and is extremely error-prone/brittle for end users. enumerated values should be integer enum-like objects/singletons imported from the extension as `from peg_oled_display import EntityToDisplay` (or some such wording), for example https://github.com/KMKfw/kmk_firmware/blob/a210c2bf62c15c70e6af4c42d5b8bf6984bece5b/kmk/extensions/rgb.py#L71-L80

docs/peg_oled_display.md
```diff
@@ -0,0 +1,97 @@
...
+### Photos
+So the config for photos is quite simple. Getting the photos maybe not so much. I will explain.
+
+oled takes 3-4 arguments 
+1. keyboard
+    * this is where we get the config from
+2.  array of data
+    * this array can have a length of 1 or 4. For images we use a length of 1.
+    * Every item in the array has 2 fields 
+    * 0: this is the reaction type right now it can be "LAYER" or "STATIC"
```
klardotsh, line 53: users should not have to pass magic strings to enable functionality, especially capitalized strings. this uses excess RAM and is extremely error-prone/brittle for end users. enumerated values should be integer enum-like objects/singletons imported from the extension as `from peg_oled_display import ReactionType`, for example https://github.com/KMKfw/kmk_firmware/blob/a210c2bf62c15c70e6af4c42d5b8bf6984bece5b/kmk/extensions/rgb.py#L71-L80

docs/peg_oled_display.md
```diff
@@ -0,0 +1,97 @@
...
+    keyboard.extensions.append(oled_ext) 
+
+```
+### Photos
+So the config for photos is quite simple. Getting the photos maybe not so much. I will explain.
+
+oled takes 3-4 arguments 
+1. keyboard
+    * this is where we get the config from
+2.  array of data
```
klardotsh, line 50: I'm making these comments in reverse order, I'm now realizing, but fine, whatever: if the data is this structured, it probably shouldn't be an anonymous array. it should be a class of its own outright. (in desktop Python I'd also suggest a `namedtuple` but we no longer use those here because their implementation [sometimes takes an external dependency](https://github.com/KMKfw/kmk_firmware/pull/203#discussion_r654896112))

kmk/extensions/peg_oledDisplay.py
```diff
@@ -0,0 +1,133 @@
...
+
+class oled(Extension):
+    def __init__(
+        self,
+        board,
+        views,
+        toDisplay: str = 'TXT',
+        oWidth: int = 128,
+        oHeight: int = 32,
+        flip: bool = False,
```
klardotsh, line 20: these inline type hints have to go in comment form rather than inline, sadly. the PEP that implements inline type hints isn't universally implemented in CircuitPython (last we checked in https://github.com/KMKfw/kmk_firmware/pull/226#issuecomment-903316710)

kmk/extensions/peg_oledDisplay.py
```diff
@@ -0,0 +1,133 @@
...
+    ):
+        displayio.release_displays()
+        self._views = views
+        self._toDisplay = toDisplay
+        i2c = busio.I2C(board.SCL, board.SDA)
+        self.width = (oWidth,)
+        self.height = (oHeight,)
```
klardotsh, line 27: since these are never used again (and tuples aren't negligible in memory), don't store these on `self` in tuple form. storing the int representation is fine, but the tuples should be passed to the SSD1306 constructor and then allowed to be garbage collected (or whatever SSD1306 does to them)

klardotsh (dismissed):
Other notes:

- rename `peg_oledDisplay.py` -> `peg_oled_display.py`
- bonus points if we can get a desktop unit test for any of this, but it's not a blocker
- @xs5871 and @kdb424 can dismiss my review at basically any point to unblock merge; I'm going to be out of town all of next week and not on GitHub and don't want to hold this up

daysgobye:
@klardotsh Thank you for looking it over I agree on 9/10 points I'm going to fix it up .

daysgobye:
I think it got it all changed over very good changes I am happy with them.

docs/peg_oled_display.md
```diff
@@ -0,0 +1,109 @@
+# Peg Oled Display
+To use this you need to make some changes to your kb.py as well as you main.py I will break it into two sections for you. 
+
+### What you can and cant do
+
+#### Can do
+* display images
```
kdb424, line 7: I'm being picky, but can we capitalize the lists? It just looks better, and is more consistent with the rest of the docs I believe. This applies to all start of lines that start a new sentence, or bullet point.

kdb424 (changes requested):
Once the changes to the docs formatting is done, it's good by me. I'll let XS have a week or two to nitpick as they are busy, but otherwise, I won't block a merge pending those fixes.

docs/peg_oled_display.md
daysgobye: good catch!

docs/peg_oled_display.md
```diff
@@ -0,0 +1,109 @@
...
+* Display text
+* Set images or text to react to your layer
+
+
+#### Cant do yet / on the way
+* React to battery percentage
+* React to WPM 
+
+## Required Libs
+you need these frozen into your circuitpython or in a lib folder at the root of your drive.
```
kdb424, line 17: Missed a few

docs/peg_oled_display.md
```diff
@@ -0,0 +1,109 @@
...
+* [Adafruit_CircuitPython_DisplayIO_SSD1306](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_SSD1306)
+* [Adafruit_CircuitPython_Display_Text](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text)
+* [download .mpy versions from here](https://github.com/adafruit/Adafruit_CircuitPython_Bundle/releases/download/20220415/adafruit-circuitpython-bundle-7.x-mpy-20220415.zip)
+
+
+## kb.py
+Your chosen board may already have these changes done If not you will need to add them.
+
+you need to add SCL and SDA to your keyboard. The pins the oled are connected to on your controller may not be called "SCL"and "SDA" they could be "GP21" and "GP13" for example. The best way to find out what they are is look at the boards pinout.
```
kdb424, line 26: This as well. Any time there is a new line, follow English rules :p

docs/peg_oled_display.md
```diff
@@ -0,0 +1,109 @@
...
+```python
+    SCL=board.SCL
+    SDA=board.SDA
+```
+
+
+## Main.py
+These are the changes that need to be made / added to your main.py
+### Config
+no mater how you are going to use the oled you need this part
```
kdb424, line 36: This one

docs/peg_oled_display.md
```diff
@@ -0,0 +1,109 @@
...
+So the config for photos is quite simple. Getting the photos maybe not so much. I will explain.
+
+Oled takes 2-3 arguments 
+
+1.  OledData
+    * OledData can take image **or** corner_one,corner_two,corner_three and corner_four
+    * Every item in OledData has 2 fields 
+    * 0: this is the reaction type right now it can be OledReactionType.LAYER or OledReactionType.STATIC
+    * 1: An array of the items you want to show for the reaction. In this example 4 images to switch on the 4 layers
+2. toDisplay this takes a OledDisplayMode TXT or IMG.
```
kdb424, line 54: I'll let this one go because it's a technical word

docs/peg_oled_display.md
```diff
@@ -0,0 +1,109 @@
...
+    * this tells the extension to load images or text.
+3. flip Boolean this will simply flip your display.
```
kdb424, line 56: You get the idea

docs/peg_oled_display.md
daysgobye: let me know if you dont like my solution

xs5871 (commented):
I don't see a reason to block. It's documented and self-contained; even if it is special purpose, it may either differentiate in the future, or provide a reference for others.
And @klardotsh did a better job at nitpicking than I would have.

daysgobye:
Thank you xs. Are we missing anything?

xs5871:
Merge is still blocked. I think @klardotsh and @kdb424 need to re-review or approve the changes?

xs5871 (approved):

xs5871:
I'm sorry, I don't know how I can mark the change requests as resolved, other than dismissing them.

kdb424:
Seems reasonable to me. Thanks for the review and merge!

PR #399, opened 2022-04-16
fix late release of oneshot keys

### LINKED ISSUE #387: Oneshot modifier continues to timeout not upon another key press
When I use my oneshot keys the modifier persists until the timeout is done (I think) rather than until the next key is pressed. This is the case with `ctrl`, `shift` and `gui` (I haven't tried `alt`).

This is the relevant code:

```python
oneshot = OneShot()
keyboard.modules.append(oneshot)
OS_LCTL = KC.OS(KC.LCTL, tap_time=None)
```

when the `tap_time=1000` the result is the same

In a conversation on the Matrix channel xs5871 thought that maybe it was typing too fast and not releasing keys: 

> "I think I found it. The oneshot key releases when the interrupting key releases. If you type fast and don't release the interrupting key before hitting another key, the oneshot key still applies."

However, I've tested this and it is not the case.

Oneshot is a lifesaver for my wrist, so thank you for your help!

xs5871:
Resolves #387 (potentially).
Would be great if @ldsands could verify before merging.

### DIFF
--- a/kmk/modules/oneshot.py
+++ b/kmk/modules/oneshot.py
@@ -29,7 +29,7 @@ class OneShot(HoldTap):
                 state.activated = ActivationType.HOLD_TIMEOUT
             elif state.activated == ActivationType.RELEASED and is_pressed:
                 state.activated = ActivationType.INTERRUPTED
-            elif state.activated == ActivationType.INTERRUPTED and not is_pressed:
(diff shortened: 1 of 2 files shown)


kdb424 (approved):

ldsands:
OneShot seems to be working for me now thanks!

PR #400, opened 2022-04-16
Enhancement: multiple scanners

xs5871:
Support for multiple (i.e. a list of) scanners.
`KMKKeyboard.matrix` can either be a single scanner (support for current implementations), or a list of scanners.
`coord_mapping` for mirrored non-square matrices or multiple scanners has to be defined manually -- there's no general solution. Individual scanner "flips" may be added in the future.

Before merging, confirmation from a split user would be appreciated.


kdb424 (approved):

kdb424:
Seemed to work for me, but can revert if others somehow find an issue.

PR #402, opened 2022-04-17
Feature scanner encoder

xs5871:
Native matrix events from an encoder, because, why not. Only makes sense in conjunction with #400.
Also specifically not meant to compete with or supersede the encoder module.


kdb424 (approved):

ldsands:
OneShot seems to be working for me now thanks!

PR #403, opened 2022-04-18
refactor Key and ModifierKey __call__ methods

rrotter:
- Add tests for ModifierKey chaining
- Rewrite ModifierKey.__call__ for correctness and readability
- __call__ now maintains handlers and meta, like clone()

### DIFF
--- a/tests/test_kmk_keys.py
+++ b/tests/test_kmk_keys.py
@@ -12,8 +12,10 @@ class TestKmkKeys(unittest.TestCase):
                 [
                     KC.HASH,
                     KC.RALT(KC.HASH),
-                    KC.RALT(KC.LSFT)(KC.N3),
+                    KC.RALT(KC.LSFT(KC.N3)),
                     KC.RALT(KC.LSFT),
+                    # Note: this is correct, if unusual, syntax. It's a useful test because it failed silently on previous builds.
+                    KC.RALT(KC.LSFT)(KC.N3),
                     KC.RALT,
                 ]
             ],
@@ -30,7 +32,7 @@ class TestKmkKeys(unittest.TestCase):
             ],
         )
         keyboard.test(
-            'Shift+AltGr+key',
+            'AltGr+Shifted key',
             [(1, True), (1, False)],
             [
                 {
@@ -42,7 +44,7 @@ class TestKmkKeys(unittest.TestCase):
             ],
         )
         keyboard.test(
-            'Shift+AltGr+key, alternate chaining',
+            'AltGr+Shift+key',
             [(2, True), (2, False)],
             [
                 {
@@ -65,8 +67,20 @@ class TestKmkKeys(unittest.TestCase):
             ],
         )
         keyboard.test(
-            'AltGr',
+            'AltGr+Shift+key, alternate chaining',
             [(4, True), (4, False)],
+            [
+                {
+                    KC.N3,
+                    KC.LSFT,
+                    KC.RALT,
+                },
+                {},
+            ],
+        )
+        keyboard.test(
+            'AltGr',
+            [(5, True), (5, False)],
             [
                 {
                     KC.RALT,
--- a/tests/test_kmk_keys.py
+++ b/tests/test_kmk_keys.py
@@ -1,0 +1,87 @@
+import unittest
+
+from kmk.keys import KC, Key, ModifierKey
+from tests.keyboard_test import KeyboardTest
+
+
+class TestKmkKeys(unittest.TestCase):
+    def test_basic_kmk_keyboard(self):
+        keyboard = KeyboardTest(
+            [],
+            [
+                [
+                    KC.HASH,
+                    KC.RALT(KC.HASH),
+                    KC.RALT(KC.LSFT)(KC.N3),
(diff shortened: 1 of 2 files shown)


tests/test_kmk_keys.py
```diff
@@ -0,0 +1,87 @@
...
+
+class TestKmkKeys(unittest.TestCase):
+    def test_basic_kmk_keyboard(self):
+        keyboard = KeyboardTest(
+            [],
+            [
+                [
+                    KC.HASH,
+                    KC.RALT(KC.HASH),
+                    KC.RALT(KC.LSFT)(KC.N3),
```
xs5871, line 15: Not a fan of that syntax, but at least `KC.RALT(KC.LSFT(KC.N3))` passes the same test.
No need to necessarily change it.

xs5871 (approved):
Looks good to go.

tests/test_kmk_keys.py
rrotter: This weird syntax here is intentional, I was trying to exercise a particular edge case. These three lines _should_ be synonymous:

- KC.RALT(KC.HASH)
- KC.RALT(KC.LSFT(KC.N3))
- KC.RALT(KC.LSFT)(KC.N3)

But, the third one actually fails on the current master branch. Maybe I should have added a comment this effect inline.

kdb424: I fail to see how the third line should not fail. The syntax looks invalid, and even if it's not, it's confusing, and should be avoided IMHO

rrotter: This this valid?
````
>>> KC.RALT(KC.LSFT)
ModifierKey(code=-1, has_modifiers={64, 2})
````
If so, what should this do?
````
>>> ALT_SHIFT = KC.RALT(KC.LSFT)
>>> ALT_SHIFT(KC.N3)
````
I'd argue that it should produce:
`Key(code=32, has_modifiers={64, 2})`
Currently it produces:
`Key(code=32, has_modifiers={-1})`
Which I think is pretty heavily broken.

I'm not arguing it's a good idea, I was just writing a test that exercises an edge case that was producing an undesirable result. Nevertheless, I'm happy to change the test if you'd like.

Comment (kdb424) [reply tests/test_kmk_keys.py:15]:
```
>>> ALT_SHIFT = KC.RALT(KC.LSFT)
>>> ALT_SHIFT(KC.N3)
```
This should fail I believe. The reason that `KC.RALT(KC.HASH)` this will pass is because we handled the cases for shifted keys by design. `KC.RALT(KC.LSFT(KC.N3))` This will work because mod+key (and therefor shifted keys) and keys passed to other keys. Meanwhile `KC.RALT(KC.LSFT)(KC.N3)` this is a key, with another key passed to it, and then just an arbitrary extra (), which seems syntactically incorrect at best. I'm not saying that we can't add support for something like your second example in some way, but the test case I noted should fail, seems like it would fail correctly, and we should handle what your intent is differently under the hood. If you feel I'm wrong, please explain why. I rarely claim to be correct, but I am questioning why you think that syntax is valid, not the idea of what you showed in `ALT_SHIFT(KC.N3)` this example. The idea, to be clear, is valid.

Comment (rrotter) [reply tests/test_kmk_keys.py:15]: Which parentheses are arbitrary?
`A(B)(C)`
is equivalent to
```
X = A(B)
X(C)
```
`A(B)C` on the other hand wouldn't even parse.

In the case here…
`KC.RALT(KC.LSFT)(KC.N3)` is equivalent to:
```
X = KC.RALT(KC.LSFT)
X(KC.N3)
```

Whether it is on one line or two, this code does the following:
1. `KC.RALT(KC.LSFT)` returns a ModifierKey with both mods in has_modifiers. I assumed that this is correct/intentional. If this behavior is wrong or unintended, then I can understand the confusion. This case seems to be handled intentionally in the code, as the special case of setting the keycode to -1 is specific to this situation (calling a ModifierKey w/ another ModifierKey as the parameter)
2. The ModifierKey returned from `KC.RALT(KC.LSFT)` is called with KC.N3 as the parameter. If this does anything at all, it should return the parameter key with the content of the ModifierKey's has_modifiers added to it. Instead, it only adds the `-1` fake keycode.

I mean, do we agree that returning `Key(code=32, has_modifiers={-1})` here is undesirable?

kdb424: TIL. Yeah, you are correct. Thanks for clearing that up.

rrotter: Of course, and thanks for reviewing the PR. This part of the code was/is very confusing. There is clearly a reason for the `# FIXME this is atrocious to read…` note in the section I replaced.

kdb424:
I'm also requesting @klardotsh to read that section over as he was the primary author of the original code, and the message listed above.

kdb424 (approved):

rrotter:
I pushed an update to the tests. I hope that's a little clearer. Not sure what else to do short of adding a link to this thread, but I figure that is what `git blame` is for.

rrotter:
rebased to current master

kdb424:
Merging for now since klardotsh is off the grid for a while, but he can still chime in and we can make any adjustments needed later.

Issue #369, opened 2022-03-28
Stack overflow in maybe_make_shifted_key


rrotter:
Use of shifted keys produces a reproducible stack overflow in maybe_make_shifted_key.

Testing on d4e72b9 (current as of this writing), CircuitPython 7.2.3, on Pi Pico.

main.py:
```import board

from kmk.kmk_keyboard import KMKKeyboard
from kmk.keys import KC
from kmk.matrix import DiodeOrientation

keyboard = KMKKeyboard()

keyboard.col_pins = (
    board.GP3, board.GP4, board.GP5, board.GP6, board.GP7, board.GP8,
    board.GP9, board.GP10, board.GP11, board.GP12, board.GP13, board.GP14, board.GP15,
)

keyboard.row_pins = (
    board.GP16,
    board.GP17,
    board.GP18,
    board.GP19,
    board.GP20,
)
keyboard.diode_orientation = DiodeOrientation.ROW2COL

keyboard.keymap = [
    [KC.LABK,  KC.RABK,  KC.LPRN,  KC.RPRN,  KC.CIRC,  KC.DLR,   KC.GRAVE, KC.HASH,  KC.EXLM,  KC.SLASH, KC.LBRC,  KC.RBRC,  KC.BSLASH,
     KC.NO,    KC.Q,     KC.W,     KC.F,     KC.P,     KC.B,     KC.TAB,   KC.J,     KC.L,     KC.U,     KC.Y,     KC.MINUS, KC.EQUAL,
     KC.ESC,   KC.A,     KC.R,     KC.S,     KC.T,     KC.G,     KC.BSPC,  KC.M,     KC.N,     KC.E,     KC.I,     KC.O,     KC.QUOTE,
     KC.LSFT,  KC.Z,     KC.X,     KC.C,     KC.D,     KC.V,     KC.NO,    KC.K,     KC.H,     KC.COMMA, KC.DOT,   KC.UP,    KC.RSFT,
     KC.NO,    KC.NO,    KC.NO,    KC.LALT,  KC.LCMD,  KC.ENTER, KC.SPACE, KC.NO,    KC.RCMD,  KC.RALT,  KC.LEFT,  KC.DOWN,  KC.RIGHT,],
]

if __name__ == '__main__':
    keyboard.go()
```

100% reliably produces the following stack overflow (yes, irrespective of using the recommended settings in boot.py):
```
Traceback (most recent call last):
  File "main.py", line 50, in <module>
[...]

PR #405, opened 2022-04-18
Index and get keys

morrijr:
Allow the KC collection to be accessed;

KC.KEY_NAME
KC['KEY_NAME']
KC.get('KEY_NAME')

This will allow requesting of keys by names that aren't valid method names.

Also, added tests to prove the keys created are the same instance, request order doesn't matter and case (where appropriate) isn't relevant. Remove code that could have caused stack overflow (recursion).

### DIFF
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -368,7 +368,7 @@ class KeyAttrDict(AttrDict):
         return self.__getattr__(key, depth=depth + 1)
 
 
-# Global state, will be filled in througout this file, and
+# Global state, will be filled in throughout this file, and
 # anywhere the user creates custom keys
 KC = KeyAttrDict()
 
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -58,323 +58,323 @@ def maybe_make_consumer_key(candidate, code, names):
 
 
 class KeyAttrDict(AttrDict):
-    def __getitem__(self, key):
-        key = key.upper()
-        try:
-            return super().__getitem__(key)
-        except Exception:
-            return self.__try_create_key(key)
+    def __getattr__(self, key):
+        return self.__getitem__(key)
 
     def get(self, key, default=None):
         try:
-            return self[key]
+            return self.__getitem__(key)
         except Exception:
             return default
 
-    def __try_create_key(self, key):
-        # Try all the other weird special cases to get them out of our way:
-        # This need to be done before or ALPHAS because NO will be parsed as alpha
-        # Internal, diagnostic, or auxiliary/enhanced keys
+    def __getitem__(self, key):
+        key = key.upper()
+        try:
+            return super(KeyAttrDict, self).__getitem__(key)
+        except Exception:
+            # Try all the other weird special cases to get them out of our way:
+            # This need to be done before or ALPHAS because NO will be parsed as alpha
+            # Internal, diagnostic, or auxiliary/enhanced keys
 
-        # NO and TRNS are functionally identical in how they (don't) mutate
-        # the state, but are tracked semantically separately, so create
-        # two keys with the exact same functionality
-        if key in ('NO', 'XXXXXXX'):
-            return make_key(
-                names=('NO', 'XXXXXXX'),
-                on_press=handlers.passthrough,
-                on_release=handlers.passthrough,
-            )
-        if key in ('TRANSPARENT', 'TRNS'):
-            return make_key(
-                names=('TRANSPARENT', 'TRNS'),
-                on_press=handlers.passthrough,
-                on_release=handlers.passthrough,
-            )
-        # Basic ASCII letters/numbers don't need anything fancy, so check those
-        # in the laziest way
-        if key in ALL_ALPHAS:
-            return make_key(
-                code=4 + ALL_ALPHAS.index(key),
-                names=(
+            # NO and TRNS are functionally identical in how they (don't) mutate
+            # the state, but are tracked semantically separately, so create
+            # two keys with the exact same functionality
+            if key in ('NO', 'XXXXXXX'):
+                return make_key(
+                    names=('NO', 'XXXXXXX'),
+                    on_press=handlers.passthrough,
+                    on_release=handlers.passthrough,
+                )
+            if key in ('TRANSPARENT', 'TRNS'):
+                return make_key(
+                    names=('TRANSPARENT', 'TRNS'),
+                    on_press=handlers.passthrough,
+                    on_release=handlers.passthrough,
+                )
+            # Basic ASCII letters/numbers don't need anything fancy, so check those
+            # in the laziest way
+            if key in ALL_ALPHAS:
+                return make_key(
+                    code=4 + ALL_ALPHAS.index(key),
+                    names=(
+                        key,
+                        key.lower(),
+                    ),
+                )
+            if key in ALL_NUMBERS or key in ALL_NUMBER_ALIASES:
+                try:
+                    offset = ALL_NUMBERS.index(key)
+                except ValueError:
+                    offset = ALL_NUMBER_ALIASES.index(key)
+
+                return make_key(
+                    code=30 + offset,
+                    names=(ALL_NUMBERS[offset], ALL_NUMBER_ALIASES[offset]),
+                )
+            if key in ('RESET',):
+                return make_key(names=('RESET',), on_press=handlers.reset)
+            if key in ('BOOTLOADER',):
+                return make_key(names=('BOOTLOADER',), on_press=handlers.bootloader)
+            if key in ('DEBUG', 'DBG'):
+                return make_key(
+                    names=('DEBUG', 'DBG'),
+                    on_press=handlers.debug_pressed,
+                    on_release=handlers.passthrough,
+                )
+            if key in ('BKDL',):
+                return make_key(
+                    names=('BKDL',),
+                    on_press=handlers.bkdl_pressed,
+                    on_release=handlers.bkdl_released,
+                )
+            if key in ('GESC', 'GRAVE_ESC'):
+                return make_key(
+                    names=('GESC', 'GRAVE_ESC'),
+                    on_press=handlers.gesc_pressed,
+                    on_release=handlers.gesc_released,
+                )
+
+            # A dummy key to trigger a sleep_ms call in a sequence of other keys in a
+            # simple sequence macro.
+            if key in ('MACRO_SLEEP_MS', 'SLEEP_IN_SEQ'):
+                return make_argumented_key(
+                    validator=key_seq_sleep_validator,
+                    names=('MACRO_SLEEP_MS', 'SLEEP_IN_SEQ'),
+                    on_press=handlers.sleep_pressed,
+                )
+            if key in ('UC_MODE_NOOP', 'UC_DISABLE'):
+                return make_key(
+                    names=('UC_MODE_NOOP', 'UC_DISABLE'),
+                    meta=UnicodeModeKeyMeta(UnicodeMode.NOOP),
+                    on_press=handlers.uc_mode_pressed,
+                )
+            if key in ('UC_MODE_LINUX', 'UC_MODE_IBUS'):
+                return make_key(
+                    names=('UC_MODE_LINUX', 'UC_MODE_IBUS'),
+                    meta=UnicodeModeKeyMeta(UnicodeMode.IBUS),
+                    on_press=handlers.uc_mode_pressed,
+                )
+            if key in ('UC_MODE_MACOS', 'UC_MODE_OSX', 'US_MODE_RALT'):
+                return make_key(
+                    names=('UC_MODE_MACOS', 'UC_MODE_OSX', 'US_MODE_RALT'),
+                    meta=UnicodeModeKeyMeta(UnicodeMode.RALT),
+                    on_press=handlers.uc_mode_pressed,
+                )
+            if key in ('UC_MODE_WINC',):
+                return make_key(
+                    names=('UC_MODE_WINC',),
+                    meta=UnicodeModeKeyMeta(UnicodeMode.WINC),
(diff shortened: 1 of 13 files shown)


morrijr:
I don't know why the 'typos' change came through with this one. It shouldn't have. If you have an issue with that, leave me a note and close this PR and I'll see if I can figure it out.

xs5871:
I can look into this, but it'll have to wait a couple of days.
Out of curiosity, could you elaborate on what is the point? By invalid _method_ names, I assume you mean _attribute_, and something like `KC['*']`?

morrijr:
Sorry, python isn't one of the languages I usually code in, but yes, I'm after a way to do KC['*'].

With regards to the eventual aim, I'm thinking of adding to the init routines to loop over the keymap and replace '*' with KC['*'] (for example). I think having;

keyboard.keymap = [ [ '[' ] ]

rather than

keyboard.keymap = [ [ KC.LBRACKET ] ]

It's not much, but I found it was visually easier and certainly helped to keep the map aligned 'prettily' to not have such widely varying 'column' widths... I think it would be an improvement?

morrijr:
Changes wise it's not as bad as the diff makes it look. Change from `__getattr__` to `__getitem__` in KeyAttrDict (we still inherit the `__getattr__` from AttrDict).  Addition of the `get` and the extraction of the original key generation. Make the key generation return the key rather than requiring another search into the dictionary. The removal of the another lookup means no need to guard against a recursive call.

rrotter (commented):
I like what you're doing here, I was looking for a way to do the same (reference special chars by char and not the name), which would clean up my own main.py quite a bit.

If `KC[str]` is a public API, it should really work anywhere it is called rather than only functioning under certain circumstances (namely, that the key has already been referenced once). This also appears to deal with the edge case that KC.A == KC.a is True and KC.a == KC.A throws an error.

Disclaimer: I'm not a maintainer, I can claim no authority here
Edit: typo

rrotter:
Just want to add/clarify, I like that this makes `KC['KEY_NAME']` work more consistently, I'm just not sure what the purpose of `KC.get('KEY_NAME')` is.

kdb424:
> Just want to add/clarify, I like that this makes `KC['KEY_NAME']` work more consistently, I'm just not sure what the purpose of `KC.get('KEY_NAME')` is.

The get is a requested feature for some niche things, and I know that at least one person has needed that function for reverse lookups. It's uses may be rare, but apparently there are valid uses for it.

rrotter:
Oh shoot, trying this on actual hardware, this is a lot of stack pressure. I'm seeing the same issue that was mostly fixed in #369:

```
main.py output:
  File "main.py", line 32, in <module>
  File "kmk/types.py", line 11, in __getattr__
  File "kmk/keys.py", line 66, in __getitem__
  File "kmk/keys.py", line 355, in __try_create_key
  File "kmk/keys.py", line 35, in first_truthy
  File "kmk/keys.py", line 317, in <lambda>
  File "kmk/keys.py", line 28, in left_pipe_until_some
  File "kmk/keys.py", line 52, in maybe_make_shifted_key
  File "kmk/keys.py", line 677, in make_shifted_key
  File "kmk/types.py", line 11, in __getattr__
  File "kmk/keys.py", line 66, in __getitem__
  File "kmk/keys.py", line 355, in __try_create_key
  File "kmk/keys.py", line 35, in first_truthy
  File "kmk/keys.py", line 197, in <lambda>
  File "kmk/keys.py", line 28, in left_pipe_until_some
  File "kmk/keys.py", line 42, in maybe_make_mod_key
  File "kmk/keys.py", line 673, in make_mod_key
  File "kmk/keys.py", line 663, in make_key
RuntimeError: pystack exhausted
```

My personal main.py is a mess that is using too much memory, but adding at least 2 frames to the stack on every `__getattr__` is still likely to cause trouble eventually (and, speaking selfishly, will break my keyboard).

Not sure what to suggest here. There are a few different avenues to reducing the number of stack frames.

kdb424:
Fighting the stack is unfortunately why we don't have as many "nice to haves" as we want, and strip things out. There's a few tricks to slim down data structures that could be done. Using consts has been one of these things we have done, but we may need to resort to more tricks on the core of KMK. We try to keep it as "open and free" as possible, but modules/extensions are where we expect most users to be, and can make the core a little harder to modify if it means we get better memory/stack performance.

morrijr:
`__getattr__` was already there as was all the lambda. `__getitem__` and the separate function is new, I admit, although (and not knowing enough python to be sure) I suspect they're there anyway, but internal to the implementation.  The removal of the call to `return self.__getattr__(key, depth=depth + 1)` should I thought have mitigated the additional call.

@rrotter what hardware was this on? I've only got a RP2040 to play with. I don't suppose you could give me a test case that demonstrates the two calls? Is it a combined/composite key?

The only *other* way I could think to solve this is by statically generating all the keys in that list. Outside of the main python, before it's copied to the micros. Removing the dynamic stuff. Not actually sure if it'd save anything.

morrijr:
Actually, I see `make_shifted_key` so I'm guessing I'm right. I'll take a look, see if I can change the make_shifted_key use the `__try_create_key` directly.

I can make `__try_create_key` static. Would that help? I don't think so, but maybe?

morrijr:
@rrotter can you give this version a try? See if that's removed enough stack frames?

morrijr:
Can somebody help me with the reason for the test failure?

`./kmk/keys.py:176:33: BLK100 Black would make changes.`

How to replicate locally would be nice to know :/

rrotter:
> Can somebody help me with the reason for the test failure?
> 
> `./kmk/keys.py:176:33: BLK100 Black would make changes.`
> 
> How to replicate locally would be nice to know :/

`pip install black` and then run black on the file to run the linter. `make test` runs black in test mode so it tests for lint but doesn't make any changes.

````
tardis:kmk_firmware rrotter$ black kmk/keys.py
reformatted kmk/keys.py

All done! ✨ 🍰 ✨
1 file reformatted.
tardis:kmk_firmware rrotter$ git diff
diff --git a/kmk/keys.py b/kmk/keys.py
index 468db7c..fcb86f7 100644
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -173,7 +173,9 @@ class KeyAttrDict(AttrDict):
                     on_press=handlers.uc_mode_pressed,
                 )
             if key in ('HID_SWITCH', 'HID'):
-                return make_key(names=('HID_SWITCH', 'HID'), on_press=handlers.hid_switch)
+                return make_key(
+                    names=('HID_SWITCH', 'HID'), on_press=handlers.hid_switch
+                )
             if key in ('BLE_REFRESH'):
                 return make_key(names=('BLE_REFRESH',), on_press=handlers.ble_refresh)
````

**Comment by rrotter** [comment]
> @rrotter can you give this version a try? See if that's removed enough stack frames?

That fixes the stack overflow for me. I'm also using an RP2040. There is one other issue that I'm seeing. I'm not sure why, but this is breaking array style lookups on custom created keycodes for me, which work fine on the current master branch. Strangely, attr stile lookups are still working.

master branch:
````
Adafruit CircuitPython 7.2.5 on 2022-04-06; Raspberry Pi Pico with rp2040
>>> import board
>>>
>>> from kmk.kmk_keyboard import KMKKeyboard
>>> from kmk.keys import KC, make_key
>>>
>>> make_key(KC.N2.code, names=('EURO','€'), has_modifiers={KC.LSFT.code, KC.ROPT.code})
Key(code=31, has_modifiers={2, 64})
>>> KC['€']
Key(code=31, has_modifiers={2, 64})
>>> KC['EURO']
Key(code=31, has_modifiers={2, 64})
>>> KC.EURO
Key(code=31, has_modifiers={2, 64})
````

index_and_get_keys branch:
````
Adafruit CircuitPython 7.2.5 on 2022-04-06; Raspberry Pi Pico with rp2040
>>> from kmk.kmk_keyboard import KMKKeyboard
>>> from kmk.keys import KC, make_key
>>>
>>> make_key(KC.N2.code, names=('EURO','€'), has_modifiers={KC.LSFT.code, KC.ROPT.code})
Key(code=31, has_modifiers={2, 64})
>>> KC['€']
  File "<stdin>", line 1, in <module>
  File "kmk/keys.py", line 377, in __getitem__
ValueError: Invalid key
>>> KC['EURO']
  File "<stdin>", line 1, in <module>
  File "kmk/keys.py", line 377, in __getitem__
ValueError: Invalid key
>>> KC.EURO
Key(code=31, has_modifiers={2, 64})
````

**Comment by morrijr** [comment]
Thanks, I'll add a test and take a look :)

**Comment by morrijr** [comment]
def test_custom_key(self):
        created = make_key(
            KC.N2.code,
            names=('EURO', '€'),
            has_modifiers={KC.LSFT.code, KC.ROPT.code}
        )
        assert created.code == 31
        assert created.has_modifiers == {2, 64,}
        assert created is KC['EURO']
        assert created is KC['€']

Test is green :(

Any other thoughts? I'll add the updated changes...

**Comment by rrotter** [comment]
I'll take a look at this tonight. Are you able to reproduce the issue I'm seeing using the REPL?

**Comment by morrijr** [comment]
Yes. What I find interesting is adding print statements to `__getattr__`, `get` and `__getitem__` like so;

`    def __getitem__(self, key):
        print(f'__getitem__ {key}')`

>>> from kmk.keys import KC
>>> KC.A
__getattr__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
__getitem__ A
Key(code=4, has_modifiers=None)

I *really* don't understand why this is happening or how to debug!

**Comment by rrotter** [comment]
It appears to be in a recursion loop, where `__getitem__` is calling itself repeatedly. This blows up the stack, but is caught by an overly broad `except Exception`, which should probably be an `except KeyError`.

Still don't know why this is happening, though I can confirm it doesn't appear to occur on the desktop.

File: kmk/keys.py [hunk 1]
```diff
@@ -62,112 +58,127 @@ def maybe_make_consumer_key(candidate, code, names):
...
-                names=('GESC', 'GRAVE_ESC'),
-                on_press=handlers.gesc_pressed,
-                on_release=handlers.gesc_released,
-            )
+            return default
+
+    def __getitem__(self, key):
+        key = key.upper()
+        try:
+            return super(KeyAttrDict, self).__getitem__(key)
```
rrotter, line 73: I think Line 77 is just calling line 70 again on CircuitPython. This eventually gets us to line 82 only b/c line 74 catches the RuntimeError (when it should be catching a KeyError).

morrijr: Agreed. I might not have time today, but I'll look into it as soon as I can.

morrijr:
OK, this passes all the tests and your example in the repl.

I've stopped the inheritance KeyAttrDict -> AttrDict -> dict and just given KeyAttrDict its own (internal) cache. Recursion is impossible now :)

The stack calls are kept to a minimum.

Thoughts?

morrijr:
OK, this passes all the tests and your example in the repl.

I've stopped the inheritance KeyAttrDict -> AttrDict -> dict and just given KeyAttrDict its own (internal) cache. Recursion is impossible now :)

The stack calls are kept to a minimum.

Thoughts?

rrotter:
Looks good. Code is working for me. Only concern I have is that this maintains the possibility to make keys with lower case names but makes it impossible to access them.

Also, feel free to use this if you want or ignore it, but on my copy of your branch I added a tiny patch that will remove another frame (https://github.com/rrotter/kmk_firmware/commits/index_and_get_keys).

xs5871:
Let me know when the PR is ready for a review. e814 was looking good, some minor things, I really liked the unittests structure/syntax/readability -- that's something we should improve on.
The most recent commit makes it much harder to keep track of what's changed; that may warrant a second PR.

xs5871:
Let me know when the PR is ready for a review. e814 was looking good, some minor things, I really liked the unittests structure/syntax/readability -- that's something we should improve on.
The most recent commit makes it much harder to keep track of what's changed; that may warrant a second PR.

morrijr:
Happy to do whatever to progress. **Quite** happy to throw this branch/PR away and redo it in smaller, more logical steps now I know what the end state should look like? Shouldn't take more than an hour or so. All the changes are there, just a matter of committing more... regularly/logically.

The last change really came about because of my realisation (I'd **read** but not **understood** the importance of) the fact that names were defined as tuples. Do you know why the names were defined as tuples rather than lists? I'm not overjoyed wrapping all the maybe_make_ in lambda. I need to figure out if there's a way of referencing a function in a collection (I can do it in other languages).

Anyway. Happy to close/junk this and redo. I don't think I'd take the e814; rrotter is correct that you could then call make_key with only a lowercase name and it not work. I suspect people wouldn't (generally), but it would trip **somebody** up **eventually**.

xs5871:
> Do you know why the names were defined as tuples rather than lists?

they're immutable and as a consequence use less memory than lists.

> I'm not overjoyed wrapping all the maybe_make_ in lambda. I need to figure out if there's a way of referencing a function in a collection (I can do it in other languages).

Not quite sure what you mean or want to do. Something like this?:
```python
x = [
    [fn0, arg00, arg01],
    [fn1, args00, arg01, arg02],
]

for _ in x:
    fn = _[0]
    args = _[1:]
    fn(*args)
```

> Anyway. Happy to close/junk this and redo. I don't think I'd take the e814; rrotter is correct that you could then call make_key with only a lowercase name and it not work. I suspect people wouldn't (generally), but it would trip **somebody** up **eventually**.

*thumbs-up-emoji*

morrijr:
Going to close this and redo with (hopefully!) more/better understanding of the Python.

PR #407, opened 2022-04-20
fixes scanner init and bi-directional communication in split module

### LINKED ISSUE #406: Bug on Split keyboards (mirror or left only for both sides)
So after the OneShot keys were fixed (thanks by the way). I updated everything and found that my halves are now only will use the left half of the keymap. I have tried several different variations/combinations of using the split options below. No combinations of these variables allows the right half to be the right half. I did rename the usb storage to `SOFLEV2L` and `SOFLEV2R` I have not removed the L and R because I cannot figure out an easy way to do it (I recently switched to Linux full time). So the EE hands feature might be involved with this somehow.

```python
target_left=True, 
split_side=SplitSide.LEFT,
split_flip=True,
```

If I had to guess it was probably the changes made to `split.py` [here](https://github.com/KMKfw/kmk_firmware/pull/402/files#diff-60be404174a94d655cb4beae62a821cc693dcab6f99b55da3917c129412b5ce0) or [here](https://github.com/KMKfw/kmk_firmware/commit/7431192e95bc2a1a0d23cdab076187ec4d9cfc23) that messed it up. I can confirm that it works fine as of the commit made on Apr 5th (7b92596). I've tried to figure out what caused it but I havn't made any progress.

I'll post my `kb.py` and `main.py` code below.

Any help would be appreciated.

PS

I don't think this is as useful but just in case. Whenever a key is simultaneously pressed at the same time as another key on the other side, this is the error that occurs.

```python
KeyNotPressed(ic=13)
  File "main.py", line 120, in <module>
  File "kmk/kmk_keyboard.py", line 422, in go
  File "kmk/kmk_keyboard.py", line 476, in _main_loop
  File "kmk/kmk_keyboard.py", line 116, in _handle_matrix_report
  File "kmk/kmk_keyboard.py", line 179, in _on_matrix_changed
KeyError: 13
```

xs5871:
Resolves #406.
There where two independent bugs in the split module: one introduced recently by #400, and one that's a month old by #366.
The older PR claims to fix something, but doesn't say what, and completely disables communication from target to initiator. If anyone finds that my fix is a regression, please enlighten me.
(ps: tested with hardware UART)

### DIFF
--- a/kmk/modules/split.py
+++ b/kmk/modules/split.py
@@ -161,7 +161,10 @@ class Split(Module):
             keyboard.coord_mapping = tuple(cm)
 
         if self.split_side == SplitSide.RIGHT:
-            keyboard.matrix.offset = self.split_offset
+            offset = self.split_offset
+            for matrix in keyboard.matrix:
+                matrix.offset = offset
+                offset += matrix.key_count
 
     def before_matrix_scan(self, keyboard):
         if self.split_type == SplitType.BLE:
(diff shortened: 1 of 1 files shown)


ldsands (commented):
I also tested this and as far as I can tell it works great. Thank you!

kdb424:
I'd ask @xs5871 to review this, but I already know they think it's ok. Merged

kdb424 (approved):

PR #408, opened 2022-04-20
Updated OneShot doc with example OneShot Modifier Combination

ldsands:
This adds an explanation, a short example and a longer block of code that contains everything needed to allow for all (left) modifier combinations.


kdb424 (approved):
Much needed improvement to the docs. Thanks!

PR #410, opened 2022-04-22
Minimum necessary to add index and get capability and tests

morrijr:
Rework of #405.

This is the minimum that gives KC.keyname, KC['keyname'] and KC.get('keyname') with tests.

### DIFF
--- a/docs/keys.md
+++ b/docs/keys.md
@@ -23,27 +23,27 @@ The next few steps are the interesting part, but to understand them, we need to
 understand a bit about what a `Key` object is (found in `kmk/keys.py`). `Key`
 objects have a few core pieces of information:
 
-* Their `code`, which can be any integer. Integers below
+- Their `code`, which can be any integer. Integers below
   `FIRST_KMK_INTERNAL_KEY` are sent through to the HID stack (and thus the
   computer, which will translate that integer to something meaningful - for
   example, `code=4` becomes `a` on a US QWERTY/Dvorak keyboard).
 
-* Their attached modifiers (to implement things like shifted keys or `KC.HYPR`,
+- Their attached modifiers (to implement things like shifted keys or `KC.HYPR`,
   which are single key presses sending along more than one key in a single HID
   report. This is a distinct concept from Sequences, which are a KMK feature
   documented in `sequences.md`). For almost all purposes outside of KMK core,
   this field should be ignored - it can be safely populated through far more
   sane means than futzing with it by hand.
 
-* Some data on whether the key should actually be pressed or released - this is
+- Some data on whether the key should actually be pressed or released - this is
   mostly an implementation detail of how Sequences work, where, for example,
   `KC.RALT` may need to be held down for the entirety of a sequence, rather than
-  being released immediately before moving to the next character.  Usually end
+  being released immediately before moving to the next character. Usually end
   users shouldn't need to mess with this, but the fields are called `no_press`
   and `no_release` and are referenced in a few places in the codebase if you
   need examples.
 
-* Handlers for "press" (sometimes known as "keydown") and "release" (sometimes
+- Handlers for "press" (sometimes known as "keydown") and "release" (sometimes
   known as "keyup") events. KMK provides handlers for standard keyboard
   functions and some special override keys (like `KC.GESC`, which is an enhanced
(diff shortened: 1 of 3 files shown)


xs5871 (approved):
Code is clean and appears to get rid of potential infinite loops during key lookup, which is nice.
Great unit tests.
If I'm being picky: the only thing missing is an update to the docs about the new functionality.

morrijr:
Picky is fine :)

RE:docs: happy to add, but not 100% sure where. When I raise the PR for using strings in the keymaps (which at the moment is my aim) then for **sure** :)

xs5871:
I guess in [keys.md(https://github.com/KMKfw/kmk_firmware/blob/master/docs/keys.md).
Another bulletpoint for `KC` access methods and case insensitivity.

morrijr:
Doc changes OK?

docs/keys.md
```diff
@@ -166,3 +166,18 @@ have my handlers attached:
 ```python
 SHREKLESS_ALT = KC.LALT.clone()
 ```
+
+You can also refer to a key by index:
+* KC['A']
+* KC['NO']
+* KC['LALT']
```
xs5871, line 173:
```suggestion
* `KC['A']`
* `KC['NO']`
* `KC['LALT']`
```

docs/keys.md
```diff
@@ -166,3 +166,18 @@ have my handlers attached:
...
+
+or the get function which has an optional argument of `default`. `default` is
+be returned if the key is not found (`None` unless otherwise specified):
+* KC.get('A')
+* KC.get('NO', None)
+* KC.get('NOT DEFINED', KC.RALT)
```
xs5871, line 179:
```suggestion
Or the `KC.get` function which has an optional default argument, which will
be returned if the key is not found (`default=None` unless otherwise specified):
* `KC.get('A')`
* `KC.get('NO', None)`
* `KC.get('NOT DEFINED', KC.RALT)`
```

docs/keys.md
```diff
@@ -166,3 +166,18 @@ have my handlers attached:
...
+
+Keys names are case-sensitive.  KC['NO'] being different from KC['no']. It is recommended
+that names are capitalised typically. The exception to this is alphabetical; KC['A'] and
+KC['a'] will both return the same, unshifted, key.
```
xs5871, line 183:
```suggestion
Key names are case-sensitive. `KC['NO']` is different from `KC['no']`. It is recommended
that names are capitalised typically. The exception to this is alphabetical; `KC['A']` and
`KC['a']` will both return the same, unshifted, key.
```

xs5871 (changes requested):
Almost there.

morrijr:
lol. *Love* documentation... :) updating. 2mins.

docs/keys.md
morrijr: I think "that names are typically capitalised" reads better?

xs5871: also "exception to this is alpha keys: "

morrijr: Have to admit. I'd love to break backwards compatibility and make KC['A'] give back a shifted KC['a']. But that's just me :)

xs5871: You can always open an issue for the community to discuss that.
The problem I see with that is that capitalized letters are the default right now. We need some kind of consistency, especially since ASCII and keycodes are very different things.

morrijr: Oh, I don't care that much! ;)

xs5871 (approved):
I'll approve this, _but_: for the future, please try to keep PRs topical and especially don't add new stuff to change requests.

morrijr:
> I'll approve this, _but_: for the future, please try to keep PRs topical and especially don't add new stuff to change requests.

What new stuff to this PR? There were the code changes, unit tests and docs... the other PR was closed because it went too far IMO.

xs5871:
Formatting/style changes in the docs that weren't the original concern of the PR.
Those were all good, but they should have gone into a different PR. It's not about the content of the changes, but about organization.

morrijr:
That wasn't me (directly), that was VSCode on save. I didn't even realise it'd changed it. I don't even know which extension did it. On the other hand; https://www.markdownguide.org/cheat-sheet/ at least it's correct now.

PR #411, opened 2022-04-22
modholdandtap implementation

masiullah:
This module allows to immitate the behaviour of ATL+TAB or CMD+TAB, etc.  
Basically, it will hold the mod and tap a key on a layer other than default layer. The mod will be released when any other key is pressed or the layer key is released.
The key will do nothing when it is placed on the default layer

Not sure if something like this is already there in KMK


masiullah:
@kdb424 would you be able to make this a draft PR as I need to do some more testing

masiullah:
I completed the missing test on the kb2040. Its ready for review

xs5871:
I'd like to submit a review, but I honestly don't understand what this does or what it is for.
I read through the code, but I guess missing some context.
@masiullah would you be so kind and expand a little on your explanation, maybe with one or two examples even?

masiullah:
> I'd like to submit a review, but I honestly don't understand what this does or what it is for. I read through the code, but I guess I'm missing some context. @masiullah would you be so kind and expand a little on your explanation, maybe with one or two examples even?

This key will help to simulate the behavior of alt + tab that is used to switch between apps for smaller keyboards where there are not a lot of keys on the default layer. To make this work put this key in non default layer and then hold the layer key and press the mod hold tap key and it work similar to alt  + tab (like you keep holding the layer key and pressing the mhat key will switch to different windows). Hope this is helpful to get the context

xs5871:
What's the difference to `KC.LALT(KC.TAB)` on a second layer?

masiullah:
> What's the difference to `KC.LALT(KC.TAB)` on a second layer?

I think that does not work like a real atl tab as you cannot hold the alt key and switch between windows. this will allow to do that

xs5871:
Thanks, I think I get it now. Is it strictly necessary to restrict that to non-default layers?

masiullah:
yes, otherwise you will not be able to hold the layer key which will act like a modifier
I was thinking to have it on default layer also but the problem was releasing the modifier and could do using a timeout but thought if some one is going to put the key on the default layer then he must be having a lots of keys and most likely will not use it as he will have dedicated keys for the mods

xs5871:
I understand that it's more likely to have this on a secondary layer, but I don't see a reason to disallow it on the default layer.
The logic you use to remove the held modifier checks for any key event, not for a layer transition.

Basically, we're talking about a key with a "personal" sticky modifier, right?

masiullah:
I think the restriction can be removed and also please let me know if module name or key name should be changed

masiullah:
@xs5871  its ready for review

kmk/modules/modholdandtap.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class ModHoldAndTap(Module):
+    def __init__(self):
+        self._timeout_key = False
+        self._active = False
+        self._prev_key = None
+        make_argumented_key(
+            names=('MHAT',),
+            validator=ModHoldAndTapValidator,
```
xs5871, line 18:
```suggestion
            validator=ModHoldAndTapMeta,
```

xs5871, line 18:
```suggestion
            validator=ModHoldAndTapMeta,
```

kmk/modules/modholdandtap.py
```diff
@@ -0,0 +1,65 @@
...
+
+            if self._active and self._prev_key is not None and is_pressed:
+                # release previous key
+                self.release(keyboard, self._prev_key)
+            self._prev_key = key
+            if is_pressed:
+                keyboard.process_key(key.meta.mod, is_pressed)
+
+                self._active = True
+            keyboard.process_key(key.meta.kc, is_pressed)
```
xs5871, line 38: This should go into `on_pressed` and `on_released` key handlers.

kmk/modules/modholdandtap.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, kc, mod):
+        self.kc = kc
+        self.mod = mod
+
+
+class ModHoldAndTap(Module):
+    def __init__(self):
+        self._timeout_key = False
+        self._active = False
+        self._prev_key = None
```
xs5871, line 15: `_active_key` may be more fitting.

xs5871, line 15: `_active_key` may be more fitting.

kmk/modules/modholdandtap.py
xs5871, line 38:
```suggestion
            if self._active and self._prev_key is not None and is_pressed:
                # release previous key
                self.release(keyboard, self._prev_key)
            self._prev_key = key
            if is_pressed:
                keyboard.process_key(key.meta.mod, is_pressed)

                self._active = True
            keyboard.process_key(key.meta.kc, is_pressed)
```

This should go into `on_pressed` and `on_released` key handlers.

xs5871, line 38:
```suggestion
            if self._active and self._prev_key is not None and is_pressed:
                # release previous key
                self.release(keyboard, self._prev_key)
            self._prev_key = key
            if is_pressed:
                keyboard.process_key(key.meta.mod, is_pressed)

                self._active = True
            keyboard.process_key(key.meta.kc, is_pressed)
```

This should go into `on_pressed` and `on_released` key handlers.

xs5871 (changes requested):
I'd prefer a more intuitive name.
Both "hold" and "tap" already have a meaning, and it refers to typing actions by the user, not something the keyboard does for you. The best I could come up with is "StickyMod", though that may be confused with "OneShot". "StickierMod", "StuckMod", "PermaMod", "TenaciousMod", I'm just exhausting a thesaurus now...

Let's start with the code changes, find a better name, and then we can polish the docs. We'll probably need a couple of rounds of reviews, if that's ok with you.

Edit: sorry, don't know why github doubled every one of my change requests...

masiullah:
@xs5871  I will go through the review comments and implement the same and multiple rounds of reviews are fine as we want to make it polished and easily used

masiullah:
@xs5871  I will go through the review comments and implement the same and multiple rounds of reviews are fine as we want to make it polished and easily used

kmk/modules/modholdandtap.py
```diff
@@ -0,0 +1,65 @@
+from kmk.keys import make_argumented_key
+from kmk.modules import Module
+
+
+class ModHoldAndTapValidator:
```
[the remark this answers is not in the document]
masiullah: Changed the name to StickyModMeta

docs/sticky_mod.md
```diff
@@ -0,0 +1,22 @@
+# Sticky Mod
+This module allows to immitate the behaviour of ATL+TAB or CMD+TAB, etc. for switching between open windows.
+The mod will be on hold and the key will be tapped. The mod will be released when any other key is pressed or the layer key is released.
```
xs5871, line 3:
```suggestion
This module allows to hold a modifier while a key is being tapped repeatedly; the modifier will be released when any other key is pressed or released.
This is for example useful if you want to switch between open windows with ALT+TAB or CMD+TAB, using only a single key.
```
```suggestion
This module allows to immitate the behaviour of ATL+TAB or CMD+TAB, etc. for switching between open windows.
The mod will be on hold and the key will be tapped. The mod will be released when any other key is pressed or the layer key is released.
```

kmk/modules/sticky_mod.py
```diff
@@ -0,0 +1,68 @@
...
+    def during_bootup(self, keyboard):
+        return
+
+    def before_matrix_scan(self, keyboard):
+        return
+
+    def process_key(self, keyboard, key, is_pressed, int_coord):
+        if self._active:
+            # release previous key if any other key is pressed
+            if self._active_key is not None:
```
xs5871, line 31: Is `self._active` even necessary when `self._current_key is not None` has the same effect?

xs5871, line 31: Is `self._active` even necessary when `self._current_key is not None` has the same effect?

xs5871 (approved):
I' happy with that.

PR #418, opened 2022-04-24
markdown spelling and formatting

rrotter:
Fixed many typographical errors. Most of these should be noncontroversial, explanation follows where there was more of a judgement call:

* Both customise and customize were used in the same article. Using both was sloppy. Chose customize, which is the preferred US and Oxford usage
* initialization is also the preferred US and Oxford usage, though it wasn't used both ways
* Several proper nouns in the documentation used two conflicting capitalizations, in each case switched to the "official" spelling for consistency:
  * "Bluetooth" vs "bluetooth"
  * "Unicode" vs "unicode"
  * "USB" vs "usb"

More straightforward choices where one spelling is clearly wrong:
* [SparkFun, not Sparkfun](https://www.sparkfun.com/)
* [Pro Micro, not ProMicro](https://www.sparkfun.com/products/15795)
* [CircuitPython spelling](https://learn.adafruit.com/welcome-to-circuitpython/what-is-circuitpython)
* [Adafruit ItsyBitsy has no space](https://www.adafruit.com/product/3800)

I'm happy to remove any changes that you don't agree with.

### DIFF
--- a/docs/adns9800.md
+++ b/docs/adns9800.md
@@ -5,7 +5,7 @@ from kmk.modules.adns9800 import ADNS9800
 keyboard.modules.append(ADNS9800(cs=board.GP0, sclk=board.GP2, miso=board.GP4, mosi=board.GP3, invert_y=True))
 ```
 
-Firmware for this sensor has to be obtained separately and placed in kmk\modules\adns9800_firmware.py
+Firmware for this sensor has to be obtained separately and placed in `kmk\modules\adns9800_firmware.py`
 ```python
 firmware = (
     b'\x03'
@@ -15,4 +15,5 @@ firmware = (
 ```
 
 ## Constructor parameters
+
 ADNS9800(cs=*cs_pin*, sclk=*clock_pin*, miso=*miso_pin*, mosi=*mosi_pin*, invert_x=*False*, invert_y=*False*)
--- a/docs/capsword.md
+++ b/docs/capsword.md
@@ -1,6 +1,6 @@
 # CapsWord
-The capsword module functions similar to caps lock but will deactivate automatically when its encounters a key that breaks the word or after inactivity timeout.  
-By default it will not deactivate capsword on numbers, alphabets, underscore, modifiers, minus, backspace and other keys like modtap, layers, etc.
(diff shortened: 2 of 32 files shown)


docs/adns9800.md
```diff
@@ -15,4 +15,5 @@ firmware = (
 ```
 
 ## Constructor parameters
-ADNS9800(cs=*cs_pin*, sclk=*clock_pin*, miso=*miso_pin*, mosi=*mosi_pin*, invert_x=*False*, invert_y=*False*)
\ No newline at end of file
+
```
xs5871, line 18: Blank line after headings are generally inconsistent. Is there a preferred style?

docs/capsword.md
```diff
@@ -24,4 +24,4 @@ keyboard.keymap = [
 
 |Key                    |Aliases             |Description                                    |
 |-----------------------|--------------------|-----------------------------------------------|
-|`KC.CW`                |`KC.CAPSWORD`       |Enables/disables capsword                      |
\ No newline at end of file
+|`KC.CW`                |`KC.CAPSWORD`       |Enables/disables CapsWord                      |
```
xs5871, line 27: Some question for space after table separator. Is there a style recommendation?

xs5871 (approved):
This is wonderful, thank you so much.
I read through all of it and hope we can keep this level of attention to "detail" up.

kdb424:
We don't have a style guide for docs code. Currently it's just Black for code formatting basically, but would take input from anyone that wants to write a sane style guide for docs.

docs/adns9800.md
rrotter: In this case I was just making this match the ptBR version of the same file, since IMO the blank line looks better. Looks like at there is some consensus on this:
https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022

docs/capsword.md
rrotter: The linter I linked above _does not_ have a rule for this. Adding the space looks better IMO, but for this PR I kept the changes that didn't effect what ends up in the final rendered HTML to a minimum.

morrijr: My preference for markdown tables is to align things in the markdown, but then I read that far more often than going to the rendered content :)

It's also far more hassle unless the linter can automatically fix it.

PR #420, opened 2022-04-25
markdownlint

rrotter:
Done so far:
* stub out relevant docs in contributing.md
* add config file: .markdownlint.yml
  * disabled every rule we're violating that `markdownlint` doesn't know how to autocorrect w/ `--fix` flag
* disable linter at a few key points in the docs
* make the needed changes to ONE MARKDOWN FILE to make it pass the linter (appropriately chose docs/contributing.md)

To do:
* Wait for #418 to merge, rebase to master (I'll do this even if I never get to the rest of this); this is important because the rest of this WILL create changes that won't merge cleanly with that commit (or any major doc changes)
* Run `markdownlint --fix '**/*.md'`, review the hundreds of changes
* commit
* Re-enable rules in `.markdownlint.yml` one at a time, run `markdownlint '**/*.md'`, and either remediate the effected docs or dicide to keep the rule
* commit
* Possibly look at git options that prevent lint changes from breaking `git blame`. 99% sure this is possible.

Notes:
The main markdown linters I found are:
* [markdownlint/markdownlint](https://github.com/markdownlint/markdownlint): ruby
* [DavidAnson/markdownlint](https://github.com/DavidAnson/markdownlint): Node.js

I'm using the latter because it appears to have more active maintenance including, critically, on the ruleset. I'm using markdownlint-cli rather than markdownlint-cli2 because there seems to be better support of the former in existing github actions.

I don't have any plans to work on this further anytime soon, so if someone finds this work valuable please feel free to pick it up and run with it.

### DIFF
--- a/.markdownlint.yml
+++ b/.markdownlint.yml
@@ -1,0 +1,36 @@
+# starting point config to get linter passing. Almost all of this should be disabled, once existing files are remediated
+# see https://github.com/DavidAnson/markdownlint for full config
+
+# Default state for all rules
+default: true
+
+# MD001/heading-increment/header-increment Heading levels should only increment by one level at a time
+MD001: false
+
+# MD013/line-length Line length
+# Maybe consider only enforcing this on new docs?
+MD013: false
+
+# MD024/no-duplicate-heading/no-duplicate-header Multiple headings with the same content
+MD024: false
+
+# MD025/single-title/single-h1 Multiple top-level headings in the same document
+MD025: false
+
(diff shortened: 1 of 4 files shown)


kdb424:
We should consider adding a make arg for this as well, and possibly (in a different PR) make a make command that should be run before committing to ensure that tests all pass, auto format code, ect, so there is less things to run before committing.

rrotter:
Closing this. It's really stale.

PR #421, opened 2022-04-25
Peg rgb matrix

daysgobye:
## peg prefix
you will see that it's prefixed with Peg. Peg is some software I am working on to let you configure your kmk board. The prefix is to denote that this is the RGB extension that peg interacts with. Anyone can use it and make changes to it but there are two RGB extensions now (this one is very different) but still. Being that not only users will interact with this but also other code some of my design decisions were about that. I have however tried to make this easy to use for the end-user.

## Differences between `peg_rgb_matrix` and `Rgb` 
The primary change is that here we map each index in your led chain to its equivalent key, eg if `KC.Q` is the second key in your keymap array the second color is your `ledDisplay` will light up "Q". Beyond that mine is just less feature full lol, no animations as I have not had time for them yet and want to do them right.

## Differences between `peg_rgb_matrix` and qmk rgb matrix
Hopefully easier to set up for the board creator, and much easier for the end-user to adjust as they see fit. 

## Closing 
In closing, I have tried to make this my best pr yet and applied a lot of the things requested in my earlier prs already. I appreciate you taking the time to look it over and let me know if you want to see anything changed.

Not in this pr but after it's merged I have also done the `led_key_pos` for the CRKBD keyboard so I can merge that in too.

### DIFF
--- a/docs/peg_rgb_matrix.md
+++ b/docs/peg_rgb_matrix.md
@@ -1,5 +1,4 @@
 # Peg RGB Matrix
-In order to use this extension the user must make changes to both their kb.py and main.py files. Below you will find a more comprehensive list of changes required in order to use this extension. 
 
 ### What you can and cannot do with this extension:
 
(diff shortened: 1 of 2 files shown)


xs5871:
Before I give this a proper review, revisit our last conversation about docs:
- inline code quoting
- 1 sentence per line max

xs5871:
Before I give this a proper review, revisit our last conversation about docs:
- inline code quoting
- 1 sentence per line max

daysgobye:
> Before I give this a proper review, revisit our last conversation about docs:
> 
>     * inline code quoting
> 
>     * 1 sentence per line max

Sorry about the one sentence per line thing. That is fixed But I tried and think my inline code quoting is correct or I am grossly off  and can you please highlight an example of where I am doing it wrong.

kdb424:
It's been a month. I don't see any blocking changes, and we can improve as time goes on. Have tested on my board, and it's working, so we'll come back to fix it if @xs5871 decides there need to be changes. I'm aware that you are busy, but I've at least tested this for you.

PR #423, opened 2022-04-25
Allow strings in keymap. Replace with keys in keyboard extension bootup.

morrijr:
Allows keymaps to be defined like:

keyboard.keymap = [['A', 'B']]

Reducing the amount of typing true, but mainly intended to open up the possibility of loading the keymap from a json or yaml file without having to add functionality to the core.

PS, I don't speak the ptBR language. I hope google translate is good enough until somebody who does can fix any errors.

### DIFF
--- a/docs/config_and_keymap.md
+++ b/docs/config_and_keymap.md
@@ -2,9 +2,9 @@
 
 KMK is configured through a rather large plain-old-Python class called
 `KMKKeyboard`. Subclasses of this configuration exist which pre-fill defaults
-for various known keyboards (for example, many QMK, TMK, or ZMK keyboards 
-are supported with a nice!nano, or through our ItsyBitsy to Pro Micro pinout adapter. 
-This class is the main interface between end users and the inner workings of KMK. 
+for various known keyboards (for example, many QMK, TMK, or ZMK keyboards
+are supported with a nice!nano, or through our ItsyBitsy to Pro Micro pinout adapter.
+This class is the main interface between end users and the inner workings of KMK.
 Let's dive in!
 
 - Edit or create a file called `main.py` on your `CIRCUITPY` drive. You can also
@@ -71,6 +71,13 @@ print(dir(board))
 
 ```python
 keyboard.keymap = [[KC.A, KC.B]]
+```
+
+- You can also define the keymap with strings. They get replaced with `Key` objects
+  when the keyboard is started:
+
+```python
+keyboard.keymap = [['A', 'B']]
 ```
 
 You can further define a bunch of other stuff:
--- a/docs/config_and_keymap.md
+++ b/docs/config_and_keymap.md
@@ -2,9 +2,9 @@
 
 KMK is configured through a rather large plain-old-Python class called
 `KMKKeyboard`. Subclasses of this configuration exist which pre-fill defaults
-for various known keyboards (for example, many QMK, TMK, or ZMK keyboards
-are supported with a nice!nano, or through our ItsyBitsy to Pro Micro pinout adapter.
-This class is the main interface between end users and the inner workings of KMK.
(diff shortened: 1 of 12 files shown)


klardotsh (commented):
I'm generally a fan. Deferring to @xs5871 for final review. Thanks!

xs5871:
My main concern initially with adding this to the core were inconsistency (ASCII != scancodes), limited functionality (no argumented keys, etc.), and the potential confusion it'll lead to (why does 'A' work, but 'Æ' / 'foobar' / etc.  doesn't?) -- I could go on; but that's not a very fun or productive discussion.

Let's put it in an extension instead and I'll give it a thumbs-up without hesitation.

morrijr:
An extension is fine, but if a key has been named 'Æ' or 'foobar' they'll map just fine. These aren't the letter that's just coincidence, it's the name. Anyway. Extension.

I'll look into it and create a new PR.

xs5871:
We can keep the PR open as draft and you can push or rebase anytime you're ready (if you don't mind).
That would have the benefit of keeping the entire discussion and progress in the same place.

morrijr:
Changing this to an extension means if any string cannot be replaced with an appropriate key the keyboard will continue to try and work as any exception raised `during_bootup` is caught and thrown away (as is the case with modules).

I initially expected that any exception of any module or extensions would stop the keyboard. As written this would simply crash the keyboard later when the key pressed was found not to be a key.

I'm going to replace the unknown key with a KC.NO, but I my personal preference would be for the keyboard to not function, preferably as soon as possible (ie; at boot).

morrijr:
I'm not happy with the state of this PR. I dislike all the changes that have piled on. I would have preferred to close and create a new PR. Suggestions on how to tidy it up welcome.

kdb424:
Extensions are meant as critical components of the keyboard, and modules are meant to not kill the keyboard if they fail. Missing keys should be a failed boot. RGB running out of memory and crashing should not kill the keyboard. Glad the system was working as intended. I haven't actually had this tested by others until now 👍

morrijr:
https://github.com/KMKfw/kmk_firmware/blob/master/kmk/kmk_keyboard.py#L340-L345

It appears to be catching everything then potentially logging, but doesn't raise the exception again. This won't fail the boot?

kdb424:
Miss remembered. I guess I didn't force the fail, but more to contain modules into a sandbox that's not critical at least. I would agree that we should probably intentionally fail out of things like this. Better in a separate PR with reference to this issue.

morrijr:
Happy to. Question though; fail out for _all_ the functions of modules and _all_ the functions of extensions or just some?

xs5871:
> I'm going to replace the unknown key with a KC.NO, but I my personal preference would be for the keyboard to not function, preferably as soon as possible (ie; at boot).

I actually prefer the `KC.NO` accompanied by a debug message during initialization.

morrijr:
There is a debug message. Has to be enabled though the same as the one on the main keyboard class.

xs5871:
I'm ok with that. A less clunky name would be the cherry on top, maybe "stringy_keymaps" or "keymap_from_string", anything that makes it clear for users who don't already know what it does. That's up to you.

xs5871:
I'm ok with that. A less clunky name would be the cherry on top, maybe "stringy_keymaps" or "keymap_from_string", anything that makes it clear for users who don't already know what it does. That's up to you.

morrijr:
Renamed extension to stringy_keymaps

xs5871:
One last thing on naming convention: modules are snake case and `stringy_keymaps.py` is fine, classes are camel case -> `StringyKeymaps`.

xs5871 (approved):

PR #425, opened 2022-04-27
add docs/doc_tools for spell checking markdown

rrotter:
This requires aspell be installed and on your path. Right now it's an
interactive tool for cleaning up the docs, but it could be adapted to be a
commit hook or GitHub action.


kdb424:
Well, I was just about to bring up words that don't exist, but you clearly know the tool. I like this.

rrotter:
This is the tooling I used for #418, and the wordlist is seeded w/ all the words (and a lot of non-words) that I needed to accept in the process of that work.

kdb424:
Looks good to me. I made a comment about a make arg, and that would apply to any tool like these.

rrotter:
Added a make command, cleaned up the bash script, and edited a few more markdown files so `make spellcheck` can pass without adding anything too questionable to the spelling dictionary.

Any suggestions on whether there'd be a better place to put the script and dictionary?  The `/docs/docs_utils` path was sort of a placeholder, and I'm not really happy with it. Would these files make sense in `/util`?

kdb424:
Util is historically where we put any tooling that we didn't know where else to put it. That seems reasonable.

rrotter:
I'm ready to call this done. Could maybe add a note about this in docs/contributing.md, but I'm not sure what to say there. 

Made a few more edits to the docs along the lines of #418 because if I'm going to commit a new test, I should make the test pass. That being said, I _did not_ add `make spellcheck` to `make test` since there is an external non-python dependency.

rrotter:
I was just thinking about this PR a little more: This tooling was quite helpful to me well proofreading the docs. I submitted this PR since there were some interest in it. TBH, i'm not sure it belongs as part of the project. It would be nice if there was a tool like Black that was python native and available from `pip`. As it is, this feels like scope creep. Maybe it's best to close this or move it back to a draft?

kdb424:
I like it as even an optional tool. It'll be there for those that want to use it, but it won't be a hard requirement like black. Thoughts?

rrotter:
Yeah, that's fine too. If a better tool emerges there's no problem with changing.

util/spell.sh
```diff
@@ -0,0 +1,48 @@
...
+
+if ! which -s aspell; then
+  echo 'aspell command not found, cannot run spell check'
+  exit
+fi
+
+ROOT=`git rev-parse --show-toplevel`
+
+# This gets us only english .md files at the moment, but is clearly brittle
+MARKDOWN_FILES=`find $ROOT -name '*.md' -maxdepth 2 -not -name '*-ptBR.md'`
```
klardotsh, line 11: I believe we also have Japanese docs somewhere in the tree; this finder line needs to at least filter those out.

util/spell.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
klardotsh, line 1: Can you run https://www.shellcheck.net/ over this (locally or with the web version) to see if it passes with `#!/bin/sh`? If so, change the shebang so it works on any POSIX shell. If not, up to you whether we mandate bash (in which case change to `#!/usr/bin/env bash` since some OSes, notably NixOS, don't provide bash at the specified path as written) or fix  whatever shellcheck complains about to become POSIX compliant.

Due to how you use `$ASPELL` you'll almost certainly need to tell shellcheck to ignore some quoting heuristics even in Bash mode; go ahead and address those if you don't mind (or I can come in when I have a minute and do so).

util/spell.sh
```diff
@@ -0,0 +1,48 @@
...
+
+if ! which -s aspell; then
```
klardotsh, line 3: preference to: `hash aspell 2>/dev/null || { echo ' ... '; exit 1 }`

notably, we need to provide an exit code on L5; by default `exit` will return 0 which implies success; that's not the case here. I believe 127 is common for "file not found" returns, but frankly I'm cool with anything other than 0.

klardotsh (commented):
I dig this. Whenever we can get these few line items addressed I'd love to merge. Thanks for this!

kdb424:
As @klardotsh seems to have opinions, if possible, could it be adopted so this can get handled properly?

klardotsh:
Indeed, thank you!

PR #430, opened 2022-04-30
Extract key generators

morrijr:
This PR pulls the key test/create logic out of the `KeyAttrDict::__getitem__` into a collection that is iterated over.

This opens up the possibility of having extensions/modules/user code altering it allowing the list to be broken up resulting in potentially a smaller list.

**Reviewers**: I am not by trade a Python developer. This refactor works (both tests and on my hardware). _**However**_ while some of the previous code used a similar technique (and lambdas) I'm not 100% sure of any potential performance (memory or speed) issues. The testing I've done has shown no speed decrease, but I'm unable to judge potential impact to memory.

### DIFF
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -21,40 +21,307 @@ ALL_NUMBERS = '1234567890'
 ALL_NUMBER_ALIASES = tuple(f'N{x}' for x in ALL_NUMBERS)
 
 
-# this is a bit of an FP style thing - combining a pipe operator a-la F# with
-# a bootleg Maybe monad to clean up these make_key sequences
-def left_pipe_until_some(candidate, functor, *args_iter):
-    for args in args_iter:
-        result = functor(candidate, *args)
-        if result is not None:
-            return result
-
-
-def first_truthy(candidate, *funcs):
-    for func in funcs:
-        result = func(candidate)
-        if result is not None:
-            return result
-
-
-def maybe_make_mod_key(candidate, code, names):
+def maybe_make_mod_key(candidate, code, names, *args, **kwargs):
     if candidate in names:
-        return make_mod_key(code=code, names=names)
+        return make_mod_key(code=code, names=names, *args, **kwargs)
 
 
-def maybe_make_key(candidate, code, names):
+def maybe_make_key(candidate, code, names, *args, **kwargs):
     if candidate in names:
-        return make_key(code=code, names=names)
+        return make_key(code=code, names=names, *args, **kwargs)
 
 
-def maybe_make_shifted_key(candidate, code, names):
+def maybe_make_shifted_key(candidate, code, names, *args, **kwargs):
     if candidate in names:
-        return make_shifted_key(code=code, names=names)
+        return make_shifted_key(code=code, names=names, *args, **kwargs)
 
 
-def maybe_make_consumer_key(candidate, code, names):
+def maybe_make_consumer_key(candidate, code, names, *args, **kwargs):
     if candidate in names:
-        return make_consumer_key(code=code, names=names)
+        return make_consumer_key(code=code, names=names, *args, **kwargs)
+
+
+def maybe_make_argumented_key(
+    candidate,
+    validator=lambda *validator_args, **validator_kwargs: object(),
+    names=tuple(),  # NOQA
+    *constructor_args,
+    **constructor_kwargs,
+):
+    if candidate in names:
+        return make_argumented_key(
+            validator, names, *constructor_args, **constructor_kwargs
+        )
+
+
+def maybe_make_alpha_key(candidate):
+    if len(candidate) == 1:
+        candidate_upper = candidate.upper()
+        if candidate_upper in ALL_ALPHAS:
+            return make_key(
+                code=4 + ALL_ALPHAS.index(candidate_upper),
+                names=(
+                    candidate_upper,
+                    candidate.lower(),
+                ),
+            )
+
+
+def maybe_make_numeric_key(candidate):
+    if candidate in ALL_NUMBERS or candidate in ALL_NUMBER_ALIASES:
+        try:
+            offset = ALL_NUMBERS.index(candidate)
+        except ValueError:
+            offset = ALL_NUMBER_ALIASES.index(candidate)
+
+        return make_key(
+            code=30 + offset, names=(ALL_NUMBERS[offset], ALL_NUMBER_ALIASES[offset])
+        )
+
+
+__unicode_noop = UnicodeModeKeyMeta(UnicodeMode.NOOP)
+__unicode_ibus = UnicodeModeKeyMeta(UnicodeMode.IBUS)
+__unicode_ralt = UnicodeModeKeyMeta(UnicodeMode.RALT)
+__unicode_winc = UnicodeModeKeyMeta(UnicodeMode.WINC)
+
+KEY_GENERATION_FUNCTIONS = (
+    # NO and TRNS are functionally identical in how they (don't) mutate
+    # the state, but are tracked semantically separately, so create
+    # two keys with the exact same functionality
+    (
+        maybe_make_key,
+        (None, ('NO', 'XXXXXXX')),
+        {'on_press': handlers.passthrough, 'on_release': handlers.passthrough},
+    ),
+    (
+        maybe_make_key,
+        (None, ('TRANSPARENT', 'TRNS')),
+        {'on_press': handlers.passthrough, 'on_release': handlers.passthrough},
+    ),
+    (maybe_make_alpha_key, (), {}),
+    (maybe_make_numeric_key, (), {}),
+    (maybe_make_key, (None, ('RESET',)), {'on_press': handlers.reset}),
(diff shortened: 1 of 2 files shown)


kmk/keys.py
```diff
@@ -21,40 +21,307 @@
...
+
+KEY_GENERATION_FUNCTIONS = (
+    # NO and TRNS are functionally identical in how they (don't) mutate
+    # the state, but are tracked semantically separately, so create
+    # two keys with the exact same functionality
+    (
+        maybe_make_key,
+        (None, ('NO', 'XXXXXXX')),
+        {'on_press': handlers.passthrough, 'on_release': handlers.passthrough},
+    ),
```
xs5871, line 95: This is definitely more memory intensive than the lambdas. You're creating a dictionary with string indices for every key.

xs5871, line 95: This is definitely more memory intensive than the lambdas. You're creating a dictionary with string indices for every key.

morrijr:
Is there a better way to do this? I didn't really want the dictionaries, but lack the python to express it better. It's probably my functional preference coming out, but I find this a much nicer thing than all the if/elsif's.

I think it would enable the manipulation of the list; split it across files that get included, reduce the possibles to reduce memory? And potentially to create the documentation from?

If it's a none starter, that's fine.

Working OK on my RP2040 though.

Is there any way on the micros to get memory usage rather than just guess work?

xs5871:
The intention is great and I support it.
The implementation is another thing. Usually, lambdas go on the stack and data structures go on the heap. I'm not sure how circuitpython handles that in detail, considering there may be some JIT optimisation going on. At the very best, we'd move some memory consumption from stack to heap, but I don't think we could save any, because globals on the heap never go out of scope thus never get garbage collected.

You'll probably not get around lambdas or closures, the latter ones me be easier on the eyes. A global list of key generating function references may be a better solution:
```python
def maybe_make_key(candidate, code, names, *args, **kwargs):
    def closure(key):
         if candidate in names:
             return make_key(code=code, names=names, *args, **kwargs)
    return closure

KEY_GENERATION_FUNCTIONS = (
    maybe_make_key(None, ('NO', 'XXXXXXX'), on_press=handler.passthrough, on_release=handlers.passthrough)
)
```
#NOQA

xs5871:
The intention is great and I support it.
The implementation is another thing. Usually, lambdas go on the stack and data structures go on the heap. I'm not sure how circuitpython handles that in detail, considering there may be some JIT optimisation going on. At the very best, we'd move some memory consumption from stack to heap, but I don't think we could save any, because globals on the heap never go out of scope thus never get garbage collected.

You'll probably not get around lambdas or closures, the latter ones me be easier on the eyes. A global list of key generating function references may be a better solution:
```python
def maybe_make_key(candidate, code, names, *args, **kwargs):
    def closure(key):
         if candidate in names:
             return make_key(code=code, names=names, *args, **kwargs)
    return closure

KEY_GENERATION_FUNCTIONS = (
    maybe_make_key(None, ('NO', 'XXXXXXX'), on_press=handler.passthrough, on_release=handlers.passthrough)
)
```
#NOQA

xs5871:
> Is there any way on the micros to get memory usage rather than just guess work?

`ustack` and `uheap`

xs5871:
> Is there any way on the micros to get memory usage rather than just guess work?

`ustack` and `uheap`

morrijr:
Closures was exactly the intent, just didn't know how to do in python :)

Only question though. This has now changed all the maybe_make_ functions to closures which changes the calling characteristics. I can't find any usages outside of keys.py and the tests, but is that OK?

Further to this change, do you see this list being split into different (files?) and recombined somehow? Do you see that as useful?

xs5871:
> Only question though. This has now changed all the maybe_make_ functions to closures which changes the calling characteristics. I can't find any usages outside of keys.py and the tests, but is that OK?

That's fine.

> Further to this change, do you see this list being split into different (files?) and recombined somehow? Do you see that as useful? Smaller list quicker startup? Pick and choose languages?

It'll make it easier to go that direction, for sure. But before that, we need to verify that this refactor isn't going to cause us issues down the line. I'm going to have to do a few tests.

xs5871:
I did some tests and measurements. TL;DR: my suggestion of closures was a terrible idea, but I found a "better" solution.

The analysis was done on my personal config on an RP2040, using custom debugging during key object initialization.
Memory info is from  `ustack` and `gc`, plus timestamps.
Disclaimer: Any numbers mentioned aren't exact and are intended to give an impression of the order of magnitude.
Allocation is fairly consistent across runs up until the KMK main loop starts, after that only max values were considered.
Due to dynamic memory management, once KMK "runs" the momentary readings are all over the place.
I tried to concentrate on metrics that are statistically representative of every usecase, (and explicitly don't claim to have achieved that).

The PR actually saves some stack allocation during key generation (~500B) and is a tiny bit slower (~3%).
However: It has a significant impact on the heap.
I saw in increase of almost 20kB in heap allocation, which persisted after initialization.
That's a considerable amount given that we have to work with a couple of hundred kB at most (~7% increase on my MCU).

After some research and brain thinking, my understandig is, that unless lambdas or closures go out of scope (which doesn't happen for global ones), their data/bytecode/whatever stays on the heap.
Sounds obvious... and in this specific case not particularly great.

However: if we take this approach one step further (as loosely suggested in the discussion), and have a generator function per "logical" key group, like:
```python
def maybe_make_fn_key(candidate):
    codes = (
        (58, ('F1',)),
        (59, ('F2',)),
        (60, ('F3',)),
        # etc.
        )

    for code, names in codes:
        if candidate in names:
            return make_key(code=code, names=names)


KEY_GENERATORS = (
# ...
maybe_make_fn_key,
# ...
)
```
we'd save about ~150B on the stack and surprisingly ~1k on the heap, compared to the current implementation.
The reduction in heap usage is still a mystery to me.
The major drawback is execution time; it's about 16% slower than the current code -- however: turns out that ~40% of the time for key generation is spent on the `gc.collect()` call after each key instantiation.
If we were to remove that (doesn't have an impact with my setup anyway), we'd still gain about 22% of time during init.

morrijr:
Possibly a stupid question/thought, but if we split the generate up into many classes/functions/whatever and 'just' created the keys/groups of keys before asking for any in KC... would that be better?

I realise it's a breaking change, but we could make it a one line ("addAllKeys()"?) to get back to all and have "addAlpha" "addNumeric" "addKeypad" etc?

That way it removes all the if/candidate checks.

xs5871:
It's a valid question.
We explicitly avoid that because of the excessive memory footprint.
The reason for the complicated late instantiation is to only allocate key objects that will actually be used.

morrijr:
I get that and I've certainly done similar (late initialisation etc) for other things.

What I'm wondering is; is the over head worth the complexity. Bear with me a minute. I'm thinking that it's probably extremely rare that a keyboard will have a '1' but not the other numbers. Similar, if it has an 'a' it will have 'b'-'z'.

If each block of things was broken out into its own class or function and only what used by the keyboard added to KC that would be the idea right?

Now. This means more effort on the part of the person preparing the keyboard. However, it also gives more control. If somebody only wants the vowels (as a silly example) it would be possible for them to create their own class/function and explicitly add those keys and only those keys to the KC. This could be extended to alternative languages and non-qwerty.

Can you think of any alternatives?

kdb424:
I'm all in for someone looking into that again. We used to just make all of the US keys *way back when*, but the core has been getting more and more... Let's just say "not user friendly" as time goes on. Python is great for new users, but it's also insanely memory hungry, and microcontrollers are only so powerful. At this point, we need to try to get the core as memory light as possible, even if that means not being as nice to work on. I'm just idly watching this as I don't have any better input, or we wouldn't be in this situation to begin with, but letting people know "how we got here" basically.

xs5871:
I agree with @kdb424.
A set of default/static keys and a set of dynamic keys could speed up initialization and potentially further reduce the memory footprint. I'd suggest doing that in a seperate PR.
That being said, we have a bit more freedom with refactoring the core, as long as each iteration is stable. Ever user-facing change needs more care; it's obviously harder to change public APIs once established.

xs5871:
Well, I did a quick test with pre-initializing alphanums, and it makes no discernable difference in speed or memory consumption. It was maybe .2 seconds faster over all for me. Still, could be a good addition if an OS doesn't use the us layout by default.

morrijr:
I'm sorry folks, I've no idea what you're asking/thinking about.

xs5871:
I'd suggest doing this bit by bit.
Refactoring the long block of code into a list of functions is the first step. You already put some good work into that. Unfortunately the lambda approach doesn't quite give the performance we need, and we have to do it like mentioned in https://github.com/KMKfw/kmk_firmware/pull/430#issuecomment-1120304936.
If you want to do that, great, if not I can clean up my thrown together test code (which is based on your work) and send it to you as PR, or submit it myself. Whichever way works for you.

In another issue or PR we can then talk about spliting the key generators further into a default set and on-demand set. That'll need some thought, because those may be configurable by the user at some point.

xs5871:
merged in #527

PR #432, opened 2022-05-01
Align debug messages with method called

morrijr:
Some of the debug messages in kmk_keyboard weren't reflecting what was happening.


kdb424 (approved):
Thanks!

PR #433, opened 2022-05-02
Fix crash when typing quickly on a split board

### LINKED ISSUE #424: Issue when typing random keys very fast
I am getting the below error in the serial when I type keys very fast. I am using KB2040

Auto-reload is on. Simply save files over USB to run them or enter REPL to disable.
main.py output:
  File "main.py", line 33, in <module>
  File "kmk/kmk_keyboard.py", line 424, in go
  File "kmk/kmk_keyboard.py", line 477, in _main_loop
  File "kmk/kmk_keyboard.py", line 115, in _handle_matrix_report
  File "kmk/kmk_keyboard.py", line 165, in _on_matrix_changed
  File "kmk/kmk_keyboard.py", line 181, in pre_process_key
KeyError: 6

wildesteven:
This fixes #424 by creating a new `KeyEvent` instance for each `matrix_update` like we already do for `secondary_matrix_update`.  Previously, we were always adding the same `KeyEvent` instance to the `matrix_update_queue` which works fine until you try to add a second one to the queue, changing the first one.   Now you can add as many `KeyEvent`s to the queue as you want without them interfering with each other.

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -1,5 +1,7 @@
 from supervisor import ticks_ms
 
+from keypad import Event as KeyEvent
+
 from kmk.consts import KMK_RELEASE, UnicodeMode
 from kmk.hid import BLEHID, USBHID, AbstractHID, HIDModes
 from kmk.keys import KC
@@ -458,7 +460,9 @@ class KMKKeyboard:
         for matrix in self.matrix:
             update = matrix.scan_for_changes()
             if update:
-                self.matrix_update = update
+                self.matrix_update = KeyEvent(
+                    key_number=update.key_number, pressed=update.pressed
+                )
                 break
         self.sandbox.secondary_matrix_update = self.secondary_matrix_update
 


kmk/kmk_keyboard.py
```diff
@@ -458,7 +460,9 @@ def _main_loop(self):
         for matrix in self.matrix:
             update = matrix.scan_for_changes()
             if update:
-                self.matrix_update = update
+                self.matrix_update = KeyEvent(
+                    key_number=update.key_number, pressed=update.pressed
+                )
```
xs5871, line 465: If this is caused by re-using a single KeyEvent object, it should be fixed in `scanners.keypad`.

wildesteven: Thanks, that makes sense.  I moved the changes to keypad.py

xs5871: In that case you can remove `KeypadScanner.current_event` completely and use `keypad.events.get` instead of `get_into`.

xs5871: In that case you can remove `KeypadScanner.current_event` completely and use `keypad.events.get` instead of `get_into`.

wildesteven: I removed `KeypadScanner.curr_event` and made the function work without it.  I also removed `KeypadScanner.__init__()` as it no longer contains any code.

xs5871 (approved):
Excellent, thanks.

PR #434, opened 2022-05-02
Add Support for RP2.65-F

BigTuna94:
The RP2.65-F is my custom 65% mechanical keyboard based on the RP2040 MCU.
The '-F' signifies "Faders", which are analog linear potentiometers build onto the board.

Open source HW repo: https://github.com/BigTuna94/RP2.65-F

Since KMK did not have support for analog potentiometers, I added a module modeled after the encoder module.


kdb424:
I'm generally a fan of this. I'll allow for some time for other reviewers to comment, but once tests are fixed, I won't block merge.

kdb424 (approved):

BigTuna94:
@xs5871 I'm always happy to receive feedback! I actually agree about the separate base/interface/handler. I'd be happy to distill the potentiometer module down to fewer classes (and remove any mention of GPIO).

As far as the consumer control, I hadn't realized that there was a method to send control codes directly... seems that `keyboard._hid_helper.hid_send()` is the method that I need..

BigTuna94:
@xs5871 I've done some revisions based on your feedback. I think it still makes sense to have a separate PotentiometerHandler class, since it makes instantiating multiple potentiometers a little easier from the user keymap perspective.

Thank you for advising me to use the kmk HID CC from the core - this setup is actually noticeably faster than the adafruit library!

kmk/modules/potentiometer.py
```diff
@@ -0,0 +1,88 @@
...
+        self.cb = move_callback
+
+        # callback function on events.
+        self.on_move_do = lambda state: self.cb(state)
+
+    def get_state(self):
+        return {
+            'direction': self.is_inverted and -self._direction or self._direction,
+            'position': self.is_inverted and -self._pos or self._pos,
+        }
```
xs5871, line 23: In my opinion the return type should be a proper class `PotentiometerState` and not a dictionary.
The encoder module also uses a dict, but I think a class is more pythonic.

xs5871, line 23: In my opinion the return type should be a proper class `PotentiometerState` and not a dictionary.
The encoder module also uses a dict, but I think a class is more pythonic.

xs5871:
Nothing left to reasonably complain about. Ready to merge from my point of view.

PR #439, opened 2022-05-10
Dynamic Sequences module

wildesteven:
This PR adds dynamic macro functionality.  It lets you record a temporary macro just by typing it without modifying your keymap. Any macro recorded this way is temporary and will remain available until your keyboard reboots.


kdb424:
I've got only 2 comments before anyone tries to review this. 1: This is great, and some swear by this for passwords or whatnot as it's not stored in eeprom. 2: "Macro" has been reserved for the programatic term, and we leave them known as "Sequences" for sanity reasons. If this is an issue, I can ping the guy that made that call in the first place and let you two battle that one out haha.

Tests also don't pass. `make fix-isort fix-formatting test` should auto fix most issues for you, and run a test to let you know what is still broken, if anything.

wildesteven:
Thanks.  I can take care of that.  Would it be best if I rename it to Dynamic Sequences?

kdb424:
That sounds reasonable. If someone wants to debate the semantics, it'll be a github issue, but for now, yeah, that's great.

kdb424 (approved):
I'll let other reviewers comment, but I won't block the merge.

klardotsh (commented):
Not leaving a binding review, just saying that at a high level: this is seriously badass, thanks! Feels like a really good use of Python and a differentiating factor from many alternative firmwares. Good stuff.

xs5871:
Sounds great, I'll give it an in-depth review.

kmk/modules/dynamic_sequences.py
```diff
@@ -0,0 +1,253 @@
...
+
+# Keycodes for number keys
+_numbers = range(KC.N1.code, KC.N0.code + 1)
+
+
+class Sequence:
+    def __init__(self):
+        self.repetitions = 1
+        self.interval = 0
+        self.sequence_data = [(set(), 0), (set(), 0), (set(), 0)]
```
xs5871, line 30: opaque data structure, suggestion: `collections.namedtuple` with `keys` and `timestamp` entries.

kmk/modules/dynamic_sequences.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def play_frame(self, keyboard):
+        # Send the keypresses at this point in the sequence
+        if not check_deadline(
+            ticks_ms(), self.start_time, self.current_slot.sequence_data[self.index][1]
+        ):
+            if self.index:
+                for key in self.current_slot.sequence_data[self.index - 1][
+                    0
+                ].difference(self.current_slot.sequence_data[self.index][0]):
```
xs5871, line 166: There are a couple of hard to read multi line statements. If it's that complicated, give it a name:
```suggestion
                prev = self.current_slot.sequence_data[self.index - 1][0]
                cur = self.current_slot.sequence_data[self.index][0]
                for key in prev.difference(cur):
```

kmk/modules/dynamic_sequences.py
```diff
@@ -0,0 +1,253 @@
...
+            (set(), self.current_slot.sequence_data[-1][1] + 20)
+        )
+
+        # Wait for the specified interval
+        self.current_slot.sequence_data.append(
+            (
+                set(),
+                self.current_slot.sequence_data[-1][1]
+                + self.current_slot.interval * 1000,
+            )
```
xs5871, line 153: hard to read multi line statement. Well, this is very much on the edge.

xs5871 (commented):
A unit test would be great at some point, could be in another PR.

morrijr:
This was on my todo, though far better done than I probably would have. Thank you for the work.

Would it be possible for the macro to be saved to/loaded from 'disk'? Optionally?

The only other thing I'd add, should there be a link in the modules.md to the docs?

kdb424:
> Would it be possible for the macro to be saved to/loaded from 'disk'? Optionally?
Circuitpython *will* allow disk access, sort of, but you *must* disable user/os access when that happens, and it's likely to cause more problems than it would help IMHO.
 
> The only other thing I'd add, should there be a link in the modules.md to the docs?
This would be great. Thank you.

xs5871:
> > Would it be possible for the macro to be saved to/loaded from 'disk'? Optionally?
> > Circuitpython _will_ allow disk access, sort of, but you _must_ disable user/os access when that happens, and it's likely to cause more problems than it would help IMHO.

`microcontroller.nvm` is probably what we'd want to use. We wanted to use that to store states, like RGB configuration for example. It's unstructured, however, and it would make sense to build some abstraction layer for it, or define a serialization format.

wildesteven:
Thanks for the review.  I updated the PR.

PR #441, opened 2022-05-12
The variable pixels is initiated in wrong way

pickme467:
Currently pixels shall be a list, but when pixels is not initiated then it is initiated as neopixel.NeoPixel object instead of list with neopixel.NeoPixel obeject element inside.

### DIFF
--- a/kmk/extensions/rgb.py
+++ b/kmk/extensions/rgb.py
@@ -110,12 +110,12 @@ class RGB(Extension):
         if pixels is None:
             import neopixel
 
-            pixels = neopixel.NeoPixel(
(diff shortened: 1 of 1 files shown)


xs5871:
This should be handled by line 123f.
`pixel` may be initialized as a `PixelBuf` or list thereof.

pickme467:
> This should be handled by line 123f.
> `pixel` may be initialized as a `PixelBuf` or list thereof.

You are right. I used a version of NeoPixel that didn't inherit from PixelBuf, so it seems like a local problem for me. I guess it can be rejected then.

PR #443, opened 2022-05-13
Update peg_oled_display.py

ldsands:
This makes is so that the SCL and SDA are taken from `kb.py` if they are explicitly declared. I don't know if this is a good way to fix it or not but this pr does fix this issue.

### DIFF
--- a/kmk/extensions/peg_oled_display.py
+++ b/kmk/extensions/peg_oled_display.py
@@ -1,3 +1,4 @@
+import board
 import busio
 import gc
 
@@ -123,9 +124,9 @@ class Oled(Extension):
     def on_runtime_disable(self, sandbox):
         return
 
-    def during_bootup(self, board):
+    def during_bootup(self, board, SCL=board.SCL, SDA=board.SDA):
         displayio.release_displays()
-        i2c = busio.I2C(board.SCL, board.SDA)
+        i2c = busio.I2C(SCL, SDA)
         self._display = adafruit_displayio_ssd1306.SSD1306(
             displayio.I2CDisplay(i2c, device_address=0x3C),
             width=self._width,


ldsands:
Nevermind that doesn't fix it. I'll probably get to this later today and come up with some alternative solution.

daysgobye:
Are you trying to make it so you can pass in pins or pull them from kb.py? 

If so then the way I would go about that (could be the wrong way) is in the __init__ pass them there and then if they are both passed set self.SCL & self.SDA and make a new bool self.use_kb=false. Then in during boot up you can check if self.use_kb==false and  do i2c = busio.I2C(self.SCL, self.SDA).

This way you can continue to use it from the kb.py the way I like and you can pass in pins :) just my $0.02

ldsands:
> Are you trying to make it so you can pass in pins or pull them from kb.py?
> 
> If so then the way I would go about that (could be the wrong way) is in the **init** pass them there and then if they are both passed set self.SCL & self.SDA and make a new bool self.use_kb=false. Then in during boot up you can check if self.use_kb==false and do i2c = busio.I2C(self.SCL, self.SDA).
> 
> This way you can continue to use it from the kb.py the way I like and you can pass in pins :) just my $0.02

I was trying to find any way to pass it in though I think I would personally prefer to pass it in from main.py rather than kb.py. Aso for how to implement this I don't know. I'm mostly self-taught in python and I do data science so classes are something that I've never taken the time to learn. I can usually muddle my way thought but I'm often confused. Feel free to implement what you just outlines.

On a somewhat unrelated note, I'm having a ton of difficulties in consistently getting the oleds to work. I imagine it has something to do with the fact that I'm trying to use this on a split and/or on RP2040 (from sparkfun). The inconsistency is leading me to really have difficulty with this.

daysgobye:
> 
> I was trying to find any way to pass it in though I think I would personally prefer to pass it in from main.py rather than kb.py. Aso for how to implement this I don't know. I'm mostly self-taught in python and I do data science so classes are something that I've never taken the time to learn. I can usually muddle my way thought but I'm often confused. Feel free to implement what you just outlines.
> 
> On a somewhat unrelated note, I'm having a ton of difficulties in consistently getting the oleds to work. I imagine it has something to do with the fact that I'm trying to use this on a split and/or on RP2040 (from sparkfun). The inconsistency is leading me to really have difficulty with this.

do you have your code pushed? I have tested this with a split on rp2040 but not the spark fun one. If you want to send me your code on the kmk server you can find me as "cole" Im happy to look at it. But I think I know the problem.

ldsands:
I'm not sure what I did to mess up the OLED code, but it is all working fine now. Thanks for your help.

PR #444, opened 2022-05-17
Add layers to pimoroni trackball

fivetide:
add alternating behavior to pimoroni - encoder-like keyboard input. Also adds layer functionality to the trackball (looks at the keyboards layers)

### DIFF
--- a/kmk/modules/pimoroni_trackball.py
+++ b/kmk/modules/pimoroni_trackball.py
@@ -1,13 +1,13 @@
-"""
+'''
 Extension handles usage of Trackball Breakout by Pimoroni
 Product page: https://shop.pimoroni.com/products/trackball-breakout
-"""
+'''
+from micropython import const
+
 import math
 import struct
 
-from micropython import const
-
-from kmk.keys import make_key, make_argumented_key
+from kmk.keys import make_argumented_key, make_key
 from kmk.kmktime import PeriodicTimer
 from kmk.modules import Module
 from kmk.modules.mouse_keys import PointingDevice
@@ -58,7 +58,7 @@ def layer_key_validator(handler):
 
 
 class TrackballMode:
-    """Behaviour mode of trackball: mouse movement or vertical scroll"""
+    '''Behaviour mode of trackball: mouse movement or vertical scroll'''
 
     MOUSE_MODE = const(0)
     SCROLL_MODE = const(1)
@@ -150,7 +150,7 @@ class KeyHandler(TrackballHandler):
 
 
 class Trackball(Module):
-    """Module handles usage of Trackball Breakout by Pimoroni"""
+    '''Module handles usage of Trackball Breakout by Pimoroni'''
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


xs5871:
An additional 2D-encoder handler sounds great. I don't agree that this should be tied to layers, though.
We can make the handlers hot swappable, but I'd like to see this done with an interface that's independed of layers. An independed interface can always be bound to a layer switch key in a personal configuration.

fivetide:
I reworked the handling according to your suggestion: handlers are separated from the keyboard's layers and can be controlled by dedicated keycodes (KC.TB_NEXT_HANDLER, KC.TB_HANDLER(idx)). 
I could imagine that separating the pointing handler into one for scrolling and one for pointing, but that would break current handling of TrackballMode.MOUSE_MODE...

kmk/modules/pimoroni_trackball.py
```diff
@@ -1,13 +1,13 @@
-'''
+"""
```
xs5871, line 1: Single quotes are the prefered string denotation. Please revert all double quotes to single quotes where applicable.

kmk/modules/pimoroni_trackball.py
```diff
@@ -48,76 +48,158 @@
...
+                trackball.pointing_device.report_y[0] = 0xFF & y
+            trackball.pointing_device.hid_pending = x != 0 or y != 0
+        else:  # SCROLL_MODE
+            trackball.pointing_device.report_w[0] = y
+            trackball.pointing_device.hid_pending = y != 0
+
+        if switch == 1:  # Button pressed
+            trackball.pointing_device.button_status[
+                0
+            ] |= trackball.pointing_device.MB_LMB
```
xs5871, line 91:
```suggestion
            trackball.pointing_device.button_status[0] |= trackball.pointing_device.MB_LMB
```

kmk/modules/pimoroni_trackball.py
```diff
@@ -48,76 +48,158 @@
...
+            trackball.pointing_device.hid_pending = True
+
+        if not state and trackball.previous_state is True:  # Button released
+            trackball.pointing_device.button_status[
+                0
+            ] &= ~trackball.pointing_device.MB_LMB
```
xs5871, line 97:
```suggestion
            trackball.pointing_device.button_status[0] &= ~trackball.pointing_device.MB_LMB
```

kmk/modules/pimoroni_trackball.py
```diff
@@ -48,76 +48,158 @@
...
+    def __init__(
+        self,
+        i2c,
+        mode=TrackballMode.MOUSE_MODE,
+        address=I2C_ADDRESS,
+        angle_offset=ANGLE_OFFSET,
+        handlers=None,
+    ):
+        self.angle_offset = angle_offset
+        if handlers is None or len(handlers) == 0:
```
xs5871, line 152:
```suggestion
        if not handlers:
```

kmk/modules/pimoroni_trackball.py
xs5871: Sorry, but that is a terrible style and the linter should catch that (no offense).

kmk/modules/pimoroni_trackball.py
```diff
@@ -48,76 +48,158 @@
...
+            if x >= 0:
+                trackball.pointing_device.report_x[0] = x
+            else:
+                trackball.pointing_device.report_x[0] = 0xFF & x
+            if y >= 0:
+                trackball.pointing_device.report_y[0] = y
+            else:
+                trackball.pointing_device.report_y[0] = 0xFF & y
+            trackball.pointing_device.hid_pending = x != 0 or y != 0
+        else:  # SCROLL_MODE
```
xs5871, line 84: `SCROLL_MODE` is essentially just another `handler` and it would be nice to have a single mechanism to set modes / handlers. We can either implement this in this PR, or open an issue for a future improvement, whichever you favour.

kmk/modules/pimoroni_trackball.py
fivetide: none taken, seems that was black trying to adhere to the 80 char per line limit

kmk/modules/pimoroni_trackball.py
fivetide: the latest commit does that: adding a seperate handler for scroll, and deprecate the old parameter while providing a sane fallback

xs5871:
Awesome, this looks ready to go once the linter is happy.

kdb424:
`make fix-formatting fix-isort test` If that test says it's clean locally, commit and push. Otherwise, fix it's complaints until it is happy, but that should do most of it.

kdb424 (approved):

PR #451, opened 2022-05-23
fix crash on releasing dropped key presses

xs5871:
We discovered a crash when keys are mashed (see #424). This crash is unrelated to the original issue, but should still be mitigated. It occurs when keys are pressed _very_ fast and some key press events are dropped, but matching key releases are propagated.
It would be nice to fix this at a lower level, but I assume this already happens in the scanner from CP and thus out of our immediate influence.


kdb424 (approved):
Simple fex, looks solid.

PR #454, opened 2022-05-26
fix KC.LM

### LINKED ISSUE #453: KC.LM completely broken
Seems like this has been lost due to me being one of the very few who use this, and I haven't checked it in a while. I'll try to take a look, but I'm not sure how to fix it thus far.

xs5871:
Resolves #453.
I break it, I fix it (sorry btw).


kdb424 (approved):
Wasn't a big deal that it was broken. Amazed I'm the first to find it broken ><
Thanks for the swift fix!

PR #455, opened 2022-05-28
RGB matrix split side using drive name

daysgobye:
Rgb matrix code can now use the drive name to work out if it is on the right or left side of the keyboard.
You can still pass the bool if you want but it will use the drive name as default.

Also getting started was just a lie.


kdb424:
Looks clean. I'll merge once tests pass.

kdb424 (approved):

daysgobye:
fixed the formatting my bad on that one.

Issue #427, opened 2022-04-27
Automatic Modifier Chaining for OneShot Keys


ldsands:
So as recommended by @xs5871 on the matrix channel I used `Combos` to allow for chaining modifier keys while using `OneShot` keys. I even added it to `docs/oneshot.md`. Unfortunately, there is a consistent application of the chained modifiers to the next two keys tapped/pressed after the `OneShot` modifier(s). This does stop it the `timeout` is reached, so it is still useful but still causes problems. I've tried both `Chord` and `Sequence` from `Combos` and I've tested the `per_key_timeout` using both `True` and `False`. I realize that just waiting for the `timeout` works or shortening the `timeout`. For me these options either makes the modifier chaining using this method unusable for some (less commonly used) shortcuts or I forget and the modifiers apply anyway.

Ideally I would like to be able to do something like the following when loading the `OneShot` module:

```python
oneshot = OneShot(chain_modifiers=True)
```

Implementing this would almost certainly involve the chaining code to be located in the `kmk/modules/oneshot.py` and/or `kmk/modules/holdtap.py` files. If this were to be done in `oneshot.py` file I think this could be done in the `process_key` function. Below I just have a `ModifierKey` if statement and a print function (in `oneshot.py`) where I see this going but I'm not sure how to extract the modifier pressed (e.g. `KC.LCTL`) to then chain it and then return the chained modifiers to then apply it to the next key pressed/interruption or await the timeout.

```python
def process_key(self, keyboard, current_key, is_pressed, int_coord):
    '''Release os key after interrupting keyup.'''
    for key, state in self.key_states.items():
        if key == current_key:
            continue

        if state.activated == ActivationType.PRESSED and is_pressed:
            state.activated = ActivationType.HOLD_TIMEOUT
        elif state.activated == ActivationType.RELEASED and is_pressed:
            state.activated = ActivationType.INTERRUPTED
        elif state.activated == ActivationType.INTERRUPTED:
            if isinstance(current_key.meta.kc, ModifierKey):  # TESTCODE:
                print("current_key is a modifier key")  # TESTCODE:
            self.ht_released(key, keyboard)

    return current_key
```

The way I thought I would go about solving this would be to do a check for a modifier key when a `OneShot` key is interrupted. If the interrupted key is another `ModifierKey` then that modifier would be extracted added to the first `OneShot` modifier and then return the two (now) chained modifiers to then await another interruption or timeout. Does this seem like a good approach to the problem?

[...]

Issue #409, opened 2022-04-20
Add support for Key Overrides


rrotter:
Key Overrides in QMK allow remapping a key based upon which mods the user is holding. This allows, for example, switching the shift layer to produce different symbols, or swapping a base key with its shifted variant. 

[Docs for QMK implementation](https://github.com/qmk/qmk_firmware/blob/master/docs/feature_key_overrides.md)

xs5871:
Could be done with combos. We're missing some knobs for combos anyway and I'm currently working on making this possible as a side effect.

kilipan:
I've written this for my own usage, maybe someone more versed in python object orientation and kmk's internal logic could refactor it into something usable? (in an oop-implementation you'd also probably get rid of the need for a global variable)

```python
from kmk.keys import KC, make_key


mods_before_modmorph = set()
def modmorph(names = {'DUMMY_KEY',}, default_kc = KC.NO, morphed_kc = KC.NO, triggers = {KC.LSHIFT, KC.RSHIFT}):
    def _pressed(key, state, KC, *args, **kwargs):
        global mods_before_modmorph
        mods_before_modmorph = triggers.intersection(state.keys_pressed)
        # if a trigger is held, morph key
        if mods_before_modmorph:
            state._send_hid()
            for mod_kc in mods_before_modmorph:
                # discard triggering mods so morphed key is unaffected by them
                state.keys_pressed.discard(mod_kc)
            state.keys_pressed.add(morphed_kc)
            state.hid_pending = True
            return state
        # else return default keycode
        state.keys_pressed.add(default_kc)
        state.hid_pending = True
        return state
    def _released(key, state, KC, *args, **kwargs):
        if {morphed_kc,}.intersection(state.keys_pressed):
            state.keys_pressed.discard(morphed_kc)
            for mod_kc in mods_before_modmorph:
                # re-add previously discarded shift so normal typing isn't impacted
                state.keys_pressed.add(mod_kc)
        else:
            state.keys_pressed.discard(default_kc)
        state.hid_pending = True
        return state
    modmorph_key = make_key(names=names, on_press=_pressed,
                            on_release=_released)
    return modmorph_key
```

Define a mod morph / key override for e.g. a comma/semicolon key like this:
```python
modmorph({'CSEM', 'COMMA_SEMICOLON'}, KC.COMMA, KC.SEMICOLON)
```
Then simply use the keycode `KC.CSEM` or `KC.SEMICOLON` in your keymap.

[...]

PR #465, opened 2022-06-05
Enhancement combos

xs5871:
Implements enhancements mentioned in #409 and #427:
- more flexible combo reset handling: enables modmorph / key override functionality. Hold down a chord -> activate combo -> release and tap one of the chord keys -> repeat combo without having to tap the entire chord.
- faster combo activation: if there's only one combo left and matching, activate it without waiting for a timeout.

This took a lot longer to get right than it should. Turns out my initial design with dynamically changing match/reset lists is getting unmaintainable. Thought the matching could be faster with a map-reduce like approach, but let's be realistic: few people will have so many combos that reallocating them constantly into different lists is a better solution than iterating over all and checking a flag.
Will port this to tag / state machine like logic at some point.

### DIFF
--- a/docs/combos.md
+++ b/docs/combos.md
@@ -14,15 +14,19 @@ corresponding handlers.
 
 Combos may overlap, i.e. share match keys amongst each other.
 
-The optional arguments `timeout` and `per_key_timeout` define the time window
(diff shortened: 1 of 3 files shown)


kdb424 (approved):
Looks great!

PR #470, opened 2022-06-07
[debug] improve debug readability

crazyiop:
little tweak in the debug output to make things more readable:
- wrap KMKKeyboard report to 100 char (changeable), not splitting a key=value
- include a \n before MatrixChange as those are the begining of  a new logical 'block'

example result:
```
MatrixChange(ic=14, pressed=True)
KeyResolution(key=Key(code=1021, has_modifiers=None))
KMKKeyboard(
  debug_enabled=True, diode_orientation=1, matrix=(<MatrixScanner object at 0x2000eef0>,),
  unicode_mode=0, _hid_helper=USBHID(REPORT_BYTES=9), keys_pressed=set(),
  _coordkeys_pressed={14: Key(code=1021, has_modifiers=None)}, hid_pending=False,
  active_layers=[1, 0], _timeouts={},
)

MatrixChange(ic=52, pressed=True)
KeyResolution(key=Key(code=20, has_modifiers={1}))
KMKKeyboard(
  debug_enabled=True, diode_orientation=1, matrix=(<MatrixScanner object at 0x2000eef0>,),
  unicode_mode=0, _hid_helper=USBHID(REPORT_BYTES=9),
  keys_pressed={Key(code=20, has_modifiers={1})},
  _coordkeys_pressed={52: Key(code=20, has_modifiers={1}), 14: Key(code=1021, has_modifiers=None)},
  hid_pending=False, active_layers=[1, 0], _timeouts={},
)

MatrixChange(ic=52, pressed=False)
KeyResolution(key=Key(code=20, has_modifiers={1}))
KMKKeyboard(
  debug_enabled=True, diode_orientation=1, matrix=(<MatrixScanner object at 0x2000eef0>,),
  unicode_mode=0, _hid_helper=USBHID(REPORT_BYTES=9), keys_pressed=set(),
  _coordkeys_pressed={14: Key(code=1021, has_modifiers=None)}, hid_pending=False,
  active_layers=[1, 0], _timeouts={},
)

MatrixChange(ic=14, pressed=False)
KeyResolution(key=Key(code=1021, has_modifiers=None))
KMKKeyboard(
  debug_enabled=True, diode_orientation=1, matrix=(<MatrixScanner object at 0x2000eef0>,),
  unicode_mode=0, _hid_helper=USBHID(REPORT_BYTES=9), keys_pressed=set(), _coordkeys_pressed={},
  hid_pending=False, active_layers=[0], _timeouts={},
)
```

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -67,37 +67,27 @@ class KMKKeyboard:
     # real known fix yet other than turning off debug, but M4s have always been
     # tight on RAM so....
     def __repr__(self):
-        return (
-            'KMKKeyboard(\n'
-            '  debug_enabled={}, '
-            'diode_orientation={}, '
-            'matrix={},\n'
-            '  unicode_mode={}, '
-            '_hid_helper={},\n'
-            '  keys_pressed={},\n'
-            '  coordkeys_pressed={},\n'
-            '  hid_pending={}, '
-            'active_layers={}, '
-            'timeouts={}\n'
-            ')'
-        ).format(
-            self.debug_enabled,
-            self.diode_orientation,
-            self.matrix,
-            self.unicode_mode,
-            self._hid_helper,
-            # internal state
-            self.keys_pressed,
-            self._coordkeys_pressed,
-            self.hid_pending,
-            self.active_layers,
-            self._timeouts,
+        return ''.join(
+            [
+                'KMKKeyboard(\n',
+                f'  debug_enabled={self.debug_enabled}, ',
+                f'diode_orientation={self.diode_orientation}, ',
+                f'matrix={self.matrix},\n',
+                f'  unicode_mode={self.unicode_mode}, ',
+                f'_hid_helper={self._hid_helper},\n',
+                f'  keys_pressed={self.keys_pressed},\n',
+                f'  _coordkeys_pressed={self._coordkeys_pressed},\n',
+                f'  hid_pending={self.hid_pending}, ',
+                f'active_layers={self.active_layers}, ',
+                f'_timeouts={self._timeouts}\n',
+                ')',
+            ]
         )
 
     def _print_debug_cycle(self, init=False):
         if self.debug_enabled:
             if init:
-                print('KMKInit(release={})'.format(KMK_RELEASE))
+                print(f'KMKInit(release={KMK_RELEASE})')
             print(self)
 
     def _send_hid(self):
@@ -107,7 +97,7 @@ class KMKKeyboard:
                 hid_report.send()
             except KeyError as e:
                 if self.debug_enabled:
-                    print('HidNotFound(HIDReportType={})'.format(e))
+                    print(f'HidNotFound(HIDReportType={e})')
         self.hid_pending = False
 
     def _handle_matrix_report(self, update=None):
@@ -120,7 +110,7 @@ class KMKKeyboard:
             idx = self.coord_mapping.index(int_coord)
         except ValueError:
             if self.debug_enabled:
-                print('CoordMappingNotFound(ic={})'.format(int_coord))
+                print(f'CoordMappingNotFound(ic={int_coord})')
 
             return None
 
@@ -141,7 +131,7 @@ class KMKKeyboard:
         int_coord = kevent.key_number
         is_pressed = kevent.pressed
         if self.debug_enabled:
-            print('\nMatrixChange(ic={}, pressed={})'.format(int_coord, is_pressed))
+            print(f'\nMatrixChange(ic={int_coord}, pressed={is_pressed})')
 
         key = None
         if not is_pressed:
@@ -156,11 +146,11 @@ class KMKKeyboard:
 
             if key is None:
                 if self.debug_enabled:
-                    print('MatrixUndefinedCoordinate(ic={})'.format(int_coord))
+                    print(f'MatrixUndefinedCoordinate(ic={int_coord})')
                 return self
 
         if self.debug_enabled:
-            print('KeyResolution(key={})'.format(key))
+            print(f'KeyResolution(key={key})')
 
         self.pre_process_key(key, is_pressed, int_coord)
 
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -68,17 +68,17 @@ class KMKKeyboard:
     # tight on RAM so....
     def __repr__(self):
         return (
-            'KMKKeyboard('
-            'debug_enabled={} '
-            'diode_orientation={} '
-            'matrix={} '
-            'unicode_mode={} '
-            '_hid_helper={} '
-            'keys_pressed={} '
-            'coordkeys_pressed={} '
-            'hid_pending={} '
-            'active_layers={} '
-            'timeouts={} '
+            'KMKKeyboard(\n'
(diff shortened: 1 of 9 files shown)


xs5871:
I've been working on better debug output on the side and didn't have the opportunity to commit anything yet.
In my opinion `_print_debug_cycle` shouldn't print all that static, never changing clutter on _every_ key event.
We can show the entire keyboard configuration at boot/init once, and the debug cycle should only contain relevant information, i.e. `active_layers`, `_coordkeys_pressed` and maybe `keys_pressed`.

kmk/kmk_keyboard.py
```diff
@@ -67,32 +68,32 @@ class KMKKeyboard:
...
+            '_hid_helper',
+            # internal stat
+            'keys_pressed',
+            '_coordkeys_pressed',
+            'hid_pending',
+            'active_layers',
+            '_timeouts',
+        ]:
+            attr = getattr(self, attr_name)
+            part = f'{attr_name}={attr}, '
```
klardotsh, line 87: I'm unsure how I feel about `getattr` in this context since this becomes less statically analyzable (similar to my vote in the other PR #461) - I love the premise of what this PR is getting at (especially as an intermediate step until a more robust debug framework is in place); would it be feasible to loop over the `self.*` attrs as references rather than strings to look up with `getattr`? perhaps

```python
for ref in [ self.debug_enabled, self.diode_orientation, ... ]:
```

klardotsh (approved):
Approving the idea; I have a remark inline that OP and @xs5871 can work out whether they want to take or leave but aside from that this LGTM.

kmk/kmk_keyboard.py
xs5871: TL;DR: agree with @klardotsh, but won't block anything

`getattr` and `self.__getattr__` are exactly the same thing. The problem comes from the `attr_name`. You can put whatever in there, and no _static_ analysis will ever find out if it resolves to a valid attribute, because it's by design a _dynamic_ attribute reference. The only way to check for validity is to run the code, which in turn defeats the purpose of a static analysis.

More importantly: verbosity isn't necessarily a bad thing, it's about finding the sweet spot for human parsability. Writing code with the fewest possible symbols is very exciting and deserves it's own space and recognition. However: it's hardly conducive in project that relies on a community of contributers with a wide range of coding background and familiarity with the code base.
In this day and age (apologies for the platitude), programming is 20% writing your own and 80% reading other peoples code. It's important (and will silently be appreciated by others) to write code that can first and foremost be understood by people with less experience. Otherwise we'd be putting inline assembly everywhere.
To be clear: I agree that code should be as concise as possible, but not conciser (I think I'm misquoting Einstein).

To give that context to this PR:
I'd go even further than @klardotsh and suggest: make it an f-string with static newlines embedded. Then the output will look exactly like the string is formated in the code, which is nice. Put the potentially long attributes in seperate lines. Once those wrap you've got so much text on screen, that dynamic line wrapping isn't going to meaningfully improve readability anyway.
I won't be blocking this PR, it's an improvement to the status quo, even with the `getattr`. It's at a place where it's not going to do much harm, but be aware that it may get refactored at some point.

crazyiop: Ok, I have updated to the fixed multiline way then.

kmk/kmk_keyboard.py
```diff
@@ -67,32 +67,17 @@ class KMKKeyboard:
...
-            self.active_layers,
-            self._timeouts,
+        txt = 'KMKKeyboard(\n'
+        txt += '  debug_enabled={}, diode_orientation={}, matrix={}\n'.format(
+            self.debug_enabled, self.diode_orientation, self.matrix
         )
+        txt += f'  unicode_mode={self.unicode_mode}, _hid_helper={self._hid_helper},\n'
+        txt += f'  keys_pressed={self.keys_pressed}\n'
+        txt += f'  _coordkeys_pressed={self._coordkeys_pressed}\n'
+        txt += '  hid_pending={}, active_layers={}, _timeouts={}\n)'.format(
```
kdb424, line 77: Why is this not an f-string but the others are?

kmk/kmk_keyboard.py
```diff
@@ -141,7 +126,7 @@ def _on_matrix_changed(self, kevent):
         int_coord = kevent.key_number
         is_pressed = kevent.pressed
         if self.debug_enabled:
-            print('MatrixChange(ic={} pressed={})'.format(int_coord, is_pressed))
+            print('\nMatrixChange(ic={}, pressed={})'.format(int_coord, is_pressed))
```
kdb424, line 129: Maybe also should be an f-string?

kdb424 (changes requested):

kmk/kmk_keyboard.py
crazyiop: those are not fstring as that would make the line too long. The goal of fstring is to give an option that in most case is  a more readable string. Here the more readable, given the lenght, is the format. as it is  each argument is below it's corresponding {}.
It was also an attempt to have both not too long line, and "the output will look exactly like the string is formated in the code" you suggested.

I can split it into more fstring line if the consistency of string formatting is wanted. (I can even push a commit to remove all `.format()` given the low number of them if you want).

kdb424: I'll leave the final call to @xs5871 but I believe that consistency makes more sense personally, but I suppose that's a valid point.

kdb424 (approved):

kmk/kmk_keyboard.py
xs5871: +1 to consistency.
When I suggested "make it an f-string" I had a single multi-line f-string in mind. The `x += '*\n'` feels quite unpythonic.

I have to say, this discussion has been going on for almost to long, considering this concerns such a minor change.
It's probably because we started with a complex, possibly over-engineered (no offense, we all do over-engineer sometimes; also it's fun), solution and step for step reverted it back to a pragmatic one, instead of starting from scratch.
At this point, we could just as well keep the original code, just with a couple of newlines thrown into it.

kdb424 (approved):
I know these aren't the changes you originally wanted in there, but I believe this is easy to read, and improves debug-ability quite a bit.

kdb424:
`make fix-formatting fix-isort test` should auto fix errors, and give you the test output before committing and pushing if you weren't aware.

crazyiop:
Sorry I did not fully tested the fstring change before upload but got back to my board:
- some attribute were wrong as the previous KMKKeyboard log hat for instance "timeouts={}" but it is actually self._timeouts use in the format parameter (I failed to see the difference).
- circuitpython cannot implicitely concat multiple fstring. So I put them in a list with a ''.join() as a workaround.

xs5871:
Looks good, verified on hardware. Thanks for sticking with the issue.
I'm going to go ahead and merge.

klardotsh:
Ah, yeah, when I wrote CircuitPython's f-strings implementation I left some comments about that concatenation error. Technically this means Micro+CircuitPython are not actually PEP-498 compliant, but both upstreams agreed with me that the tradeoff was worthwhile. Definitely confusing the first time one runs into it, though :\ https://github.com/adafruit/circuitpython/pull/2690/files#diff-e4737fa9a48c19ab35f9083a424c50059a6e94008f7d2aeaa4418eab7b3bc332R86-R102

PR #471, opened 2022-06-08
[rgb] fix set_rgb_fill method

crazyiop:
The code was wrongly iterating the neopixel, running `fill()` on every
individual pixels.

I suggested the change that was sucessfully tested by @ziptyze on discord.

### DIFF
--- a/kmk/extensions/rgb.py
+++ b/kmk/extensions/rgb.py
@@ -287,10 +287,9 @@ class RGB(Extension):
         Takes an RGB or RGBW and displays it on all LEDs/Neopixels
         :param rgb: RGB or RGBW
         '''
-        for pixels in self.pixels:
-            pixels.fill(rgb)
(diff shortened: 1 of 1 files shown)


kdb424 (approved):
Simple mistake, but huge issue. Thanks.

xs5871:
This breaks the RGB extension. `RGB.pixels` is a list of pixel buffer objects, not a bare pixel buffer.
The original code is correct and working.

crazyiop:
@xs5871 ha yes I have  read the code too quickly. I guess then that for some reason on @ziptyze board the line https://github.com/KMKfw/kmk_firmware/blob/90c4ce99f3ff781f8361f01c847ed0d60b037748/kmk/extensions/rgb.py#L123 did not correctly trigger...

kdb424:
Welp. This is why we have a revert button. Coffee, then reviews >_>

PR #472, opened 2022-06-08
Revert "[rgb] fix set_rgb_fill method"

kdb424:
Reverts KMKfw/kmk_firmware#471
More coffee, less merging in the morning.

### DIFF
--- a/kmk/extensions/rgb.py
+++ b/kmk/extensions/rgb.py
@@ -287,9 +287,10 @@ class RGB(Extension):
         Takes an RGB or RGBW and displays it on all LEDs/Neopixels
         :param rgb: RGB or RGBW
         '''
-        self.pixels.fill(rgb)
-        if not self.disable_auto_write:
-            self.pixels.show()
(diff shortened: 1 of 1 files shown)


klardotsh:
What's the rationale for the revert? I didn't follow the original PR going in so I have zero context on what this is and why we're reverting it.

klardotsh:
Further, once reverted, what's the plan forward? Should this functionality be rebuilt? Patched and re-merged in a similar form to what exists? Forgotten entirely?

xs5871:
Original PR ment to "fix" a problem that a user in chat was having. `RGB.pixels` implements a list of pixel buffers to iterate over (for multiple target enpoints), and they managed to assign a bare pixel buffer to it.
Obviously, the fix "worked" with their erroneous configuration, but breaks everyone elses.

There's no functionality to be rebuild, everything was already working as intended. I actually re-tested every possible RGB configuration (*I could think) and could not reproduce their issue.

xs5871 (approved):

PR #482, opened 2022-06-18
Improve split doc

crazyiop:
I have recently setup my split board without trouble with the doc but also bylooking at the source. I have since helped a few peoples understand some part of the split parameters on the discord. The doc is not clear enough and many subtle things only get understood by reviewing the code, what some user are not capable of. This is an attempt to fix those grey area in the split doc

I have also removed an unused (duplicate) parameter of the split module.

I'll appreciate english reviewer. As english is not my primary language, I have certainly some errors in there...


docs/split_keyboards.md
```diff
@@ -35,38 +49,71 @@ split = Split(
...
-# Wired
-split = Split()
-# Wireless
-split = Split(split_type=SplitType.BLE)
+from storage import getmount
+
+side = SplitSide.RIGHT if str(getmount('/').label)[-1] == 'R' else SplitSide.LEFT
+
+if side == SplitSide.RIGHT:
+    keyboard.col_pins = ...
```
kdb424, line 79: This is only needed if PCB's are not interchangeable between sides, which is the common practice. maybe note that this is optional if they have different pinouts.

docs/split_keyboards.md
```diff
@@ -35,38 +49,71 @@ split = Split(
...
-Rename your CIRCUITPY drive to something different. The left side must
-end in L, the right must is in R. The name must be 11 characters or less! This is
-a limitation of the filesystem. You will receive an error if you choose a name
-longer than 11 characters. Instructions on how to do that are
-[here](https://learn.adafruit.com/welcome-to-circuitpython/renaming-circuitpy).
-For example on NYQUISTL for left and NYQUISTR for the right.
+This is a sane default settings.
+
+But you still can choose to set all that manually. Some reasons would be:
+- easy both side debug and upload by having both halves connected to usb (and uart between them). Explicit `split_side` and only one `split_target` is needed to avoid each half acting as a `split_target` (duplicating each keypress).
```
kdb424, line 70: Scentences start with a capital letter. `easy` should be `Easy` as an example. This happens many places. Please check them all that were added to keep the docs as consistent as possible.

docs/split_keyboards.md
```diff
@@ -76,3 +123,22 @@ In order to enable it, you must:
...
 - pass `use_pio=True` into `Split()` constructor
+
+The file is available [here](https://github.com/adafruit/Adafruit_CircuitPython_PIOASM/releases). From the `adafruit-circuitpython-pioasm-x.y-mpy-0.x.y.zip`, copy the `lib` folder (or its content) to the root of your circuitpython drives. Make sure to get the zip corresponding to your circuitpython version as the mpy format can evolve!
+
+Continuing our example of the KB2040, one side's UART being fliped will need to use a pin capable of only tx as rx (and vice-versa). So in order to use that qwiic/stemma connector you will also need `use_pio = True`.
+
+#### `data_pin`/`data_pin2`
+For UART `SplitType`: on the master or `split_target` side, `data_pin` is the one use for RX, `data_pin2` the one for TX.
+
+### EE HANDS
```
kdb424, line 134: Because this name is confusing for some, maybe we change this to `### EE HANDS/AUTO HANDEDNESS`

kdb424 (changes requested):
Over all, I think this is useful. We may also want to merge the `Drive names` section with `EE HANDS` section as they are the same thing in the end. I can't comment on the code change implications, but the docs look like it's mostly an improvement.

docs/split_keyboards.md
```diff
@@ -35,44 +48,92 @@ split = Split(
...
+- Install Circuit Python in 7.2+ version
+- Add `adafruit_pioasm.mpy` library to lib or root folder of a board
+- Pass `use_pio=True` into `Split()` constructor
+
+The file is available [here](https://github.com/adafruit/Adafruit_CircuitPython_PIOASM/releases). From the `adafruit-circuitpython-pioasm-x.y-mpy-0.x.y.zip`, copy the `lib` folder (or its content) to the root of your circuitpython drives. Make sure to get the zip corresponding to your circuitpython version as the mpy format can evolve!
+
+Continuing our example of the KB2040, one side's UART being fliped will need to use a pin capable of only tx as rx (and vice-versa). So in order to use that qwiic/stemma connector you will also need `use_pio = True`.
+
+#### `data_pin`/`data_pin2`
+For UART `SplitType`: on the master or `split_target` side, `data_pin` is the one use for RX, `data_pin2` the one for TX.
```
kdb424, line 128: We don't use master/slave naming schema. Please use "target" in term of "master" and "initiator" in term of "slave" at any point in docs/code. If that has slipped into the docs/code, it's a bug, and should be changed to the intended terminology.

crazyiop:
@kdb424 I apllied your suggestions/remarks.

About the drives names/ EE HANDS sections. They were actually merged before. EE HANDS needs the drives to be renammed but there are other benefits to renammed them. I thought that the renamming deserved it's own part, and should even suggest people to adopt it even if they are not using EE HANDS to at least put people in a similar state in this regards.

kdb424:
> @kdb424 I apllied your suggestions/remarks.
> 
> About the drives names/ EE HANDS sections. They were actually merged before. EE HANDS needs the drives to be renammed but there are other benefits to renammed them. I thought that the renamming deserved it's own part, and should even suggest people to adopt it even if they are not using EE HANDS to at least put people in a similar state in this regards.

Maybe I missed some changes, but other than the code and therefor the ability to swap hands, being the same is improved? I believe that it should be done regardless (especially on wired boards) because of the cable situation, but code simplification was the primary goal, and the convenience of which side it's plugged into was secondary to me when adding it originally.

crazyiop:
@kdb424 renamming the drives allow:
- to have the kind of messy hardware discrepancy like the example code pushed (I confess I have something similar on my board) but not using ee hands as I want to connect both side to the pc to get both debug while playing with the kmk internals. I would even say that the 'messy code example' might be the guideline to handle hardware discrepancy (I've seen at least once that I remember on the discord someone who implied they used two different code, one for each side, for this).
- having clear drive names allow me to use the kmk keyboard while working on another electronic project who also use a circuitpython drive and not have all names being 'CIRCUITPY' and create confusion... (still no need to use ee hands to benefits from this)

Those are the two situation I'm in and see the benefits on renaming the drives without using ee hands yet. I guess some more might exists...

kdb424:
> @kdb424 renamming the drives allow:
> 
>     * to have the kind of messy hardware discrepancy like the example code pushed (I confess I have something similar on my board) but not using ee hands as I want to connect both side to the pc to get both debug while playing with the kmk internals. I would even say that the 'messy code example' might be the guideline to handle hardware discrepancy (I've seen at least once that I remember on the discord someone who implied they used two different code, one for each side, for this).
> 
>     * having clear drive names allow me to use the kmk keyboard while working on another electronic project who also use a circuitpython drive and not have all names being 'CIRCUITPY' and create confusion... (still no need to use ee hands to benefits from this)
> 
> 
> Those are the two situation I'm in and see the benefits on renaming the drives without using ee hands yet. I guess some more might exists...

That's fair. I wasn't suggesting that anything be *removed* from the docs, but basically a note left that some things, like defining rows/cols per side is optional so people don't assume that's the default for splits. It's all good info, but I hope to present it in the way most clear to people going in completely new so they see things that are there, and know what parts are optional, or only apply to some situations is all.

crazyiop:
@kdb424 Sorry I think I misunderstood you previous message.

I have already updated the wording for that part to:

 "Keep in mind that the normal case (interchangeable pcb between sides) do **not** need this. The following code is **not** a guideline, but an extreme example showing the flexibility of kmk (would realistically be applicable only in messy handwired keyboard):"

I think that makes it clear that it is not 'the example to follow'.

docs/split_keyboards.md
```diff
@@ -35,44 +48,92 @@ split = Split(
...
+
+#### `split_target_left`
+The `split_target` refer to the side that will send keypress over usb.
 
-If you're using RP2040 based board and want to use other pins that the one with hardware support of RX/TX, you can try
-using PIO implementation of two-way wire protocol. Typical use cases for it are premade boards, made with QMK's
-bitbanging protocols in mind.
+If you use EE HANDS (`split_side = None`) this parameter will be automatically overriden (a side connected to a usb host will act as a `split_target`).
+
+If setting this manually, be sure to reflect the actual usb connection and make sure to use the same value for both sides !
```
xs5871, line 105:
```suggestion
Setting `split_side = None` (similar to EE HANDS in QMK) this parameter will be overriden.
```
* KMK doesn't have EE HANDS. It's a similar feature that just copied the name from QMK.
* I removed some redundant text.

docs/split_keyboards.md
```diff
@@ -35,44 +48,92 @@ split = Split(
...
-- install Circuit Python in 7.2+ version
-- add `adafruit_pioasm.mpy` library to lib or root folder of a board
-- pass `use_pio=True` into `Split()` constructor
+- Install Circuit Python in 7.2+ version
+- Add `adafruit_pioasm.mpy` library to lib or root folder of a board
+- Pass `use_pio=True` into `Split()` constructor
+
+The file is available [here](https://github.com/adafruit/Adafruit_CircuitPython_PIOASM/releases). From the `adafruit-circuitpython-pioasm-x.y-mpy-0.x.y.zip`, copy the `lib` folder (or its content) to the root of your circuitpython drives. Make sure to get the zip corresponding to your circuitpython version as the mpy format can evolve!
+
+Continuing our example of the KB2040, one side's UART being fliped will need to use a pin capable of only tx as rx (and vice-versa). So in order to use that qwiic/stemma connector you will also need `use_pio = True`.
```
xs5871, line 125:
```suggestion
The file is available [here](https://github.com/adafruit/Adafruit_CircuitPython_PIOASM/releases). From the `adafruit-circuitpython-pioasm-x.y-mpy-0.x.y.zip`, copy the `lib` folder (or its content) to the root of your circuitpython drives. Make sure to get the zip corresponding to your circuitpython version as the mpy format can evolve!

Continuing our example of the KB2040, one side's UART being fliped will need to use a pin capable of only tx as rx (and vice-versa). So in order to use that qwiic/stemma connector you will also need `use_pio = True`.
```
* Please don't use "[here]" as a name for a link. Give it the name or description it's pointing at.
* The second paragraph is very confusing again. When did we get from RP2040 to KB2040? What does "use a pin capable of only tx as rx" mean? Why are we talking about qwicc/stemma suddenly? Sorry, but this is all very incoherent.

xs5871 (changes requested):
General note: check the heading levels. Currently "Config" is a sub-heading of "Bluetooth split".
In addition the general "Config" sits in the middle of the document, with more specific configuration hints before and afterwards.

docs/split_keyboards.md
```diff
@@ -35,44 +48,92 @@ split = Split(
...
+
+If setting this manually, be sure to reflect the actual usb connection and make sure to use the same value for both sides !
+
+#### `uart_flip`
+If your boards are connected through the **same** pins (like gpio**4** of board A to gpio**4** of board B): use `uart_flip = True`.
+
+If your boards are connected through **different** pins (like gpio**4** of board A to gpio**10** of board B): use `uart_flip = False`.
+
+For example if using the qwiic/stemma connector of the [KB2040](https://www.adafruit.com/product/5302) (supposed to be for I2C) for the uart connection. You will need to use `uart_flip = True` as for i2c sda connect to sda and scl to scl.
+
```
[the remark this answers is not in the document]
crazyiop: I removed the example to avoid confusion.

xs5871 (approved):
Heading levels and order are still a bit off, but we can tackle that another day.

crazyiop:
@xs5871 I fixed the heading. Sorry I missed the first comment about it !

xs5871 (approved):

kdb424 (approved):
Looks good to me. I've been silently following this. Guess I'll merge.

Issue #347, opened 2022-03-01
Rotary encoder divisor


mbellitti:
I am trying to add a rotary encoder to my build. The model I am using has 24 retents but 12 cycles per rotation, so with the default settings KMK picks up every other "click" (rather than all of them).

The problem is solved in `rotaryio` by setting the `divisor` flag to `2` (see [here](https://github.com/adafruit/circuitpython/pull/5468)). Is there something similar in the current implementation of KMK?

kdb424:
Try changing the movement from 1 to 2 on lines 50, 56, and 58 (make that -2), and let us know how that works.

mbellitti:
I've played around a little bit with the encoder module and I understand what the problem is. 

The pin output follows the sequence 00>01>11>10 when turned in one direction and 11>01>00>10 when turned in the other. The retents (the stable positons of the encoder, the "clicks") are on both 00 and 11, but the current version of the code only updates when the new state is 11 (see [here](https://github.com/KMKfw/kmk_firmware/blob/067b7d2e5fc3ea963fa85806974404019de443fe/kmk/modules/encoder.py#L61)). 

I've modified the code to make it work in my case, but I'm seeing some bouncing when I turn the wheel too fast, which is really annoying. 

My solution looks like this 
```python
    def update_state(self):
        # Rotation events
        new_state = (self.pin_a.get_value(), self.pin_b.get_value())

        if new_state != self._state:

            if new_state[0] == new_state[1]:
                print(new_state, self._state,end="\n\n")
                if new_state[0] == self._state[0]:
                    self._direction = 1
                else:
                    self._direction = -1

                self._pos += self._direction
                self.on_move_do(self.get_state())

            self._state = new_state
```
and it mostly works. I think a more robust solutions would be to use `rotaryio`, so I think I will write a wrapper around that. Maybe we could have two version of the encoder module, one that uses rotaryio and one that doesn't. If I understand correctly the main reason not to use rotaryio is that it requires using two consecutive pins, but I think that's perfectly fine.

kdb424:
If you port to roteryio, we'll rope someone in to help merge the modules in if you need help, but we prefer, unless you can present a good case, to have only one module for a feature to prevent confusion. That doesn't mean we aren't interested in your issue, but solving it the right way for everyone, so we'll work with you on it. The discord/matrix is great for active help

mbellitti:
[...]

PR #484, opened 2022-06-21
Encoder resolution setting

zyumbik:
Added support for GPIO encoders with different resolutions: previously only ones with 4 state changes per detent were supported, I added support for encoders with 2 state changes per detent. The resolution can be set globally (before initializing the pins) or individually for each encoder.

I called the setting “resolution” as it's not technically the same as “divisor” (e.g. in rotaryio). Also QMK uses the same name and I'm used to it, I find it quite intuitive. 

Could be a better (no rotaryio) alternative to #357. Or both could be used without conflicts. Also this solves #347.

Tested with 5 different encoders on separate pins (vertical 24 and 20 detents with resolution = 4, EVQWGD001 roller with resolution 4, two models of encoders with 30 detents per revolution with resolution = 2), everything works as expected, no issues whatsoever. No skipped pulses or double activations. 

Docs updated (and slightly improved overall), tests and formatting ran.

Haven't looked into adding this for I2C encoders: they seem quite rare, I don't have any.

### DIFF
--- a/docs/encoder.md
+++ b/docs/encoder.md
@@ -1,32 +1,31 @@
 # Encoder module
-Add twist control to your keyboard! Volume, zoom, anything you want
+Add twist control to your keyboard! Volume, zoom, anything you want.
 
-I2C encoder type has been tested with the Adafruit I2C QT Rotary Encoder with NeoPixel
+I2C encoder type has been tested with the Adafruit I2C QT Rotary Encoder with NeoPixel.
 
-**Note:** If split with encoders and both sides should work, it's currently necessary to use the encoder-scanner explained at the bottom of [scanners docs](scanners.md).
+**Note:** If you have a **split** keyboard with encoders and **both sides** should work, it's currently necessary to use the encoder-scanner explained at the bottom of [scanners docs](scanners.md).
 
 ## Enabling the extension
-The constructor(`EncoderHandler` class) takes a list of encoder, each one defined as either:
+The constructor(`EncoderHandler` class) takes a list of encoders, each one defined as either:
 
-* a list of pad_a pin, pad_b pin, button_pin and optionally a flag set to True is you want it to be reversed
+* a list of `pad_a` pin, `pad_b` pin, `button_pin` and optionally a flag set to `True` is you want encoder direction to be reversed
 * a `busio.I2C`, address and optionally a flag set to True if you want it to be reversed
 
-The encoder_map is modeled after the keymap and works the same way. It should have as many layers (key pressed on "turned left", key pressed on "turned right", key pressed on "knob pressed") as your keymap, and use KC.NO keys for layers that you don't require any action.
-The encoder supports a velocity mode if you desire to make something for video or sound editing. 
+The `encoder_map` is modeled after the keymap and works the same way. It should have as many layers (key pressed on "turned left", key pressed on "turned right", key pressed on "knob pressed") as your keymap, and use `KC.NO` keys for layers that you don't require any action. The encoder supports a velocity mode if you desire to make something for video or sound editing. 
 
 
 
 ## How to use
-How to use this module in your main / code file
+How to use this module in your `main.py` / `code.py` file.
 
-1. load the module
+1. Load the module.
 ```python
 from kmk.modules.encoder import EncoderHandler
 encoder_handler = EncoderHandler()
 keyboard.modules = [layers, modtap, encoder_handler]
 ```
 
-2. Define the pins for each encoder (pin_a, pin_b, pin_button [or `None` if the encoder's button is handled differently or not at all], True for an inversed encoder)
+2. Define the pins for each encoder: `pin_a`, `pin_b` for rotations, `pin_button` for the switch in the encoder (set to `None` if the encoder's button is handled differently or not at all). If you want to invert the direction of the encoder, set the 4th (optional) parameter `is_inverted` to `True`.
 ```python
 #GPIO Encoder
 encoder_handler.pins = ((board.GP17, board.GP15, board.GP14, False), (encoder 2 definition), etc. )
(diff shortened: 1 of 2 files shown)


kdb424 (approved):
I like this, but I'll wait for someone to do a code review proper. Looks a lot more readable.

docs/encoder.md
```diff
@@ -42,27 +43,49 @@ SDA = board.GP0
...
+    (board.GP14, board.GP15, None, False, 4), 
+    (board.GP26, board.GP27, None, False, 2),
+    (board.GP26, board.GP27, None ), # will be set to global default
+)
+```
+
+## Handler methods overrides
 
+Encoder methods `on_move_do` and `on_button_do` can be overridden for complex use cases.
+
```
xs5871, line 85:
```suggestion
```

kmk/modules/encoder.py
```diff
@@ -58,19 +58,21 @@ def update_state(self):
...
-                    # some steps may be missed. This makes it behaves more
-                    # naturally
-                    real_movement = round(self._movement / 4)
+            if new_state[0] == new_state[1]:
+                # when the encoder made a full loop according to its resolution
+                if self._movement >= self.resolution - 1:
+                    # 1 full step is 4 movements (2 for high-resolution encoder),
+                    # however, when rotated quickly, some steps may be missed.
+                    # This makes it behave more naturally
+                    real_movement = round(self._movement / self.resolution)
```
xs5871, line 67:
```suggestion
                    real_movement = self._movement // self.resolution
```
prefer integer division, needs test on hardware

xs5871 (changes requested):
It's a bit of an ask, but if possible I'd prefer using the same parameter convention as [rotaryio](https://docs.circuitpython.org/en/latest/shared-bindings/rotaryio/index.html#) to spare us same headaches in the future.
That means replacing `resolution` with `divisor` (which seems to be the inverse ?).

zyumbik:
I wanted to do that but I couldn't figure out how since he divisor works slightly differently, it's not just inverted (see [rotaryio](https://docs.circuitpython.org/en/latest/shared-bindings/rotaryio/index.html) docs). 

> The divisor of the quadrature signal. Use 1 for encoders without detents, or encoders with 4 detents per cycle. Use 2 for encoders with 2 detents per cycle. Use 4 for encoders with 1 detent per cycle.

It's 1 for 4 and 2 for 2. I checked the source and it's all very complicated to me.

Although now that you mention it, I can see the way I can try to change it. Will try today.

Also not sure what to do with encoders that have 1 detent per cycle (I don't even know where to find these to test). For those I can't guarantee if it's gonna work (if I understood how rotaryio works maybe I could). But I think it's ok if we omit those for now.

docs/encoder.md
zyumbik: This was in the original docs so I thought I'd keep it. I think it can be useful to know it's possible even though no examples are given at the moment.

zyumbik: Ah, sorry, for some reason I thought this meant removing text. Thanks for your edits! :)

zyumbik:
I started reworking this and thinking about how bad of a name “divisor” is. And then was like “wait a minute, this is already correct!” Let's look at what the rotaryio docs say again:

> Use 1 for encoders without **detents**, or encoders with 4 detents per **cycle**.

**Detent** is a “click” of encoder as far as I understand. When the encoder is idle, it stays at a detent. When the encoder reaches a detent, it should output an action (such as a keypress).

**Cycle** is always exactly the period of 4 changes of the signal. `a: 0011` and `b: 0110` for CW rotation. The start and end of the period is the same, hence it's a cycle.

This article helped me understand this very well: https://www.electronicspecifier.com/products/sensors/understanding-the-encoder-resolution

This makes perfect sense for smooth encoders (without detents) to have a resolution (“divisor”) of 1 because every tiny rotation (any single from the cycle of four) would correspond to the imaginary detent and should produce an action event (keypress). For most common encoders with detents, the encoder completes all 4 steps of the cycle per single detent. And for more precise ones that I have, it produces only two state changes per detent, hence the resolution is 2.

I'll look through the code again with this new information in mind and commit if I find ways to improve it. And now that I know how encoders with resolution of 1 work, I'll be able to implement that. 

-----

Regarding the name. I'm not sure if you implied we need to change the name or only make the `resolution` function the same as `divisor`, but in any case I would suggest keeping the name as “resolution”. In my opinion, “divisor” is a bad name, I don't want to spread it further. If you don't know exactly how encoders work (which was me the day before sending the PR), you won't be able to understand it and use it intuitively. However it might be useful to mention in the docs that resolution corresponds to divisor in rotaryio (in case anyone getting started with KMK is already familiar with it — I wasn't). 

Additionally, `resolution` in my code and understanding corresponds to the number of state changes to fire the action event (if state changes > resolution, do action). The term `divisor` on the other hand sounds like it is the number by which the cycle should be divided to get the number of state changes (if state changes > 4 / divisor, do action) — but it doesn't work that way in rotaryio. It works the same as I described `resolution`. So either they use the term incorrectly (since it doesn't divide anything) or I'm missing something.

Sorry for lots of text!

zyumbik:
I'm pretty happy with the changes thus far, I think it's ok to be merged. Let me know if I need to change anything else.

Firstly, I modified one of my encoders to be smooth and thus have the resolution of 1. I couldn't figure out a way to make it work with resolution of 1 without false activations (maybe possible with debouncing) but I was able to adjust the code to make it work reliably with resolution of 2 (which also improved the handling of all encoders).

Secondly, I modified the docs to address the rotaryio divisor, improved some things overall and added the encoder pinout to the docs as someone suggested in the discord. 

Ran the tests and formatting tools again, all seems to be ok.

docs/encoder.md
```diff
@@ -42,27 +48,67 @@ SDA = board.GP0
...
+encoder_handler.map = [ (( KC.VOLD, KC.VOLU, KC.MUTE), (encoder 2 definition), etc. ), # Layer 1
+                        ((Zoom_out, Zoom_in, KC.NO), (encoder 2 definition), etc. ), # Layer 2
+                        ((KC.A, KC.Z, KC.N1), (encoder 2 definition), etc. ), # Layer 3
+                        ((KC.NO, KC.NO, KC.NO), (encoder 2 definition), etc. ), # Layer 4
                       ]
 ```
 
+## Encoder resolution
+
+Depending on your encoder precision, it may send 4 or 2 pulses on every detent. By default the encoder resolution is set to 4, but if your an encoder only activates on every second detent (skips pulses), set the resolution to 2. If the encoder activates twice on every detent, set the value to 4.
```
xs5871, line 70:
```suggestion
Depending on your encoder precision, it may send 4 or 2 pulses on every detent. By default the encoder resolution is set to 4, but if your encoder only activates on every second detent (skips pulses), set the resolution to 2. If the encoder activates twice on every detent, set the value to 4.
```

xs5871 (commented):
I agree that the description of `divisor` in `rotaryio` isn't great, but I disagree that `resolution` is a better name:
1. The resolution of a rotary encoder specifically would be the angular difference between detents, or number of state changes per rotation.
2. As implemented, setting the resolution from 2 to 4 will make the encoder operate on a lower resolution, which is the opposite of what may be expected.

zyumbik:
Yeah, I can see how that would be confusing. Any suggestions? All that comes to my mind is longer names such as `counts per detent` or `counts per cycle`. Maybe something like `cycle length`. I'll be happy to change the name to whatever you find reasonable. If you choose `divisor`, so be it. :)

xs5871:
Considering that we will integrate `rotaryio` at some point, I'd rather keep `divisor` for consistency.

zyumbik:
So rotaryio is coming? Maybe there is no need in these changes in this case? I worked on this because it was previously said in #347 that rotaryio is not wanted (for some reason) and it's also said in the encoder.py file. Is there a point in this PR if rotaryio is going to replace it?

In any case, no problem, I'll rename this setting to divisor.

xs5871:
There's already a PR #357 for including `rotaryio` as a peripheral driver. It's still WIP, the encoder module has become pretty convoluted and needs a proper refactor.
`rotaryio` isn't "unwanted", rather it has to be integrated with the existing encoder handling code. It'll be an addition, not a replacement.

Your contribution is nonetheless valuable. There are certain hardware configurations where `rotaryio` can't be used, that's why we have an independent implementation.

zyumbik:
I renamed resolution to `divisor`, formatted everything and tested once again. All works fine. Also sorry for force pushing, I'm very bad at git, hope it's ok. I just wanted to sync my changes with main branch. 

I have another addition to the module: support for multiple encoders on a shared pin. This way you can use three pins for 2 encoders instead of 4, e.g. A & B, B & C instead of  A & B, C & D. This of course would work with more pins (I wrote a detailed [guide](https://teletype.in/@gleb.sexy/153-encoders)). This only would work for encoders with 4 steps per detent because one encoder can stay at (1, 1) instead of (0, 0) thus preventing the correct reading of the other encoder. I also tested it and it works fine. Should I add it to this PR or add it separately? https://github.com/zyumbik/kmk_firmware/blob/multiple-encoders-support/kmk/modules/encoder.py

xs5871:
Cool. The force push shouldn't be an issue, git should resolve that automatically.

About the shared pin encoders: That can be a PR in it's own right. The current method of defining encoder pins and having the `EncoderHandler` initialize the hardware isn't very flexible and may not be ideal for that. We may think about providing the handler with a list of encoder modules instead (similar to scanners or rgb pixelbuffers, for example).

xs5871 (approved):

PR #485, opened 2022-06-21
[doc:rgb] fix link to neopixel lib

crazyiop:
The link is hardcoded to a commit from 2018-12-21 (version 3.3.5) while
we are currently at 6.3.3.

The rgb makes use of the fact that NeoPixel is a subclass of PixelBuf
but that was not true back then. Hence some people following the link in
the doc get a neopixel library incompatible with kmk.

### DIFF
--- a/docs/rgb.md
+++ b/docs/rgb.md
@@ -2,9 +2,7 @@
 Want your keyboard to shine? Add some lights!
 
 ## CircuitPython
-If not running KMKPython, this does require the NeoPixel library from Adafruit.
-This can be downloaded 
-[here](https://github.com/adafruit/Adafruit_CircuitPython_NeoPixel/blob/6e35cd2b40575a20e2904b096508325cef4a71d3/neopixel.py).
+This does require the [NeoPixel library from Adafruit](https://github.com/adafruit/Adafruit_CircuitPython_NeoPixel/blob/main/neopixel.py).
 It is part of the [Adafruit CircuitPython Bundle](https://github.com/adafruit/Adafruit_CircuitPython_Bundle).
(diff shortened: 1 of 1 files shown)


docs/rgb.md
```diff
@@ -3,8 +3,8 @@ Want your keyboard to shine? Add some lights!
 
 ## CircuitPython
 If not running KMKPython, this does require the NeoPixel library from Adafruit.
-This can be downloaded 
-[here](https://github.com/adafruit/Adafruit_CircuitPython_NeoPixel/blob/6e35cd2b40575a20e2904b096508325cef4a71d3/neopixel.py).
+This can be downloaded
+[here](https://github.com/adafruit/Adafruit_CircuitPython_NeoPixel/blob/main/neopixel.py)
```
xs5871, line 7:
```suggestion
This does require the [NeoPixel library from Adafruit](https://github.com/adafruit/Adafruit_CircuitPython_NeoPixel/blob/main/neopixel.py).
```

KMKPython is abandoned + improve bad practice "[here]" link

xs5871 (changes requested):

xs5871:
Sorry for the unclear request: KMKPython is abandoned; please also remove the mention from the doc, since we're already revising it. I couldn't mark that line in the review, but I should have stated that more clearly.

crazyiop:
Ha sorry, its me who didn't get it. Fixed

But why couldn't you open a thread on that line ? It's not possible to comment on unchanged line in github ? (It's the first project I really use github PR... I'm only familiar with gitlab where that is possible)

xs5871:
If it is possible, I couldn't figure it out.

xs5871 (approved):

PR #491, opened 2022-06-25
Set brightness to 100 when the board powers on

nogg3r5:
There didn't seem to be a way to do this, unless I missed something somewhere?

### DIFF
--- a/docs/led.md
+++ b/docs/led.md
@@ -45,6 +45,7 @@ All of these values can be set by default for when the keyboard boots.
 from kmk.extensions.led import AnimationModes
 led_ext = LED(
     led_pin=led_pin,
+    brightness=100,
     brightness_step=5,
     brightness_limit=100,
     breathe_center=1.5,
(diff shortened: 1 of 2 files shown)


klardotsh (commented):
I'm wondering if 100% is actually the best UX here - in the event the user has blinding LEDs, maybe 50% would be a more sane default?

I don't hold strong opinions on this one, particularly.

nogg3r5:
You're probably right here, I was testing with one, very dull LED.

docs/led.md
```diff
@@ -45,6 +45,7 @@ All of these values can be set by default for when the keyboard boots.
 from kmk.extensions.led import AnimationModes
 led_ext = LED(
     led_pin=led_pin,
+    brightness=100,
```
kdb424, line 48: Why do the docs say something, but then do something different? This number changes

kdb424 (changes requested):

xs5871:
> I'm wondering if 100% is actually the best UX here - in the event the user has blinding LEDs, maybe 50% would be a more sane default?
>

Maybe not the _best_ UX in case of aircraft landing lights, but the _expected_ one. To my knowledge, every PWM, PixelBuffer, etc. API initializes to 100% brightness, unless explicitly specified otherwise.

kdb424 (approved):

PR #492, opened 2022-06-25
Update assuming #491 gets approved

nogg3r5:
Add brightness initial value to document as per #491


klardotsh (changes requested):
go ahead and combine this into the other PR; they are the same logical unit of work

nogg3r5:
Changes merged into #491

PR #493, opened 2022-06-25
Update Officially_Supported_Microcontrollers.md

Jawfish:
Added the RP2040 variant of the SparkFun Pro Micro and the Adafruit KB2040, as well as renamed the "RP2040" entry to the Raspberry Pi Pico, which is what it was specifically referring to, and clarified its downside. I also provided software pin name conversion between the popular ATmega32U4-based Pro Micro et al. and the two boards I added, using the pin names as they appear to CircuitPython. I have personally used these two boards to run keyboards with KMK as drop-in replacements for the Pro Micro.


Jawfish:
Accidentally deleted my branch which seems to have automatically closed this.

PR #494, opened 2022-06-27
Add RapidFire module

Jawfish:
Implements a module that lets users hold a key to repeatedly send a tap.

### DIFF
--- a/docs/rapidfire.md
+++ b/docs/rapidfire.md
@@ -7,6 +7,7 @@ Some instances where this may be useful are:
 - MMOs and other games where you are encouraged to repeatedly spam a key
 - More responsive volume up and volume down
 - Faster cursor key navigation
+- Combine with the [Mouse Keys](https://github.com/KMKfw/kmk_firmware/blob/master/docs/mouse_keys.md) module to create rapid-fire mouse clicks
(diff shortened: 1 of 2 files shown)


xs5871 (commented):
Overall pretty neat. Got some minor suggestions.

xs5871:
Sorry that I didn't come up with this earlier: Did you model this after some "reference" implementation?
I'm asking because I keep stumbling over the `repeat` option thinking it refers to a maximum number of repeats, which of course doesn't make sense. In my opinion something like `interval`, `repeat_delay` or similar would be more obvious. I'm also considering here that humans generally tend to not read the entire documentation.
Similar argument with `wait`: in every other "delayed action" module we've used the term `timeout`. That's more for consistency than clarity.

Jawfish:
Decided to rename `randomize` to `enable_interval_randomization` with the justification that the improvement in clarity is worth the tradeoff in verbosity since in most cases users would be using the default value anyway.

kdb424 (approved):
Looks pretty readable to me. I need to try this at some point, but really useful feature to have.

xs5871:
Agreed. I'm happy with this.

PR #496, opened 2022-06-30
Implement Text Replacement module

Jawfish:
Adds a module to bring [espanso](https://github.com/federico-terzi/espanso)-like functionality to the keyboard (substitute a string of typed characters with another string). I've been using this and haven't run into any issues, but I imagine there will be untested edge cases. I would like to write tests for this once I familiarize myself a bit more with how to do that properly for KMK.

I feel this differs adequately from [combos](https://github.com/KMKfw/kmk_firmware/blob/master/docs/combos.md) in that it allows for live typing. I found trying to use the sequences in the combos module for text replacement while typing to be too disruptive, especially for autocorrection, as it suppresses the sending of key presses until it has a match or times out. 

Another problem with using combos for this is that if you have a sequence you want to match, say, "abc", and another that has the same character as that sequence, say, "bed", typing "abed" won't allow the "bed" sequence to be activated - you have to either wait for the "abc" match to time out after typing "a", or type "abbed".

I don't see these as issues with combos - I believe that module works as intended, but that the intent was not replacing text in this manner.


kdb424:
At least with the prefix key, that looks a lot like "leader mode" which is a Vim function inspired thing that got sucked into combos. I generally like this functionality, but wonder if leader mode may solve a similar problem. I see that you may have other edge cases, and believe that this functionality is valid. Main "concern" I have initially is "should this be part of combos, or it's own thing?". If you feel that it's infeasible to add to combos, especially knowing about the leader key (your text expansion example), then I'll give it a whole review as I like the concept. You bring up points that have me believe it's worth a review, but wanted to make sure it should be distinct.

xs5871:
Well, like both of you said, a good chunk of this functionality can be done with combos `sequences`. The edge cases where it doesn't work is something we can easily integrate into combos, or extend it to a third matcher. The advantage would be, that we obviously only have to vet one module with unit tests, and we reduce the amount of confusion around the order of modules with sideeffects...
Downside is, and I mentioned that in the last combo commit, the module should be refactored first, or it's going to be a literal headache to extend.

The big difference I see is that the proposed module doesn't replace text in flight. It waits until a match, then fires backspaces followed by the replacement string. Not sure if I particularly like it, but that's necessary for responsive typing and longer phrases.
Now I'm really interested (algorithmically) if we could stick that in combos.

I skimmed through the code and can already see possible inconsistencies or unexpected behaviour. If it's going in, it definitely needs unit tests.

Jawfish:
I originally used combos for text replacement, but, as mentioned, found it disruptive enough while typing that I took the time to write this for my own personal use. It becomes especially noticeable if you have a sizeable dictionary of words to auto-correct (i.e., words that aren't prefixed with a symbol) - much of what you type will be delayed. 

Another consideration is cases where delays are not only distracting, but detrimental. Imagine a user wants to play a game in which they chat and they wish to retain the auto-correct functionality without adding a delay to keys they need to press to play the game.

Perhaps I focused too much on the text-expansion functionality, as the greatest non-overlapping use case compared to combos is auto-correction, but I do feel it has value there as well compared to combos, such as not getting lost when matching longer sequences, which could be seen as an accessibility feature.

As for whether it should go into combos - personally, I feel the use case combos is trying to solve with its sequences is vim-like leader key functionality, and, even though text substitution is possible with that, that's not its explicit purpose and is outside of the scope of that module. I suppose it would ultimately be up to you, as the author, if you think it should cover my use cases as well.

I do agree it needs unit tests, however, and I'll start looking into adding them.

xs5871:
FYI: the underscore thing is a bug (see #499).

Jawfish:
I'm currently writing unit tests for this, but have run into some issues. When my module is enabled in the test keyboard, the following code fails:
```python
test_dictionary = {'a': 'b'},
keyboard = KeyboardTest(
    [TextReplacement(test_dictionary)],
    [[KC.A,]]
)
keyboard.test(
    [(0, True), (0, False),
    [{KC.A}, {}], # this fails
)
```
Even though when pressing the key personally using the exact code I'm testing, I can see that it works. Testing for `KC.BACKSPACE` also fails, but testing for the replacement text passes:
```python
test_dictionary = {'a': 'b'},
keyboard = KeyboardTest(
    [TextReplacement(test_dictionary)],
    [[KC.A,]]
)
keyboard.test(
        test_dictionary = {
    [(0, True), (0, False),
    [{KC.B}, {}], # this passes
)
```

It seems to be related to something that happens after I change the state in my code once a match is found, at which point everything is done in `before_hid_send`. Any insight?

Jawfish:
I suppose I don't really need to test for the preceding keypresses, as that's covered by `test_kmk_keyboard.py`, but the issue remains when trying to check for `KC.BACKSPACE` being sent, which I need to do to make sure the correct number of backspaces are sent.

xs5871:
There's a bug in our unit test code... both of your examples _should_ fail.
I tried to avoid touching that horrendous code, guess I don't have an excuse anymore.

Jawfish:
I've implemented tests that cover the following:

- The user dictionary is converted into the correct sequence of keycodes for every valid symbol, letter, and number, including whether they require shift or not
- An informative error is thrown when the user dictionary contains an invalid character (that is to say, a character that doesn't have an associated keycode)
- `Character`, `Phrase`, and `Rule` classes get constructed as expected for every valid symbol, letter, and number, shifted and unshifted
- The `Phrase` class indexes and resets correctly, keeping track of current potential matches
- The `Rule` class manages `to_substitute` and `substitution` phrases properly
  
I'll wait for #504 to be merged before writing tests for the following:

- The expected replacement keycodes are sent by the keyboard when the user types a corresponding match
- `KC.BACKSPACE` is sent the expected number of times after a valid match is found
- Keys for characters that should be shifted are wrapped in `KC.LSHIFT()` when being sent

kdb424 (approved):
I like this. I'll allow others to code review as they are more qualified than I, but seems like a great feature.

kmk/modules/text_replacement.py
```diff
@@ -0,0 +1,202 @@
...
+
+class Phrase:
+    '''Manages a collection of characters and keeps an index of them so that potential matches can be tracked'''
+
+    def __init__(self, string: str) -> None:
+        self._characters: list[Character] = []
+        self._index: int = 0
+        for char in string:
+            try:
+                key_code = getattr(KC, char.upper())
```
xs5871, line 45:
```suggestion
                key_code = KC[char]
```
`key.upper()` is also applied in `KC.__get_item__`

kmk/modules/text_replacement.py
```diff
@@ -0,0 +1,202 @@
...
+        if self._state == State.DELETING:
+            # send backspace taps equivalent to the length of the phrase to be substituted
+            to_substitute: Phrase = self._matched_rule.to_substitute  # type: ignore
+            to_substitute.next_character()
+            if not to_substitute.index_at_end():
+                keyboard.tap_key(KC.BSPC)
+            else:
+                self._state = State.SENDING
+                # force-release modifiers so sending the replacement text doesn't interact with them
+                # it should not be possible for any modifiers other than shift to be held upon rule activation
```
xs5871, line 167: should not? Make a unit test.

tests/test_text_replacement.py
```diff
@@ -0,0 +1,216 @@
...
+        self.symbols = '`-=[]\\;\',./~!@#$%^&*()_+{}|:\"<>?'
+        self.everything = ALL_NUMBERS + ALL_ALPHAS + ALL_ALPHAS.lower() + self.symbols
+        self.test_dictionary = {'aa': 'b', 'b': 'aa', '!': '@', 'dccc': 'dcbb'}
+        self.text_replacement = TextReplacement(self.test_dictionary)
+        self.keyboard = KeyboardTest(
+            [self.text_replacement],
+            [
+                [KC.A, KC.B, KC.N1, KC.LSHIFT, KC.LCTRL, KC.C, KC.D],
+            ],
+            debug_enabled=True,
```
xs5871, line 19:
```suggestion
            debug_enabled=False,
```
default for debugging is "off"

tests/test_text_replacement.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+
+        return super().setUp()
+
+    def test_keyboard_events_are_correct(self):
```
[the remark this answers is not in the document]
Jawfish: I've implemented all the tests you mentioned, but a particularly long test to test multiple backspaces only passes when I spam `self.keyboard._main_loop()` at the end of `def do_main_loop(self):` in `keyboard_test.py`, as mentioned in #506. Since the test does pass when I do that, I could comment that particular test out and push my changes - how would you like me to proceed?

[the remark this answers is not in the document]
xs5871: Forgot to mention: Every test that relies on the keyboard doing things after the last key event (timeouts, etc...) need a short delay after the last key event:
```python
        self.keyboard.test(
            'backspace is only tapped as many times as necessary to delete the difference between the key and value',
￼            [ ...
￼                (5, True),
￼                (5, False),
                50,
￼            ], ...
)
```

Jawfish:
Renamed to String Substitution as suggested (that's actually what I originally intended to call it anyway, so no argument there), added the suggested test cases plus a couple others I thought of, and made the suggested syntax changes. 

I also improved guarding against non-shift modifiers by adding an `IGNORING` state which becomes active if any non-shift key that inherits `ModiferKey` is pressed.

`make test` should pass once #508 or similar is merged.

PR #500, opened 2022-07-02
implement arbitrary code execution (ACE) over data serial.

xs5871:
Every application needs a gaping hole in security, ready to be exploited.
I don't even remember how I got the idea, but in the past people were asking questions like "*how can I tell my keyboard `x` from my host OS*".
Well, here's an API that allows arbitrary code execution over the built-in data serial (which is different from the REPL/console).
This is as much a proof-of-concept as it is probably a terrible idea.

How to use:
1. enable the data serial in `boot.py`:
```python
import usb_cdc
usb_cdc.enable(data=True)
```
2. load module (obvious)
3. assuming serial on `/dev/ttyACM1` and _explicitly_ set to send `\n` as line break (this is not always the default):
```bash
$ echo "keyboard.active_layers" > /dev/ttyACM1
$ cat /dev/ttyACM1
[0]
$ echo "exec('from kmk.keys import KC; keyboard.tap_key(KC.Y)')" > /dev/ttyACM1
$ y
```

It's in draft because still poc, no docs, not sure if we even want this...

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -29,7 +29,7 @@
 - [Mouse keys](mouse_keys.md): Adds mouse keycodes
 - [OneShot](oneshot.md): Adds support for oneshot/sticky keys.
 - [Power](power.md): Power saving features. This is mostly useful when on battery power.
-- [SerialACE](serialace.md): Arbitrary Code Execution over the data serial.
+- [SerialACE](serialace.md): [DANGER - _see module README_] Arbitrary Code Execution over the data serial.
 - [Split](split_keyboards.md): Keyboards split in two. Seems ergonomic!
 - [TapDance](tapdance.md): Different key actions depending on how often it is pressed.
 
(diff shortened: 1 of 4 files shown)


kdb424:
I do believe that this has a place, but should have MASSIVE warnings on it. I'm thinking feeding host OS system to the OLED's or something of that sort. I both hate this, but also like what you could do with it.

kdb424 (approved):
I'm in favor of this. I'll let @klardotsh reject if he sees a reason to, but otherwise, this looks like a horrible, yet great feature for those that want to do that. Toys for the sake of toys make sense in some cases, and this hits that.

klardotsh (approved):
> Every application needs a gaping hole in security, ready to be exploited.

lmfao

hells yeah

Funny enough, just this past week at DEFCON in the Boardsource booth (arguably the most apt place for a module like this to get turned into a horror show) I was telling someone "I don't know, but you could probably hack some shit up over the serial bus to do that" when asked how to show host system data on an OLED attached to a NiceNano. Hope that guy finds this and gets a giggle.

Happy self-pwning, everyone!

klardotsh (approved):
No further commentary on my end. Clear for merge.

PR #501, opened 2022-07-02
Explicitly define CircuitPython 7.0 or higher as the minimum supported version

Jawfish:
As per the conversation in #205, I've specified 7.0 as the explicit minimum version for KMK in preparation for static typing and made the requirement more apparent by placing it on the main readme.


kdb424 (approved):
I'm ok with this

PR #503, opened 2022-07-03
[status_indicator] add hardware agnostic extension for layer indication

crazyiop:
After a few hacky try I got a versatile led layer/capslock indicator for myself. As I've seen a few time asked on the discord how to do such layer indicator, I cleaned it up so that it is hardware independent. 

I am myself using it now (the example with a twist for the each side of my split keyboard).

### DIFF
--- a/docs/status_indicator.md
+++ b/docs/status_indicator.md
@@ -1,0 +1,80 @@
+# Status Indicator
+This extension allow you to customize reaction to layer and hid changes (capslock/ver-num/..). It provides only the kmk internal handling and makes no hardware assumption. This part is up to you so that this extension will be compatible with leds, neopixels, lcd or even servo !
+
+This extension is then for you if you want to act on your keyboard hardware on layer change. Other extensions can do that but assume certain hardware. Check if you are in one of those case:
+- [Status LED](extension_statusled.md) for several, individual leds (a gpio per led).
+- [Peg Oled Display](peg_oled_display.md) for oled display.
+
+## How to use
+
+You need to subclass `Indicator` and add your hardware setup in `__init__()` and customize the `update()` method.
+
+Here is a simple example using a neopixel strip to set a color for each layers, and one for capslock (caplock being higher priority):
+
+```python
+import board
+from kmk.extensions.status_indicator import Indicator
+from kmk.kmk_keyboard import KMKKeyboard
+from neopixel import NeoPixel
+
+
+class MyIndicator(Indicator):
+    # base layer -> off, layer 1 -> blue, layer 2 -> red
+    LAYERS_COLORS = [(0, 0, 0), (0, 255, 0), (255, 0, 0)]
+    CAPS_COLOR = (128, 128, 128)  # white
+
+    def __init__(self, pixel_pin, num_pixels):
+        super().__init__()  # Needed !
+        self.np = NeoPixel(pixel_pin, num_pixels)
+
+    def update(self):
+        if self.get_caps_lock():
+            # caps lock on:
+            self.np.fill(self.CAPS_COLOR)
+        else:
+            self.np.fill(self.LAYERS_COLORS[self.layers[0]])
+
+    def on_powersave_enable(self, sandbox):
+        # turn off neopixel
+        self.np.fill((0, 0, 0))
+
+    def on_powersave_disable(self, sandbox):
+        # restore neopixel
+        self.update()
+
+
+keyboard = KMKKeyboard()
+
+my_indicator = MyIndicator(board.D4, 6)
+keyboard.extensions.append(my_indicator)
+```
+
+## Available options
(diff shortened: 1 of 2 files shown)


xs5871:
Wouldn't it make sense to show users how to connect the indicator with one of the existing led extensions, instead of introducing a fourth one?

xs5871:
So. I looked at it for a while and think something similar happend when I tried to get my indicator into the wild.
The actual `Indicator` extension has barely any usefull code in it, arguably not enough to warrant a seperate module. All the interesting stuff is happening in the doc: More or less a "how to build your own indicator". This personalized code doesn't make sense in a one-size-fits-all module, resulting in a very sparse `Indicator` extension, coupled with a somewhat complicated child class in the doc.

What do you think about making this a pure HowTo with two examples, maybe one inherited from `Status_indicator` hooking into the PWM LED extension, and one a custom animation for addressable RGB. (I could contribute the second example.)

crazyiop:
@xs5871 To answer your first comment: 

No one of the current module can do what I want, I don't know much about the other people on discord but they ask about custom implementation right away, letting me think that they maybe already considered the existing module before. Some very common hardware/feature should definitely have their own module but they cannot realistically take into account all the hardware/custom features people will want, hence the need for something basic. The fact that this extension is something  for everyone to build upon makes it rather different for the others that makes assumption and hard-code choices. Calling it just 'a fourth one' is misleading.

My personal example that wont realistically fit the existing extension (afaik): I have a split with neopixels in each side. I want them to be one color for capslock, one for every layer !=0. If both condition are met at once, one side will have the capslock color, the other the layer color. Also, in the end, I don't want to use the NeoPixel class, but a custom subclass HSVPixel that can take an int (hue) as assignment in addition to the rgb tuple (and using [the corrected rainbow from fastled](https://github.com/FastLED/FastLED/wiki/FastLED-HSV-Colors) instead of the plain classic spectrum like what is used in the `hsv_to_rgb()` of the rgb extension...)


----
second comment:

After thinking about it, I think that what drives me to do this is actually a missing common base for the different indicator feature. I don't see the problem in having not much code if that small part is meant to be reused/copied/reinvented (sometimes more poorly) numerous times...

Imho:
- it is useful to provide a clean base for people wanting to do very custom thing.
- the existing modules could even inherit from it, so they will be more standardized.

docs/status_indicator.md
```diff
@@ -0,0 +1,80 @@
...
+
+class MyIndicator(Indicator):
+    def __init__(self, ...):
+        ...
+
+        self.brightness_level = 0
+        make_key(names=('LED_LVL',), on_press=self._led_level)
+
+    def _led_level(self)
+        self.brightness_level = (self.brighteness_level + 1) % 4
```
xs5871, line 75: Similar to the above. We already have a PWM LED extension that provides special control keys.

kmk/extensions/status_indicator.py
```diff
@@ -0,0 +1,21 @@
...
+class Indicator(LockStatus):
+    def __init__(self):
+        super().__init__()
+        self.layers = []
+
+    def after_hid_send(self, sandbox):
+        backup = self.report
+        super().after_hid_send(sandbox)
+        if self.report != backup:
+            self.update()
```
xs5871, line 13: Not sure why this is necessary. You could just as easily check for a change in `status_indicator.get_caps_lock()` directly, or any other metric for that matter.

kmk/extensions/status_indicator.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def after_matrix_scan(self, sandbox):
+        if self.layers != sandbox.active_layers:
+            self.layers = sandbox.active_layers.copy()
+            self.update()
```
xs5871, line 18: There's also no need to split the detection of states into two hooks, you can put a list of checks into the same function, like `if any(...): update()`.

xs5871 (commented):
There's also the `statusled` extension which tries to solve similar problem, and also doesn't reuse any of the existing drivers. There's some potential for consolidation.

hendrix04:
So, I agree that this code is a bit TOO generic.

The real problem here is due to a few things...
1) The statusLED extension is WAY too generically named
2) The statusLED extension is WAY too opinionated
3) We can't really do much about either without potentially breaking a lot of people's code.

Can we deprecate the current statusLED extension and create a v2? If so, what would v2 look like in terms of instantiating? Do we just create an extension called "statusLEDv2"? Do we some how make an optional parameter as part of the constructor that triggers v2?

Another alternative would be to leave statusLED alone and add the ability to assign callback functions to lock_status for each status type... The more that I think about it, the more that makes sense honestly. That and some better documentation as pointed out above.

xs5871:
This discussion is addressed and resolved in #553.

PR #504, opened 2022-07-03
Fix unittest

xs5871:
Pretty much a complete refactor of the keyboard emulation unittest + fixes to get everything working again.
Current version was impossible to read, had terrible debug output, and most importantly didn't do bjective comparisons of generated and desired keycode reports.
KC has an `__iter__` now, so we can do reverse lookups.

### DIFF
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -60,6 +60,9 @@ def maybe_make_consumer_key(candidate, code, names):
 class KeyAttrDict:
     __cache = {}
 
+    def __iter__(self):
+        return self.__cache.__iter__()
+
     def __setitem__(self, key, value):
         if DEBUG_OUTPUT:
             print(f'__setitem__ {key}, {value}')
--- a/kmk/modules/combos.py
+++ b/kmk/modules/combos.py
@@ -232,8 +232,8 @@ class Combos(Module):
 
         if not combo._remaining:
             self.activate(keyboard, combo)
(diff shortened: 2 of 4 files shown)


xs5871:
Still don't understand why my local linter is sometimes different than the github one.

xs5871:
The unittests run perfectly fine in my local environment... I don't understand this.

xs5871:
Couldn't reproduce the issue locally reliably (had it once in the first of about 100 runs). Maybe some random race condition...
Not a great fix, but it seems to do the job. I welcome any further insights at this point.

Jawfish:
Experiencing some strange behavior with this:
On Windows (Python 3.10.5), `python -m unittest` has 6 fails
On Ubuntu (Python 3.8.10), it has 1 fail (but `make test` passes)
On MacOS (Python 3.9.10), it has no fails
This happens consistently. Tested on a fresh clone of the repo.

The failed tests are:
`test_chord`, `test_sequence`, `test_holdtap_chain`, `test_oneshot`, `test_modtap` and `test_normal_key` on Windows, only `test_chord` in Ubuntu.

Seems to be some kind of timing issue.

LukeDRussell:
I get failures on Windows 11 as well:
test_chord, test_sequence, test_oneshot, test_modtap, test_normal_key
All fail with `AssertionError: reports don't match up`

PR #507, opened 2022-07-09
Adds an alias for the LEADER keycode

LukeDRussell:


kmk/modules/combos.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, combos=[]):
         self._key_buffer = []
 
         make_key(
-            names=('LEADER',),
+            names=('LEADER','LDR'),
```
kdb424, line 74: Forgot a space. Linter be angry.

kdb424 (changes requested):

kdb424 (approved):

Issue #506, opened 2022-07-06
Unit tests fail on certain platforms


Jawfish:
After #504 was merged, running `python -m unittest` on an untouched clone of the repo has failures on Windows and Ubuntu, but not MacOS.

Windows 11:
```
FAIL: test_chord (tests.test_combos.TestCombo)
AssertionError: reports don't match up
FAIL: test_sequence (tests.test_combos.TestCombo)
AssertionError: reports don't match up
FAIL: test_holdtap_chain (tests.test_hold_tap.TestHoldTap)
AssertionError: reports don't match up
FAIL: test_oneshot (tests.test_hold_tap.TestHoldTap)
AssertionError: reports don't match up
FAIL: test_modtap (tests.test_tapdance.TestTapDance)
AssertionError: reports don't match up
FAIL: test_normal_key (tests.test_tapdance.TestTapDance)
AssertionError: reports don't match up
```

Ubuntu 20.04:
```
FAIL: test_chord (tests.test_combos.TestCombo)
AssertionError: reports don't match up
```

However, `make test` passes.

xs5871:
Now it fails in the runner again.
Can not reproduce with python3.10.
Can reproduce with python3.8.

Jawfish:
I get failures the following cases:
Python 3.10.5 on Windows 11
Python 3.8.10 on Ubuntu 20.04, same machine as Windows 11
Python 3.10.5, on Arch, different machine
I don't get any failures on Python 3.9.10 on macOS.

By removing the sleep code from `keyboard_test.py`, I can pass consistently* on Windows and Arch:
Before
```python
def do_main_loop(self):
    self.keyboard._main_loop()
    time.sleep(0.001)
    self.keyboard._main_loop()
    time.sleep(0.001)
```
After
```python
def do_main_loop(self):
    self.keyboard._main_loop()
```

*I got a fail for `test_chord` the first time running `python -m unittest` on each platform, but not thereafter. It continues to fail on Ubuntu 20.04.

I'm not familiar enough with this class to determine if that's an appropriate solution, but it seems the issue is related to how various machines/operating systems track time, and maybe some differences between how it's handled with different Python versions under the hood.

xs5871:
I put the sleep statement in to emulate KMKs maximum loop execution frequency.
I get the tests to pass if I paste a bunch of `self.keyboard._main_loop()` at the end of `do_main_loop`.

kdb424:
[...]

PR #508, opened 2022-07-09
unit test fail may be caused by a stuck combo

xs5871:
Let's see if this still fails.

### DIFF
--- a/kmk/modules/combos.py
+++ b/kmk/modules/combos.py
@@ -193,6 +193,7 @@ class Combos(Module):
                     else:
                         combo._remaining.insert(0, key)
                         self._matching.append(combo)
+                    self.reset(keyboard)
 
                 elif not combo._remaining:
                     continue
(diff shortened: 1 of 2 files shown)


xs5871:
may resolve #506
@Jawfish could you check with your abundance of different operating systems?

Jawfish:
It passes on Arch/Ubuntu/macOS, but still fails on Windows. There is one less failure on Windows - `test_holdtap_chain` passes now.

xs5871:
Hm. Not great, but it's something.
I think I remember the original author of the unit test mentioning the random loop of `keyboard_go` as "necessary to make it pass", but that's hardly a satisfactory explanation for what's going on.
Can we merge this as a step in the right direction, but keep #506 open?

Jawfish (approved):
I don't have a problem with it. I imagine most contributors will be running something Unix-based anyway (it passes on WSL, too)

kdb424 (approved):
I vote that we merge this. It works on "all" platforms if you count WSL, and honestly, that seems like a valid shim to me. If someone wants to offer a different fix later, awesome, but this is better than nothing, and is technically always available, much the same as "docker only runs on linux" even though people use it on other platforms *sorta technically they don't but whatever*.

PR #520, opened 2022-07-19
Add keycode for software reload

Jawfish:
Adds `KC.RELOAD` for software reload which makes use of [`supervisor.reload()`](https://docs.circuitpython.org/en/latest/shared-bindings/supervisor/index.html#supervisor.reload). This differs from the current `KC.RESET` because it only reloads the software, preserving any active serial connections and doesn't clear things like tracebacks. Useful for anyone that [turns off auto-reloading](https://docs.circuitpython.org/en/latest/shared-bindings/supervisor/index.html#supervisor.disable_autoreload).

### DIFF
--- a/docs/keycodes.md
+++ b/docs/keycodes.md
@@ -186,6 +186,7 @@
 |Key                    |Description                                                          |
 |-----------------------|---------------------------------------------------------------------|
 |`KC.RESET`             |Restarts the keyboard                                                |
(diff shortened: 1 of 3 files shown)


kdb424 (approved):
I really like this addition. That's insanely useful, and should honestly be the default way to do dev work because that's so much nicer of a UX

PR #524, opened 2022-07-20
Add 'BSPACE' keycode for backspace

skarrmann:
The KMK documentation lists keycode name `BSPACE` as a valid name for the backspace key: http://kmkfw.io/docs/keycodes#basic-keys

However, using `KC.BSPACE` in one's keymap throws the following error:

```
ValueError: Invalid key: BSPACE
```

To resolve this issue, this PR simply adds `BSPACE` as a valid alias for the backspace key.


kdb424 (approved):
Simple oversight. Seems reasonable. Thanks!

PR #526, opened 2022-07-21
Add reverse scrolling mode to the Pimoroni Trackball module

slabua:
I added the good old reverse scrolling mode.

Scrolling parameter is optional, defaults to natural scrolling direction.

### DIFF
--- a/docs/pimoroni_trackball.md
+++ b/docs/pimoroni_trackball.md
@@ -32,7 +32,7 @@ trackball = Trackball(i2c, mode=TrackballMode.MOUSE_MODE, handlers=[
     KeyHandler(KC.UP, KC.RIGHT, KC.DOWN, KC.LEFT, KC.ENTER), 
     # on layer 1 and above use the default pointing behavior
     PointingHandler(),
-    # use scrolling=ScrollDirection.NATURAL (default) or REVERSE to change the scrolling direction
+    # use ScrollDirection.NATURAL (default) or REVERSE to change the scrolling direction
     ScrollHandler(scrolling=ScrollDirection.NATURAL)
 ])
 
--- a/docs/pimoroni_trackball.md
+++ b/docs/pimoroni_trackball.md
@@ -33,7 +33,7 @@ trackball = Trackball(i2c, mode=TrackballMode.MOUSE_MODE, handlers=[
     # on layer 1 and above use the default pointing behavior
     PointingHandler(),
     # use ScrollDirection.NATURAL (default) or REVERSE to change the scrolling direction
-    ScrollHandler(scrolling=ScrollDirection.NATURAL)
+    ScrollHandler(scroll_direction=ScrollDirection.NATURAL)
 ])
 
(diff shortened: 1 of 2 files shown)


xs5871 (changes requested):
Looks sensible, except we prefer enum-like objects instead of magic strings, something like this:
```python
class ScrollDirection:
    NATURAL = const(0)
    REVERSE = const(1)

# ...
if self.scrolling == ScrollDirection.REVERSE:
```

slabua:
> Looks sensible, except we prefer enum-like objects instead of magic strings, something like this:
> 
> ```python
> class ScrollDirection:
>     NATURAL = const(0)
>     REVERSE = const(1)
> 
> # ...
> if self.scrolling == ScrollDirection.REVERSE:
> ```

Thanks, it looks much better indeed!
Code updated accordingly.

klardotsh (commented):
Are there clearer terms than "natural" and "reverse" here? I take "Natural Scrolling" to mean what it does in, say, MacOS or ChromeOS (in that gesturing the tracking device downwards will move a document upwards), and "reverse" means nothing useful to me.

ChromeOS calls "down is up" scrolling "Australian" scrolling, which is a clever tongue-in-cheek (if Northern Hemisphere centric) way of disambiguating these things.

Which makes me think - maybe calling the constants "DOWN_IS_DOWN" and "DOWN_IS_UP" is clearer? Idk.

klardotsh:
I also approved and ran tests here, which then failed the linter. `make fix-lint` is in order

slabua:
> Are there clearer terms than "natural" and "reverse" here?
> [...]

I think those are the universally recognised terms for either type of scrolling.

> Which makes me think - maybe calling the constants "DOWN_IS_DOWN" and "DOWN_IS_UP" is clearer? Idk.

Maybe better to specify it explicitly, yes.

slabua:
> I also approved and ran tests here, which then failed the linter. `make fix-lint` is in order

hm 🤔

slabua:
Should I update the code with the clearer definitions of "DOWN_IS_DOWN" and "DOWN_IS_UP"?

xs5871:
My 2 cents and not at all a strong opinion: I'd prefer the natural/reverse nomenclature. If "natural" is the default, then "reverse" is obvious.

On another note: what's your opinion on renaming the `scrolling` parameter to `scrolling_direction`?

slabua:
> My 2 cents and not at all a strong opinion: I'd prefer the natural/reverse nomenclature. If "natural" is the default, then "reverse" is obvious.

Alrighty

> On another note: what's your opinion on renaming the `scrolling` parameter to `scrolling_direction`?

I agree on `scrolling_direction`, it is semantically more correct.

slabua:
> I agree on `scrolling_direction`, it is semantically more correct.

 Perhaps `scroll_direction` then?

slabua:
Code has been updated accordingly

xs5871 (approved):

PR #529, opened 2022-07-22
implement global debug interface.

xs5871:
A "universal" debug interface that:
- can be used from anywhere in the code base (no access to `KMKKeyboard.debug_enabled` required)
- shows where the message is coming from (named debug instances)
- has timestamps... I'm amazed we had to go without one for so long
- could easily be extended to optionally show other metrics, like stack size or differential memory at ever message

Originally I planned to have the is-debug-enabled-check also in the debug class. Then I noticed that we have f-strings everywhere now, and those would be evaluated before the check happens, wasting precious CPU cycles.
I quite like the syntax of f-strings, but is it worth keeping and carrying an `if debug.enabled:` through the code?


kdb424 (approved):
This is amazing! So much nicer than what we had.

PR #530, opened 2022-07-24
Update mouse_keys.py

kbjunky:
Added check for USB enabled half in case of a split keyboard.
Cleaning of hid_pending state after sending.


kdb424 (approved):
Seems like it makes sense.

kbjunky:
Indeed it does. I've stumbled upon this bug with my recent split build. Issue is mouse keys are sent directly via HID in KMK_Keyboard class and on the half that has no USB connection this would kill it. With this patch USB enabled half will send those keys directly and the one without USB with relay on split functionality in KMK.

Issue #528, opened 2022-07-21
Chord doesn't always respect timeout value


Jawfish:
The chord feature of the combos module doesn't use a user-set timeout until that timeout has passed since the last keystroke.
Example:
```python
keyboard.modules = [
    combos,
]
combos.combos = [Chord((KC['1'], KC['2']), KC['3'], timeout=5000)]
keyboard.keymap = [KC['1'], KC['2'], KC['3'], KC['B']]
```
Typing 1 or 2 within the first 5000 milliseconds of the previous non-chord keypress will send 1 or 2 rather than waiting up to 5000ms for the next keypress. The chord still works if you press it within the default 50ms timeout. After 5000ms has passed since the last keystroke, it works as expected.

Current behavior:
1. B is pressed
2. B is released
3. The following occurs within 5000ms of B being released:
  a. 1 is pressed
  b. 1 is sent by the keyboard to the computer
  c. 2 is pressed
  d 2 is sent by the keyboard to the computer
  e. 2 is released
  f. 1 is released
11. 5000ms passes
12. The following occurs within 5000ms
  a. 1 is pressed
  b. 2 is pressed
  c. 3 is sent by the keyboard to the computer

Expected behavior:
1. B is pressed
2. B is released
3. The following occurs within 5000ms of B being released:
  a. 1 is pressed
  b. 2 is pressed
  c. 3 is sent by the keyboard to the computer

xs5871:
That's "intentional". You pressed B+1+2 within the combo timeout, and that key combination doesn't match the chord 1+2.
Is what you expect a rolling matching window?

Jawfish:
B is released before 1 is pressed in my example.

Jawfish:
I updated my comment to make it more clear, previously I just said 'tap' to mean a press-and-release action.

Jawfish:
Here's a test I wrote for this that fails:

```python
# the config
combos.combos = [
    Chord((KC.F, KC.G), KC.Z),
]
self.keyboard = KeyboardTest(
    [combos, layers],
    [
        [KC.A, KC.B, KC.C, KC.D, KC.E, KCMO, KC.F, KC.G],
        [KC.N1, KC.N2, KC.N3, KC.N4, KC.N5, KC.LEADER],
    ],
)

# the test
keyboard.test(
    'match: Other pressed before combo',
    [
        (1, True),
        (1, False),
        (7, True),
        (6, True),
        (6, False),
        (7, False),
        t_after,
[...]

PR #531, opened 2022-07-24
Fix key pressed before combo making combo wait for timeout

Jawfish:
Fix for #528. 

Resets combos on key-up if there isn't a current match.

Added 4 new tests for this case, too, with different delays in key press/release and checking that it works whether the other key is a combo key or "inert".

I also added some type hints to help me with debugging and left them in, hope that's alright since it'll need to be done anyway for #205.

### DIFF
--- a/kmk/modules/combos.py
+++ b/kmk/modules/combos.py
@@ -175,6 +175,7 @@ class Combos(Module):
 
                 key = combo.result
                 break
+
         else:
             # Non-active but matching combos can either activate on key release
             # if they're the only match, or "un-match" the released key but stay
@@ -229,6 +230,7 @@ class Combos(Module):
         # Reset on non-combo key up
         if not self._matching:
             self.reset(keyboard)
+
         return key
 
     def on_timeout(self, keyboard, combo):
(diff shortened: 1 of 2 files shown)


kmk/modules/combos.py
```diff
@@ -223,6 +226,9 @@ def on_release(self, keyboard, key, int_coord):
                 self._key_buffer.append((int_coord, key, False))
                 key = None
 
+        # Reset on non-combo key up
+        if not self._matching:
+            self.reset(keyboard)
```
xs5871, line 231: This obviously only resets if _no_ combos are matching, and doesn't reset non-matching combos if there're potential matches left. I'm not 100% sure that's perfect, but it's at least a step foward.

kmk/modules/combos.py
```diff
@@ -171,7 +175,6 @@ def on_release(self, keyboard, key, int_coord):
 
                 key = combo.result
                 break
-
```
xs5871, line 174: Please leave the line break. Visual separation of functional blocks in the code aids readability.

kmk/modules/combos.py
xs5871, line 231:
```suggestion
            self.reset(keyboard)
            
```
add an empty line, see above.

xs5871 (approved):

xs5871:
Thanks!

Issue #495, opened 2022-06-27
Combos interfer with hold-tap (but only if Combos module is loaded after Holdtap module!)


ald0o:
## The issue

- load modules `HoldTap` and `Combos` in this order
- declare a hold-tap key with `hold_preferred=True`; `tap_interrupted` can be `True` or `False` (the bug will be triggered in both cases through different mechanisms). For instance: 
```python
HT = KC.MT(KC.ENT, KC.LSFT, prefer_hold=True)
```
- declare a chord using HT

Within both holdtap timeout and combo timeout, do: press HT, press A, release A, release HT:

Expected: the keyboard should send "A" (shifted A).
Happening instead: a "a" (unshifted) followed by a new line (or a new line followed by a "a").

## Proposed explanation

`process_key` method from the `Combos` module returns `None` for a key that belongs to a combo.
Consequence: in class `KMKKeyboard`, method `pre_process_key`, the `on_press` handler is not called for `HT`.
But this handler is supposed to register the fact that this holdtap key is being held (in array `key_states` in class `HoldTap`), so it does not happen (at least not until the combo is cancelled... which means too late!).

So, when we press `A`,  because `key_states` is empty (or at least does not contain HT), nothing particular happens in the `process_key` method of class `HoldTap`. More precisely:

- if `tap_interrupted` is  `False` for `HT` (i.e. the interruption should happen on press, that is now), there is no interruption and the hold behavior is not triggered (at least not before timeout) 
- if `tap_interrupted` is `True` (i.e. the interruption should happen on release), the press of A is not saved in `send_buffer`, preventing the release of A to trigger the interruption (so, here too, HT remains a tap).

## What about changing the module order?

Well, if `Combos` is loaded first, when `Combos`' `process_key` runs for A, the hypothetical current combo will be cancelled and the module will replay all the keys it had stored in this prospect, before returning. Hence, in particular, the `on_press` handler of HT will be executed before `HoldTap`'s `process_key`, which will now do what it is supposed to do, since `HoldTap` module knows a hold-tap key is currently pressed.

## Proposed fix

Easy one: write in the doc that `Combos` should be loaded first.

More failsafe one: add a priority attribute in every module and have the modules list sorted according to priority (`Combos` would have higher priority than `HoldTap`).

xs5871:
Yes, this is known behavior and "technically" mentioned in the `Combos` docs.
The order of modules is intentionally not enforced and it's up to the user to decide which module they want to give precedence.
[...]

PR #537, opened 2022-07-27
add continue-with-module-nr-idx to key processing

xs5871:
This is an approach to make modules with side effects more consistent (related to #495).
At the moment, I think only hold-tap (and it's decendents) and combos profit. Modified keys can be re-inserted into the `process_key` pipeline after the module that modified them, basically "picking up where they left of". It's not a super clean, automatic thing for all modules (yet), I acknowledge that; but it's a simple enough solution, doesn't break unit-tests or changes too much of the core firmware.
Input from contributors of #495 is of course welcome, especially towards perceived effectivenes.

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -191,6 +191,10 @@ class KMKKeyboard:
 
         return self
 
+    def resume_process_key(self, key, is_pressed, int_coord, module):
+        index = self.modules.index(module) + 1
+        self.pre_process_key(key, is_pressed, int_coord, index)
+
     def remove_key(self, keycode):
         self.keys_pressed.discard(keycode)
         return self.process_key(keycode, False)
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -158,8 +158,8 @@ class KMKKeyboard:
     def debug_enabled(self, enabled):
         debug.enabled = enabled
 
-    def pre_process_key(self, key, is_pressed, int_coord=None):
-        for module in self.modules:
+    def pre_process_key(self, key, is_pressed, int_coord=None, index=0):
+        for module in self.modules[index:]:
             try:
                 key = module.process_key(self, key, is_pressed, int_coord)
                 if key is None:
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -1,9 +1,15 @@
+try:
+    from typing import Optional
+except ImportError:
+    pass
+
 from supervisor import ticks_ms
 
 from kmk.consts import UnicodeMode
 from kmk.hid import BLEHID, USBHID, AbstractHID, HIDModes
-from kmk.keys import KC
+from kmk.keys import KC, Key
 from kmk.kmktime import ticks_add, ticks_diff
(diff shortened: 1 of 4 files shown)


klardotsh (commented):
My only real commentary on a quick read is that tracking `next_module` seems like a concern of the Keyboard class, and not of each module individually. That would also make that type of pipelining operation easily unit testable in a single location.

Otherwise, looks sane to me!

xs5871:
You're absolutely right. Incidentally that's what I tried first and got tangled up in a terrible, un-working mess, as first implementations go. So I settled for the bare minimum, easyiest solution without disrupting to much of the current code base.
Should have been a draft from the start.

klardotsh:
Optional side quest since I just thought of it: seems like maybe a good introductory case for #226/#497 too - this would be a new "interface" modules can speak over, and is nice and tightly scoped (read: shouldn't spill over into tons of the other code), so any relevant modules could be type checked against the new interfaces, but everything else left alone?

xs5871:
@klardotsh what's your opinion on this implementation?
I'm not used to using type hints in python yet, but it seems like something we may want to insist on for core contributions, bit for bit.

kdb424:
Type hints are on my hint list when I have spare time to actually write actual code. IMHO, that's a massive improvement for everyone. Tooling loves it, tests love it, and reduces confusion. Klardotsh and I seemed to agree that the core should enforce it (talked off platform) IIRC.

kdb424 (approved):
I'm in favor of this change. I'll let @klardotsh offer a final code review, but this seems like it will be a powerful thing going forward, and solve some issues. Maybe this should be documented somewhere for it's uses or in a way that people will know why this exists with some simple examples.

kmk/kmk_keyboard.py
```diff
@@ -1,9 +1,15 @@
+try:
+    from typing import Optional
+except ImportError:
+    pass
```
klardotsh, line 4: I'm not aware of a situation where this import can fail _and_ inline type annotations will be implemented. I'd imagine any Python implementation going through the effort to implement the lexer/parser bits would also implement the stdlib bits?

"version-resilient" type annotations in Python take the comment form, but we decided in https://github.com/KMKfw/kmk_firmware/pull/501 to just bump minimum CPy version to 7.x, which implements the non-comment annotations anyway.

klardotsh (approved):
Awesome! No meaningful implementation feedback here, LGTM.

As for requiring type annotations on core contributions going forward, I'd be very down with that, and in the spirit of #205, I'd also support a gradual "go back and type the things" effort.

kmk/kmk_keyboard.py
xs5871: I've got CPy v7.3.2 on my test board, and it doesn't have the `typing` builtin.

klardotsh: and the ref to Optional is.... I suppose ignored by the parser later on, then?
 
if it runs, :shipit: , interesting tradeoffs but it's a microcontroller, I get it

xs5871: Just for info: I checked a couple of repos, and that's the idiom official adafruit libs use... Guess it's the best we can do for now.

PR #542, opened 2022-08-01
Update support.md

xgpt:


xs5871:
Wouldn't it be more helpfull in the long run to fix it, instead of putting "it doesn't work" in the docs? (no offense)
Just as a general question: what do we want it to mean? We don't have versions per se.
Could we put the number of commits or merged PRs in there via a hook, for example?

kdb424:
We also only have discord as a secondary bridge, so we will not want that linked in the docs. That requires users to sign up to a platform that we allow access to, but don't directly "support". We would do better to fix things, or remove things from docs if they are untrue as opposed to leaving "it's broke" in them with discord links.

klardotsh (changes requested):
Bluntly: if I were an end user, I would have no idea what actions to take based on this snippet of the README. Furthermore, Discord is a hidden black box without an account (which, especially given Discord is a second-class citizen and only the Matrix room is "officially supported", not everyone will have), so linking to a message hidden in its depths is not a great form of documentation. The content of the message, if used in documentation, should be copied into hard Markdown docs here in the tree.

Lastly, context: yes, copied-from-git is now basically the only value anyone will ever see. Way back when, the "official" way of installing KMK was from ZIP files that had that value overwritten at build time (in CI) before being compiled down to MPY. As discussed in a recent issue I'm too lazy to go dig up, we don't do that anymore. Thus, the real docs fix here is to simply remove the request to dump that constant, and frankly, also to remove the constant from the source tree. It's eating up extremely valuable bytes on flash memory and is literally never used or even usable anymore.

klardotsh:
While I'm unilaterally closing this since we've got three votes all to the same effect from the maintainer crew, I do still want to explicitly thank @xgpt for bringing this to our attention, and have opened #544 as a backlog "documentation tech debt" item so we circle back to fixing the core issues exposed by this discussion.

Thank you!

PR #545, opened 2022-08-02
[usb] Allow kb to lost usb connection without getting lost

crazyiop:
When the usb connection is lost (for example when plugged into a pc that
keep the usb powered when shut down) the kb is lost and need to be
unpluged to be rebooted.

My guess is that the failling `hid_report.send()` change some internal
state that could not recover by itself.

This prevent even trying to send the report when the usb is not
connected. This fix has been tested working on me and another discord
user so far. The result seems consistent even if the test period has not
been very long yet.

### DIFF
--- a/kmk/hid.py
+++ b/kmk/hid.py
@@ -1,3 +1,4 @@
+import supervisor
 import usb_hid
 from micropython import const
 
@@ -229,6 +230,9 @@ class USBHID(AbstractHID):
                 continue
 
     def hid_send(self, evt):
+        if not supervisor.runtime.usb_connected:
+            return
+
         # int, can be looked up in HIDReportTypes
         reporting_device_const = evt[0]
 


kmk/kmk_keyboard.py
```diff
@@ -87,7 +87,7 @@ def _print_debug_cycle(self, init=False):
             debug(f'keys_pressed={self.keys_pressed}')
 
     def _send_hid(self):
-        if self._hid_send_enabled:
+        if self._hid_send_enabled and runtime.usb_connected:
```
xs5871, line 90: Wouldn't that only address USB connections and probably break bluetooth?
We may even want to catch connection errors in the HID classes, not `KMKKeyboard`.
Definitely needs hardware testing.

crazyiop: @xs5871 Ho good point! Haven't thought of bluetooth... Fixed

kmk/hid.py
```diff
@@ -229,6 +230,11 @@ def post_init(self):
                 continue
 
     def hid_send(self, evt):
+        if not supervisor.runtime.usb_connected:
+            # No more usb host: pc shut down, but still power on usb
+            # Don't try to send data as it mess with internal state
```
xs5871, line 235: Sorry, but this is not a helpfull comment.

xs5871 (approved):
Awesome, thanks!

PR #547, opened 2022-08-03
Add Pico87 board & update Lock Status

hendrix04:
This PR adds the Pico 87 board from Bolt Industries (https://github.com/mountainmannm/Bolt-Industries) and it adds a new callback to Lock Status so that it is easier to keep status LEDs in sync with the different lock statuses.


xs5871:
At the very least we have to split this into two self contained topic branches, one for the new board, one for the status led issue.

hendrix04:
> At the very least we have to split this into two self contained topic branches, one for the new board, one for the status led issue.

More than happy to, but might I ask why? This repo doesn't seem to do any releases or tagging so I assume everyone is running off of master. Since the board code is dependent on the changes to lock_status.py, doing the changes in 2 different PRs means tracking dependencies between two different PRs and if for some reason the changes to lock_status.py need to be reverted, the board code also needs to be reverted. How would you suggest we track that dependency?

docs/lock_status.md
```diff
@@ -16,8 +16,38 @@ when the lock is enabled and `False` otherwise.
 
 |Method                    |Description |
 |--------------------------|------------|
-|`locks.get_caps_lock() `  |Num Lock    |
```
kdb424, line 19: Is there a reason that caps_lock appears to have been replaced?

kdb424 (changes requested):

kdb424:
> > At the very least we have to split this into two self contained topic branches, one for the new board, one for the status led issue.
> 
> More than happy to, but might I ask why? This repo doesn't seem to do any releases or tagging so I assume everyone is running off of master. Since the board code is dependent on the changes to lock_status.py, doing the changes in 2 different PRs means tracking dependencies between two different PRs and if for some reason the changes to lock_status.py need to be reverted, the board code also needs to be reverted. How would you suggest we track that dependency?

It doesn't technically *have* to be split, however whatever is the most resistance will bog down merging of things we don't have an issue with. We *generally* try to merge boards and features separately as if something needs reverted, we can revert commits, which then would strip out unintended things. I'm not saying "You must do this or we won't look at it" but we respectfully request PR's to have separate things in each of them. A feature, or bugfix, or board, ect.

docs/lock_status.md
hendrix04: There was a typo in the documentation. The method get_caps_lock was listed twice, once for caps lock and once for num lock. If you look at the side by side diff, you will see the change more clearly.

hendrix04:
> > > At the very least we have to split this into two self contained topic branches, one for the new board, one for the status led issue.
> > 
> > 
> > More than happy to, but might I ask why? This repo doesn't seem to do any releases or tagging so I assume everyone is running off of master. Since the board code is dependent on the changes to lock_status.py, doing the changes in 2 different PRs means tracking dependencies between two different PRs and if for some reason the changes to lock_status.py need to be reverted, the board code also needs to be reverted. How would you suggest we track that dependency?
> 
> It doesn't technically _have_ to be split, however whatever is the most resistance will bog down merging of things we don't have an issue with. We _generally_ try to merge boards and features separately as if something needs reverted, we can revert commits, which then would strip out unintended things. I'm not saying "You must do this or we won't look at it" but we respectfully request PR's to have separate things in each of them. A feature, or bugfix, or board, ect.

I totally get that, and if there was some sort of tagging / release system I would be 100% on board with doing 2 different PRs. Given that there isn't, in my experience, it is better to package them together. As mentioned earlier, the board is reliant on the change to lock_status, so if you reverted that without reverting the board, it would break the board.

xs5871:
@kdb424's answer covers the gist of of.
It's about organization and keeping it manageable for maintainers and contributors.
- Changes to core need a different level of discussion and foresight than new boards.
- Especially for discussions about core development, we want these discussions to be searchable and findable. That's a lot easier when the title _and_ content match up, and don't contain other "random" bits and pieces -- both in the first place, and when there's the need to reference it in the future.
- Tying into that is catching the interest of other contributors. There's no need for everyone to read through the entire code proposal for new boards, but changes to core (and modules/extensions) benefits from being easily distinguished as "something more important you may want to join the discourse on".

About how to split: you can for example make the board PR a draft, add a comment like "depends on #xxx", or leave the part that needs changes out for of the first PR and make a second PR to improve / enhance the board, once there's support for the enhancement.

I'd like this to be split especially, because there's added value in discussing the solution you propose and alternatives, and circling back: it'd be awesome if that discussion was properly labeled and represented as it's own PR.

ryansbrink:
FWIW, the `lock_status.py` changes resolved the issues I was having with my caps lock LED getting out of sync on my Pico-based keyboard.  Would at least be great to see that and the example added to the README merged into master soon.

hendrix04:
This can be rejected due to https://github.com/KMKfw/kmk_firmware/pull/553

PR #552, opened 2022-08-05
fix mouse_keys not working at all

### LINKED ISSUE #534: Mouse keys not working
Some new features have been added since I last download kmk, but when I downloaded the newest version, mousekeys stopped working. When I try to move the mouse, it does move a tiny bit but then it stops. I switched back to the old version of `kmk/modules/mouse_keys.py` and it works again, so I assume that's where the issue is. Diff between newest version and the version that worked for me:
```diff
126c126
<         if self.pointing_device.hid_pending and keyboard._hid_send_enabled:
---
>         if self.pointing_device.hid_pending:
128d127
<             self.pointing_device.hid_pending = False
```

I am running kmk on a pico if that helps.

xs5871:
resolves #534.
Quick and dirty fix to get mouse keys working at all, at least.

### DIFF
--- a/kmk/modules/mouse_keys.py
+++ b/kmk/modules/mouse_keys.py
@@ -111,15 +111,15 @@ class MouseKeys(Module):
                 # print(self._next_interval)
                 if self.move_step < self.max_speed:
                     self.move_step = self.move_step + 1
-                    if self._right_activated:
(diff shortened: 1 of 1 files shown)


kdb424 (approved):

kdb424:
Something is better than nothing. Yolo

PR #553, opened 2022-08-06
Docs: How to react to lock status changes

hendrix04:
Add a callback to lock status so that thing  such as LEDs can be kept in sync.

### DIFF
--- a/kmk/extensions/lock_status.py
+++ b/kmk/extensions/lock_status.py
@@ -17,10 +17,11 @@ class LockStatus(Extension):
     def __init__(self):
         self.report = None
         self.hid = None
+        self._report_updated = False
         for device in usb_hid.devices:
+
             if device.usage == HIDUsage.KEYBOARD:
                 self.hid = device
-                self.hid.get_last_received_report()
 
     def __repr__(self):
         return f'LockStatus(report={self.report})'
@@ -40,8 +41,15 @@ class LockStatus(Extension):
     def after_hid_send(self, sandbox):
         if self.hid:
             report = self.hid.get_last_received_report()
-            if report[0] != self.report:
+            if report and report[0] != self.report:
                 self.report = report[0]
+                self._report_updated = True
+            else:
+                self._report_updated = False
+        else:
+            # _report_updated shouldn't ever be True if hid is
+            # falsy, but I would rather be safe than sorry.
+            self._report_updated = False
         return
 
     def on_powersave_enable(self, sandbox):
@@ -50,17 +58,28 @@ class LockStatus(Extension):
     def on_powersave_disable(self, sandbox):
         return
 
+    @property
+    def report_updated(self):
+        return self._report_updated
+
+    def check_state(self, lock_code):
+        if self.report:
+            return bool(self.report & lock_code)
+        else:
+            # Just in case, default to False if we don't know anything
+            return False
+
     def get_num_lock(self):
-        return bool(self.report & LockCode.NUMLOCK)
+        return self.check_state(LockCode.NUMLOCK)
 
     def get_caps_lock(self):
-        return bool(self.report & LockCode.CAPSLOCK)
+        return self.check_state(LockCode.CAPSLOCK)
 
     def get_scroll_lock(self):
-        return bool(self.report & LockCode.SCROLLLOCK)
+        return self.check_state(LockCode.SCROLLLOCK)
 
     def get_compose(self):
-        return bool(self.report & LockCode.COMPOSE)
+        return self.check_state(LockCode.COMPOSE)
 
     def get_kana(self):
-        return bool(self.report & LockCode.KANA)
+        return self.check_state(LockCode.KANA)
--- a/kmk/extensions/lock_status.py
+++ b/kmk/extensions/lock_status.py
@@ -19,7 +19,6 @@ class LockStatus(Extension):
         self.hid = None
         self._report_updated = False
         for device in usb_hid.devices:
-
             if device.usage == HIDUsage.KEYBOARD:
                 self.hid = device
 
@@ -63,6 +62,7 @@ class LockStatus(Extension):
         return self._report_updated
 
     def check_state(self, lock_code):
+        # This is false if there's no valid report, or all report bits are zero
         if self.report:
             return bool(self.report & lock_code)
         else:
--- a/kmk/extensions/lock_status.py
+++ b/kmk/extensions/lock_status.py
@@ -67,3 +67,4 @@ class LockStatus(Extension):
 
     def get_kana(self):
         return bool(self.report & LockCode.KANA)
+    
--- a/kmk/extensions/lock_status.py
+++ b/kmk/extensions/lock_status.py
@@ -14,10 +14,9 @@ class LockCode:
 
 
 class LockStatus(Extension):
-    def __init__(self, fn=None):
+    def __init__(self):
         self.report = None
         self.hid = None
-        self.fn = fn
         for device in usb_hid.devices:
             if device.usage == HIDUsage.KEYBOARD:
                 self.hid = device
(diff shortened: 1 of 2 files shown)


xs5871:
I don't believe that a callback is the best way to go about this.
Would could argue about the shortcomings: it's opaque; can't be validated; you don't no where, when and under what conditions the callback is run; there's no error handling. There's probably more and most of that could be "fixed", but that discussion wouldn't really getting us anywhere.

The recommended (and some might argue nicer, pythonic) solution is extension by inheritance in userspace:
```python
# in your main.py
from kmk.extensions.led import LED
from kmk.extensions.lock_status import LockStatus

leds = LED(led_pin=[board.GP27, board.GP28])

class LEDLockStatus(LockStatus):
    def set_lock_leds(self):
        if self.get_caps_lock():
            leds.set_brightness(50, leds=[0])
        else:
            leds.set_brightness(0, leds=[0])

        if self.get_scroll_lock():
            leds.set_brightness(50, leds=[1])
        else:
            leds.set_brightness(0, leds=[1])

    def after_hid_send(self, sandbox):
        super().after_hid_send()
        self.set_lock_leds()

keyboard.extensions.append(leds)
keyboard.extensions.append(LEDLockStatus())
```

As I mentioned in #503, I don't think there's any functionality missing in the modules, we're missing proper documentation with examples of how to string different modules and extensions together.

kmk/extensions/lock_status.py
```diff
@@ -15,11 +15,12 @@ class LockCode:
 
 class LockStatus(Extension):
     def __init__(self):
-        self.report = 0x00
+        self.report = None
```
xs5871, line 18: Why is necessary to set the report to `None`? It's always a byte and the "no locks are enabled" state is `0x00`.

kmk/extensions/lock_status.py
```diff
@@ -15,11 +15,12 @@ class LockCode:
...
         self.hid = None
         for device in usb_hid.devices:
             if device.usage == HIDUsage.KEYBOARD:
                 self.hid = device
+                self.hid.get_last_received_report()
```
xs5871, line 23: You're not even storing the report anywhere. I don't get what this changes.
Even so, `get_last_received_report` is called periodically anyway, why is a call specifically during init (not even during bootup) necessary?

xs5871 (commented):
Sure thing. I do have questions though, primarily for clarification, so that I can understand what's happening.

kmk/extensions/lock_status.py
hendrix04: There is no way to know what the state it at the initialization of this class. Therefore inferring that the state is `0x00` is incorrect and can lead to false information being reported in people's code.

kmk/extensions/lock_status.py
hendrix04: Calling `get_last_received_report` here forces `after_hid_send` to be invoked setting the appropriate value for `report` in the class object.

hendrix04:
> Sure thing. I do have questions though, primarily for clarification, so that I can understand what's happening.

Answers on the code comments.

kmk/extensions/lock_status.py
xs5871: The only reason that this wouldn't update once the keyboard firmware is running is when you poll the state once during the setup and never again. And at this point `None` would throw an exception.
I don't see how initializing it to `None` would help.
Sorry if I'm being tedious, I just don't want to "fix" something that's not broken.

kmk/extensions/lock_status.py
xs5871: Would it make more sense to put a `self.report = self.hid.get_last_received_report` in `during_bootup`?

kmk/extensions/lock_status.py
hendrix04: I made the change based on my findings when testing my original change. Now that I think about it though, my callback was inside of a conditional that only called the function if the report changes. Since report was initializing as 0 and the last report was 0, there was no change and my function was never getting called.

I could see someone wanting to do the same thing so that they could only trigger code when the report actually changes.

As far as exceptions go, aren't all exceptions caught and ignored unless debugging in which case it just prints to log? It isn't good practice to purposely throw exceptions for no reason, but setting the value to a random value also doesn't sit well with me either. Technically, initializing it to None is the most a appropriate value at the time. I am open to suggestions here.

kmk/extensions/lock_status.py
hendrix04: So, my initial code did that. The problem was two fold.
1) I would have had to duplicate my callback code in the init function and I hate duplicating code
2) `after_hid_send` was getting called either way so why not just utilize the code that already exists?

While writing my other comment it dawned on me that with your suggested implementation, we lost the ability to know definitively that report has changed.

Maybe we want another class variable that holds whether or not report was updated on the most recent invocation of `after_hid_send`?

kmk/extensions/lock_status.py
xs5871: > I could see someone wanting to do the same thing so that they could only trigger code when the report actually changes.

A `has_changed` class attribute for users to check against sounds reasonable.

> As far as exceptions go, aren't all exceptions caught and ignored unless debugging in which case it just prints to log? It isn't good practice to purposely throw exceptions for no reason, but setting the value to a random value also doesn't sit well with me either. Technically, initializing it to None is the most a appropriate value at the time. I am open to suggestions here.

Well, 0 is at least a valid value, and it meaning "no lock is active" isn't totally arbitrary as a fallback if in doubt. That value is only maintained for the first run of the firmware main loop anyway.
The reason that there's a check for `self.hid` and the report isn't polled at init, if I remember correctly, is that in certain circumstances the hid connectivity isn't there at all (split keyboard), delayed after init (at boot, BLE), or intermittently not availabel (due to BIOS/OS handover, or OS driver re-initializations).
Exceptions are only caught if any of the `get_xxx_lock` methods are called from within module / extension handlers, not during setup. And even then, something like a split not connected to USB may try to throw > 200 exceptions per second, which it won't manage, because it'll slow the entire device down to a crawl.

I think that could be somewhat mitigated by checking against the aforementioned future `has_changed` attribute. But as an internal API, the `get_xxx` methods should never be allowed to throw an exception in my opinion. Then I'd rather initialize it to 0x80 for example. That'll trigger the first `has_changed` without interfering with the `get_xxx` methods.

hendrix04: Ah, yeah definitely use cases that I am not aware of.

Fun fact though, according to https://docs.circuitpython.org/en/latest/shared-bindings/usb_hid/index.html, `get_last_received_report` can return `None` so it isn't unprecedented.

Maybe a better idea would be to put appropriate checks in the code to avoid doing something that throws an exception.

kmk/extensions/lock_status.py
```diff
@@ -49,17 +58,28 @@ def on_powersave_enable(self, sandbox):
     def on_powersave_disable(self, sandbox):
         return
 
+    @property
+    def report_updated(self):
+        return self._report_updated
+
+    def check_state(self, lock_code):
+        if self.report:
```
xs5871, line 66: This tripped me up. I think we could add a comment "This is false if there's no valid report, or all report bits are zero".

kmk/extensions/lock_status.py
```diff
@@ -15,9 +15,11 @@ class LockCode:
 
 class LockStatus(Extension):
     def __init__(self):
-        self.report = 0x00
+        self.report = None
         self.hid = None
+        self._report_updated = False
         for device in usb_hid.devices:
+
```
xs5871, line 22:
```suggestion
```
remove line break.

xs5871 (commented):
The two comments aside, I like it a lot. Special thanks for adding documentation! Very much appreciated.

kmk/extensions/lock_status.py
hendrix04: Hmm, valid point here... Should we change this to `if self.report is not None`? While functionally it would return the same value, it is more correct.

xs5871: In my opinion a comment is sufficient.
I don't mind the implicit double check, and in case of zero it's a bit faster then checking against zero again later.

PR #561, opened 2022-08-28
Shrink ./kmk/ with `nox` & `mpy-cross`.

LukeDRussell:
The nice!nano CircuitPython doesn't leave enough storage for an
uncompiled ./kmk/ folder, with main.py and Bluetooth libraries.

This introduces a soft dependancy to kmk on `nox`, as a OS-agnostic
alternative to `make`. Bonus is the use of native python to describe
steps / options / sessions.

## To test this PR:
```sh
pip install pipx
pipx install nox
nox compile
```


LukeDRussell:
Vote 1: `black` conformance with double quotes. Being different with the single quotes has no benefit to this project.

kdb424:
Can I ask why the addition of nox over makefiles? This would cause more maintenance over just a makefile. As for black not allowing double quotes, this was done as both @klardotsh and I both agreed that single quote was the only quote to us, but it should be enforced one or the other to keep consistency with code so people know what they should use. It's infuriating to have to constantly remember which is which depending on which line of code I'm on, especially considering that auto formatting exists.

LukeDRussell:
Why not `make`?

1. Because `make` doesn't (easily) install on Windows, as per the first comment. _I gave up after 2 hours_.
2. So our build scripts can be written in Python. _The single biggest advantage KMK has over QMK and ZMK is it's written in (Circuit)Python (2nd being access to Micro and Circuit Python ecosystems). I want to write in Python, not C, make, bash, etc_.

Why `nox` in particular?
1. Tasks are written in pure Python.
2. It's popular. Pip and Pipx use it.
3. It takes care of its own venvs. 

I also considered [Invoke](https://www.pyinvoke.org/), which has the interesting option of being used as library and packaged into CLI tools (`kmk` perhaps?). I figured `nox` being able to manage it's virtualenvs is an advantage for contributing newcommers. It may have also just 'felt nicer' to me. But I'd be happy to have a go at reimplementing this with Invoke. Both `nox` and `invoke` improve KMK's ability to be contributed to by more folks across any Operating System.

As for quotes, I agree completely about consistency. I'm suggesting we go to double-quotes as that won, convincingly. The issue link in KMK/README.MD has been closed for four years now, and it's done nothing to [the popularity of black](https://github.com/psf/black/network/dependents) - 6243 packages depend on it as I write. 

I'm also suggesting that maintaing the exception isn't worth the effort for any contributor. By exculding string normalisation, we can't just run `black` to have it taken care of. Instead, we all have to write in whatever our preference is, then wait for Flake8 to complain, then go and replace the double-quotes. Or worse, wait for CI to fail and hope the contributor understands how to read the failure so a maintainer doesn't have to explain the problem.

I'm happy to send this to another Issue if the maintainers will entertain futher discussion.

LukeDRussell:
P.S. I'm happy to continue with the "infrastructure" work updating dependencies, improving automation, improving cross-OS use, etc, etc.

LukeDRussell:
I could also just put this as a stand-alone script as `./util/compile.py`, moving the dependency discussion to an Issue.

kdb424:
I'm not going to block this. I was mostly curious. Sorry if I came off harsh. I'm not dead set on the tooling that we have, but I'd like to unify it all depending on the other maintainers responses to this tool. Make isn't perfect, but it's the tool that we knew, but nox seems nice, and I understand why you need access to mpy-cross on all platforms. I don't have enough authority to make this call, and it also effects them.

LukeDRussell:
Hey Kyle, I'm sorry if I'm coming across agressively, that's not my intent. I am trying to come across convincing though 😊. 
I really am happy to just reimplement the compile script as a stand alone .py file. I was hoping using Nox for this use case would be a good demo - though perhaps not the wisest idea to introduce a dependancy for a demo...

kdb424:
I'm not opposed. I simply had questions, you answered them to my liking, and I'm passing it off to other maintainers to get their thoughts on this. I know most of us had historically worked on Linux/BSD/Mac, so this was news to me. I agree that compiling of code is massively helpful, especially as the project gets bigger, and we must unfortunately cope with the n!n's horrid space divide unless someone fixes that upstream, or use a fork like bs-python from Boardsource that fixes that wall. I was caught a bit off gaurd, and I'm fine with it personally if it's of use to the project and passes other maintainers. It may be something we move the entire make system to to prevent double maintenance and/or a divide are now my primary concerns is all :)

klardotsh (commented):
My thoughts here are multi-fold:

- `Makefile`s are and were what I am and was used to, and work universally on POSIX systems (of which all major OSes except Windows and to some extent Plan 9 are, and Windows has WSL2 to bridge the gap). They are language-independent, which is also excellent: whether I'm faced with a Rust, Python, C, or Bash project, if a `Makefile` exists, I simply type `make` and can reasonably expect the correct things to happen. At a bare minimum, if we move to something like `nox`, I'd like to retain that ability (much like how, in NodeJS projects, I often have a Makefile target `make deps` that calls `npm ci` - I love having a reasonably standardized way to walk into a project).
- `nox` seems to combine Invoke and `pipenv`, the latter already being a tool we use. I see the overall Python meta seems to have moved from `pipenv` to `poetry` (an option I loved four years ago when spinning up this repo, but was not stable enough yet. The author consistently writes slam dunk projects, though, so the outcome doesn't surprise me 😄). My gut reaction is that I'd prefer to keep the management of dependencies separate from the management of scripts, which would lean towards a pipenv+invoke or poetry+invoke combo, rather than the single tool `nox` doing so many tasks at once. I don't hold this opinion so strongly that I'd block merge over it.
- It should not be part of this PR, but yeah, eliminating the quotes override thing for `black` is fine. I had my reasons for wanting to do that in 2018, they were wrong in retrospect (or perhaps they weren't - at the time, I never expected KMK to have contributors other than @kdb424 and myself and maybe a handful of other MicroPython nerds), we can fix that. Make sure we do the entire global reformat in a single commit in a single, dedicated PR, and I'm down with that.

All this is to say, I'm not opposed to any of the concepts proposed here, I just have opinions on how many jobs each tool has, and on the segmentation of pull requests. None of my feedback is especially blocking in the grand scheme. Thanks for proposing this and for the context comments above!

xs5871:
I don't mind outsourcing the compilation to a script that's cleverer and easier to read than the crude inline shell stuff in the Makefile. That sounds great, actually.

LukeDRussell:
So I've refactored this to be a stand-alone script for now, without any change to the `Makefile`. With further work this may become a session in `nox` or a parameter to `invoke` (or any other CLI KMK introduces).  

I'd like for this PR to be merged, with the much smaller scope.

kdb424:
I'm totally in favor of this. To be totally clear, I wasn't totally opposed to nox. I just wanted to know why, and to get other maintainers feedback. I'll happily merge this, and be open to review a follow up with that as the sole goal. Thanks for working with us on this, and feel free to mention this PR to add context to any follow ups so discussion doesn't need repeated or get lost.

kdb424 (approved):

PR #562, opened 2022-08-28
Split seperation of BLE started

kdb424:
Thinking we may be better off splitting out BLE from wired splits. This would require additional docs changes, but I wanted to open this as a draft for feedback to see if code duplication for the sake of maintenance and ease of reading each file made more sense than having them tangled. I started to separate more into functions just to improve readability in things like the `during_bootup` function as well, but I'm open to feedback on direction before I put a ton more time on this. It's untested, but mostly structural changes. I'll test wired before taking this out of draft status, and request that someone test BLE after I commit changes to the docs. I considered making a split_common, but some of the BLE stuff really had to do things different, so I wanted to just de-tangle as much as I could.

### DIFF
--- a/kmk/modules/split.py
+++ b/kmk/modules/split.py
@@ -1,13 +1,11 @@
 '''Enables splitting keyboards wirelessly or wired'''
 import busio
 from micropython import const
-from supervisor import runtime, ticks_ms
+from supervisor import runtime
 
 from keypad import Event as KeyEvent
 from storage import getmount
 
-from kmk.hid import HIDModes
-from kmk.kmktime import check_deadline
 from kmk.modules import Module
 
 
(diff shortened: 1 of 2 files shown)


xs5871:
The long-term idea once was to put all the low-level split communication stuff into transport modules. We already started that with the pico PIO UART.

What exactly does BLE do different? It needs a bit of logic to initiate and keep the connection healthy, other than that, it's literally just UART without wires. We should be able to reuse most of the split code.

xs5871:
I made a dirty mockup based on your changes of what I had in mind: https://github.com/xs5871/kmk_firmware/tree/split_shuffle
Carefull, it's an untested example and will most likely not work. Thought it would be easier to communicate that via code.

kdb424:
I agree that your solution is better. I'd much rather use that as a basis going forward. Closing this for now. Feel free to open as a draft and allow maintainers to edit, and I'll work off of that as a base if you would like.

PR #575, opened 2022-09-07
draft concept for ble-split to transport refactor

xs5871:
As discussed in #562 here's a draft of how to seperate the BLE communication logic out into a "black box" transport.
The idea is to have "transports" that handle the wire-level communication (UART, PIO UART, BLE, I2C, ...), which only handle the transfer of a sequence of bytes; the higher level split protocol is then independent of the low level transfer protocol.
I'm currently not in a position to test and debug this suggestion and leave the PR editable to maintainers.


kdb424:
Continuing work on #618

PR #578, opened 2022-09-08
Update combos.md

### LINKED ISSUE #573: make_key not working; issue with lambda?
when I use the code from the docs (combos page), I get an error:
```
  File "code.py", line 191, in <module>
  File "kmk/kmk_keyboard.py", line 424, in go
  File "kmk/kmk_keyboard.py", line 477, in _main_loop
  File "kmk/kmk_keyboard.py", line 101, in _handle_matrix_report
  File "kmk/kmk_keyboard.py", line 151, in _on_matrix_changed
  File "kmk/kmk_keyboard.py", line 182, in pre_process_key
  File "kmk/kmk_keyboard.py", line 188, in process_key
  File "kmk/keys.py", line 474, in on_press
TypeError: function takes 0 positional arguments but 4 were given
```
(in this instance, I am not trying to use it in a combo yet, just assigning KC.MYKEY to a key on my keyboard)

If I do try to use it in a combo, I get a different error:
```
  File "code.py", line 192, in <module>
  File "kmk/kmk_keyboard.py", line 424, in go
  File "kmk/kmk_keyboard.py", line 484, in _main_loop
  File "kmk/kmk_keyboard.py", line 255, in _process_timeouts
  File "kmk/modules/combos.py", line 153, in <lambda>
  File "kmk/modules/combos.py", line 240, in on_timeout
ValueError: object not in sequence
```

I attempted to look for a keymap in the repo which used `make_key` to see if I could use their code and try to make it work. I ended up finding the "shrek_is_love" code (lol) in the Klarank keymap. This code works, in that it will print Shrek's face to my debug console, but it won't actually type anything like an HID keyboard. I attempted to change the print statements to `send_string `and even `simple_key_sequence`, but neither worked. The keyboard functions fine, it even registers the keypresses. That key just doesn't type anything.

I also tried incorporating lambda into the shrek code, but receive the same error as before with the code from the docs (but often with different numbers - "takes 4 positional arguments and 5 were given" etc)

So, I can print stuff to the console just fine, but if I try to use it to actually output text using lambda, I get the positional arguments error.

I am not sure if  this is a bug or I am simply making a mistake somehow! Any help would be appreciated.

I am using an ItsyBitsy nRF52840 Express

xs5871:
fix example code in combos docs.
resolves #573

### DIFF
--- a/docs/combos.md
+++ b/docs/combos.md
@@ -37,7 +37,7 @@ keyboard.modules.append(combos)
 
 make_key(
     names=('MYKEY',),
-    on_press=lambda: print('I pressed MYKEY'),
+    on_press=lambda *args: print('I pressed MYKEY'),
 )
 
 combos.combos = [


kdb424 (approved):

kdb424:
Thanks for getting that done quickly!

PR #584, opened 2022-09-14
fix import typo in cg_swap.md

ChrisChrisLoLo:
update doc typo so that import will now correctly point to the `modules` directory rather than the non-existent `module` directory


xs5871 (approved):
appreciated

PR #587, opened 2022-09-16
Add key position combos.

scorp200:
Added an option to use key coord in the matrix to make combos.
this allows to press / release combos regardless of layer and Key.


kdb424 (approved):
Looks good to me. Need tests to pass before merge.

kmk/modules/easypoint.py
```diff
@@ -120,7 +120,7 @@ def _read_raw_state(self):
         return x, y
 
     def getSignedNumber(self, number, bitLength=8):
-        mask = (2 ** bitLength) - 1
+        mask = (2**bitLength) - 1
```
kdb424, line 123: Not sure why this was touched. It needs to not have been.

kdb424:
There's also a .DS_Store file that needs to not be included.

kmk/modules/combos.py
```diff
@@ -127,7 +139,9 @@ def on_press(self, keyboard: KMKKeyboard, key: Key, int_coord: Optional[int]):
         for combo in self.combos:
             if combo._state != _ComboState.MATCHING:
                 continue
-            if combo.matches(key):
+            if isinstance(combo, (Chord, Sequence)) and combo.matches(key):
+                continue
+            elif isinstance(combo, Coord) and combo.coord_matches(int_coord):
                 continue
```
xs5871, line 145: I'm not a big fan of that pattern. We should pass both `key` and `int_coord` into `combo.matches()`, and decide there which one to use.

kmk/modules/combos.py
```diff
@@ -65,6 +68,15 @@ def matches(self, key):
             return False
 
 
+class Coord(Combo):
+    def coord_matches(self, int_coord: int):
```
xs5871, line 72: See previous comment:
```suggestion
class Coord(Combo):
    def matches(self, key: Key, int_coord: int):
```

xs5871 (commented):
Good idea. I knew that that was a missing feature, and planned to implement it as soon as someone asked for it. Less work for me.

We may even think about adding a switch to `Chords` and `Sequences`, instead of adding a new class, like:
```
Chord((KC.1, KC.2), KC.3)
Chord((7, 8, 9), 6, match_coord=True)
```
edit: or, even fancier, we could make the combo figure out during `__init__`, wether it's a key, or coord combo.

... and at least one or two unit tests.

kmk/modules/combos.py
scorp200: yeah thats actually a good idea, I wasn't sure if my approach is a good practice, will make the changes

PR #588, opened 2022-09-16
GC to reduce memory fragmentation

kdb424:
Seems like a little memory management at boot leaves more ram free, and less fragmented. Doesn't seem to noticeable change boot times. Would like to hear feedback from other users if there are any issues. Any board I have tested, including the one that typed everything in the PR and this message is running it fine.


xs5871:
Does this fix any actual issues? I've never seen that _fragmentation_ is any problem. In any way, calling gc manually generally doesn't reduce fragmentation.
I did some extensive tests when I removed the gc calls from our code. I saw zero benefit from having them there.

On a much more important note: a gc call blocks for a _very_ long time. We should never, ever put that in the main loop.

kdb424:
> Does this fix any actual issues? I've never seen that _fragmentation_ is any problem.
It can cause issues during boot if things aren't cleaned up. I did benchmark total free memory after a fresh boot without and with this, and there was at least some extra free memory that was usable. 

> In any way, calling gc manually generally doesn't reduce fragmentation. I did some extensive tests when I removed the gc calls from our code. I saw zero benefit from having them there.
This may have changed in Cpy upstream versions. I've noticed changes for sure in it's behavior, and certain things massively benefit from it. Images for oleds being a massive one, but it could apply to other areas, which is why I looked into it. More reading on that here. https://learn.adafruit.com/Memory-saving-tips-for-CircuitPython?view=all

> On a much more important note: a gc call blocks for a _very_ long time. We should never, ever put that in the main loop.
I can concede that this may be correct. I do agree that in older cpy versions, I did notice, and feel, a massive lag spike in the main loop, and removing it helped massively. I added it in the main loop mostly to check how it responded on other's boards, but on the board that I have typed all of this on, it's running constantly without any hitches. Conversely, I did used to notice some hitches at seemingly random times before I added this. I'd like to see more testing, and understand why you have these concerns, and why I am leaving this up here for people to test. Even a year ago, GC responded very differently than it does on my boards now. I should have opened this as a draft, which was a mistake.

xs5871:
Well, in general you shouldn't ever need to touch the garbage collection, unless the code is garbage (no offense btw.). That may not be strictly true in the embedded world, but constantly invoking gc should be considered a dirty hack at best.

For example:
If OLEDs need large chunks of contiguous memory they should allocate that early on, during instantiation, way before the keyboard code starts running.

sidenote:
I testet with 7.3.2 very recently. `gc.collect()` can take anywhere between 2-20ms, if I recall correctly. There should be accurate numbers in a past PR, I can dig it up if you're interested. That's between not-noticable and definitely noticable for a fast typest. I'm not one of the _performanc-at-all-costs_ people, but artificially increasing lag for no apparent benefit, just in case, doesn't sit well with me either.

PR #592, opened 2022-09-17
Add key coord combos - Replaces #587

scorp200:
Added an option to use key coord in the matrix to make combos. this allows to press / release combos regardless of layer and Key.


kdb424 (approved):

xs5871 (approved):
neat

PR #596, opened 2022-09-20
sequences update with new example

kdb424:
Helped a user create this macro, and thought this may help others with the same issue. `no_press' and 'no_release' are used in the sequences source, but not documented well here, so this seemed like a good inclusion.


xs5871 (approved):

xs5871:
I think there's nothing controversial here.

PR #600, opened 2022-09-22
add type hinting for kmk_keyboard

xs5871:


kdb424 (approved):
It's been attempted before, but I'll take it in chucks rather than a grand sweep that fails. Ship it!

kmk/kmk_keyboard.py
```diff
@@ -69,7 +71,7 @@ class KMKKeyboard:
     # 6.0rc1) this runs out of RAM every cycle and takes down the board. no
     # real known fix yet other than turning off debug, but M4s have always been
     # tight on RAM so....
-    def __repr__(self):
+    def __repr__(self) -> None:
```
klardotsh, line 74: repr always returns a String; does this actually pass pyright?...

xs5871: Just me failing reading comprehension.
Nothing passes pyright, btw. There's so much dynamic typing going on, especially `None` as default or canary -- type hinting is mostly decorative right now.

klardotsh: Ah, yeah. Much of early KMK's code (a lot of which survives to this day, either exactly, or in spirit) was from my pre-strongly-typed-languages days (C doesn't count for this definition; more the Rust/TypeScript/Haskell/etc. stuff I've gotten into since), and `None` canaries are a thing I now harp against but definitely remember doing lots of.

Makes sense, thanks for the context!

PR #602, opened 2022-09-23
add type hinting for kmk.keys

xs5871:


kmk/keys.py
```diff
@@ -6,13 +11,20 @@
 from kmk.types import UnicodeModeKeyMeta
 from kmk.utils import Debug
 
+# Type aliases / forward declaration; can't use the proper types because of circular imports.
+Keyboard = object
+Key = object
```
klardotsh, line 16: This sounds like the typing has caught a code smell then: those circular imports should be flattened down into a separate file imported in both places, perhaps.

klardotsh (approved):

kmk/keys.py
xs5871: I don't see how. `KMKKeyboard` needs `Keys`, but `Keys` have `KMKKeyboard` as method argument.
Unless we make up dummy classes without implementation (which this more or less is), then that's an irreducible circular dependency.

klardotsh: Thinking out loud: is this a case for what in TypeScript we'd call an "interface"? (maybe that's what you're getting at, actually).

```typescript
interface Thing {
  foo: string
  bar: number | undefined
  // I don't actually remember if this syntax is right right now
  buildOutput: () => string
}
```

And then anything that has those two properties and that method, with that signature, can be passed to something expecting a `Thing`

It's a layer of indirection in exchange for a layer of safety. In TypeScript, this stuff is all erased at compile time and has no runtime cost. Unsure what the Python equivalent would be, since ABCs have runtime componentry.

xs5871: That's a decent analogy, altough obviously pythons typehints only check for the name of the type, not for any implementation details (afaik).
There may be a tiny runtime impact: the type alias may be held in the modules dictionary; I'm not familiar enough with the CP parser and vm to give an educated answer.

kdb424 (approved):
I'm generally in favor, but I'll let @klardotsh merge when they feel it's ready. I don't know enough here to have valid arguements.

PR #606, opened 2022-09-29
added Raspberry Pi Pico pinouts for the Pimoroni Keybow

Corteil:
added Raspberry Pi Pico pinouts and auto selection. to keybow.py 

Hardware Raspberry Pi Pico with a Red Robotics Pico to Zero Adaptor.

See link for more information about the adaptor.

https://www.tindie.com/products/redrobotics/pico-2-pi-adapter-board/

### DIFF
--- a/boards/pimoroni/keybow/keybow.py
+++ b/boards/pimoroni/keybow/keybow.py
@@ -1,4 +1,4 @@
-'''
+"""
 KMK keyboard for Pimoroni Keybow.
 
 WARNING: This doesn't currently function correctly on the Raspberry Pi Zero,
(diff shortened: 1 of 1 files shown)


kdb424:
This reminds me that I really should have a better answer for this. I'll probably offer up a PR to allow repinning things just by stating which board something is, and making a list of "from" and "to" boards so a board can be automatically repinned without the need for something like this. I'm not flat out rejecting this PR, but I may use it as the motivation, and knowledge, or what pins should go to where for this board, and as a test for the new function if this sounds reasonable.

kdb424:
There's some test problems with this PR. I've also started said work in #608 if you want to consider converting to that possibly? Otherwise, once tests are passing, willing to merge either way.

Corteil:
Update code should hopefully pass test now, used 'black' to format the file and made some manual adjustments and remove unnecessary whitespace.
Please do what's best for you and the project.

kdb424:
Added commits and had to open a new PR as I couldn't push to your branch. Your code is still there, and you are still the author. I've just adopted getting it fixed/merged.

PR #608, opened 2022-10-02
Quickpin initial support

kdb424:
This is intended to help ease the need for multiple kb.py by unifying pins per footprint. More footprints like feather/xiao and more can be added, but this is my initial attempt at creating the system being as careful to keep it light as I could. Additional support to ease transition from firmwares such as QMK have also been added.


daysgobye (approved):
This is awesome just what I was thinking. Would it add over head make it accessible with  dot notation instead on index? Just because it may be more maintainable then indexes.

kdb424:
> This is awesome just what I was thinking. Would it add over head make it accessible with dot notation instead on index? Just because it may be more maintainable then indexes.

Dot notation would require a dict I believe. I would have to rethink the AVR pro micro support, and come up with names that may not always be the same for pins, but if you can come up with a sane system for that, I'm not opposed.

klardotsh (commented):
As a general rule, images shouldn't usually go into Git directly: [Git LFS](https://git-lfs.github.com/) is ideally suited for that type of thing, and GitHub, Gitea, etc. all support it.

Not a blocker here since it's one 400K image. Not the end of the universe. Just not a precedent to continue setting (or clones of KMK will end up getting huge: once a binary blob is in git, the only way to get it out is to fully rewrite history using something like the [BFG](https://github.com/rtyley/bfg-repo-cleaner))

klardotsh (commented):
Other note: we have `AttrDict` to allow dot notation, and while `1` is not a valid attribute name in Python, we already do hacks like `KC.N0` to plumb to the `0` key from the keys collection: while I'm not 100% sure it's cleaner than just `board.pins.P0_01` or whatever, we *can* use `AttrDict` or similar to achieve ergonomics like `pins[0] -> pins.P0`

klardotsh (approved):
Let's give @xs5871 last rights of refusal here since I don't actually have strong opinions on implementation here other than "this is a cool idea, and the syntax seems fine to me, but then, I work in code so...."

docs/quickpin.md
```diff
@@ -0,0 +1,103 @@
+# Quickpin
+
+Quickpin helps devs quickly translate pinouts between boards of similar pinouts.
```
xs5871, line 3:
```suggestion
Quickpin helps devs quickly translate pinouts between boards of similar footprint.
```

docs/quickpin.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Adding boards to quickpin support
+
+Quickpin format is simply a list of pins in order of all through hole pins,
+going anticlockwise starting at the top left. The orientation should be with the
+chips facing toward you, with USB facing the top. If this isn't appliable, or
+otherwise is not true, it should be stated in a comment in the file. Any pin
+that is not addressable in software should be left as `None` to fill the space,
+and align pins correctly for all boards. All boards should be stored in 
+`kmk/quickpin/<footprint>/boardname.py`
```
xs5871, line 70:
```suggestion
`kmk/quickpin/<footprint>/boardname.py`.
```

docs/quickpin.md
```diff
@@ -0,0 +1,103 @@
...
+
+Pro Micro RP2040 shown as an example
```
xs5871, line 72:
```suggestion
Pro Micro RP2040 shown as an example:
```

xs5871 (changes requested):

xs5871 (approved):

PR #609, opened 2022-10-02
Boot.py docs better example

kdb424:
This should auto detect better and cause less people that copy paste to get a board in a softlock state.

### DIFF
--- a/docs/boot.md
+++ b/docs/boot.md
@@ -38,18 +38,21 @@ import storage
 import usb_cdc
 import usb_hid
 
-# This is from the base kmk boot.py
-supervisor.set_next_stack_limit(4096 + 4096)
+from kb import KMKKeyboard
+from kmk.scanners import DiodeOrientation
+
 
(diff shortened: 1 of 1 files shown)


docs/boot.md
```diff
@@ -38,18 +38,21 @@ import storage
 import usb_cdc
 import usb_hid
 
-# This is from the base kmk boot.py
-supervisor.set_next_stack_limit(4096 + 4096)
```
klardotsh, line 42: Are we intentionally ending stack size overrides? Do boards generally have enough stack now without these hacks?

kdb424: Uhhh, we have boot.py still, but no boards that I have tested lately have required that, and have taken a hit on memory by doing so. I can add this back if it's desired. I left the stock file there for users to reference as well, and I may want to make a note specifically about that stack limit as opposed to just "forcing it" by leaving it in examples. No rp2040 board, even stacked with features, has required it.

xs5871:
There's nothing to break we can't fix again.

PR #610, opened 2022-10-02
added Raspberry Pi Pico pinouts for the Pimoroni Keybow: Maintainer edit

kdb424:
Replacement PR for #606 to fix formatting.

### DIFF
--- a/boards/pimoroni/keybow/keybow.py
+++ b/boards/pimoroni/keybow/keybow.py
@@ -41,8 +41,8 @@ from kmk.extensions.rgb import RGB, AnimationModes
 from kmk.kmk_keyboard import KMKKeyboard
 from kmk.scanners.keypad import KeysScanner
 
-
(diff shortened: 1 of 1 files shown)


xs5871 (commented):
Not a blocking concern, but out of curiosity:
Why use `board_pins()` functions that return lists, instead of putting the list as tuple in the `if isBoard()` blocks?

kdb424:
> Not a blocking concern, but out of curiosity: Why use `board_pins()` functions that return lists, instead of putting the list as tuple in the `if isBoard()` blocks?

I'll eventually port this to the footprint format in my other PR if that ever merges. I didn't strongly review that because of that factor.

klardotsh (approved):
Since the pins ergonomics are handled in the other PR, no majorly blocking feedback here.

xs5871 (approved):

PR #479, opened 2022-06-16
Support for tap behavior autorepeat for modtap and layertap keys.

ald0o:
This is an implementation of QMK-like autorepeat for tap-hold keys.

Assuming you have a tap hold key, you can simulate a long press of the tap behavior (which should trigger the autorepeat feature if set in the OS). To this purpose: within the hold delay, tap the key once, then press it again and keep it pressed as long as you want the autorepeat to last.
The behavior can be disabled by passing the optional argument `repeat = False`.

I did not write the doc yet (what are the concerned files? only `modtap.md` and  `layers.md`?)

PR #611, opened 2022-10-02
Feature holdtap repeat

### LINKED ISSUE #593: Add `Tapping Force Hold` to HoldTap
# Context:
For smaller 40% boards, it is often desirable to have a key such as the backspace key double as a layer key. For example:
```python
KC.MT(KC.BSPC, KC.LSHIFT)
```

However, tapping and then holding on the backspace key doesn't result in the backspace key register as being auto-pressed. This results in users needing to manually tap the backspace key or use some other method to delete sizable strings of text.

This configuration feature is one of the few outlined in the defacto home row mods article: https://precondition.github.io/home-row-mods#tapping-force-hold. Having this feature, along with [Ignore Mod-Tap Interrupt](https://precondition.github.io/home-row-mods#ignore-mod-tap-interrupt) and [Permissive Hold](https://precondition.github.io/home-row-mods#permissive-hold) can make KMK more than viable for any 40%/sub  40% layout, and could be very attrative to a growing segment of the keyboard community

# Solution
Implement a `Tapping Force Hold`, which is implemented in QMK and ZMK. Tapping force hold effectively "auto-presses" a tapping key if it has been tapped repeatedly. Details about Tapping Force Hold can by found in the QMK docs [here](https://github.com/qmk/qmk_firmware/blob/master/docs/tap_hold.md#tapping-force-hold). It would be awesome to have this as an optional argument, especially for layouts that are dependent on this feature

xs5871:
Continuation of #479.
Resolves #593.

* repeat is reset after `tap_time`
* repeat works for both tap/hold keycode
* fixed `KC.HT` getting over-written by modules that inherit from and are loaded after HoldTap.

### DIFF
--- a/docs/modtap.md
+++ b/docs/modtap.md
@@ -31,7 +31,7 @@ keyboard.modules.append(modtap)
 ## Custom HoldTap Behavior
 The full ModTap signature is as follows:
 ```python
-KC.MT(KC.TAP, KC.HOLD, prefer_hold=True, tap_interrupted=False, tap_time=None)
+KC.MT(KC.TAP, KC.HOLD, prefer_hold=True, tap_interrupted=False, tap_time=None, repeat=True)
 ```
 * `prefer_hold`: decides which keycode the ModTap key resolves to when another
   key is pressed before the timeout finishes. When `True` the hold keycode is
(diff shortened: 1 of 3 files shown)


kdb424 (approved):
Glad to see this making it's way in!

ChrisChrisLoLo:
This is awesome! Thanks for adding this feature in!

PR #613, opened 2022-10-07
Corrected Right Angle Bracket shortened version

alex-ries:
Fixed a minor typo where the right angle bracket shortened version said "GT" instead of "RT"

### DIFF
--- a/docs/keycodes.md
+++ b/docs/keycodes.md
@@ -152,7 +152,7 @@
 |`KC.COLON`              |`KC.COLN`          |`:`                |
 |`KC.DOUBLE_QUOTE`       |`KC.DQUO`, `KC.DQT`|`"`                |
 |`KC.LEFT_ANGLE_BRACKET` |`KC.LABK`, `KC.LT` |`<`                |
-|`KC.RIGHT_ANGLE_BRACKET`|`KC.RABK`, `KC.GT` |`>`                |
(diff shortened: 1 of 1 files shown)


xs5871:
Thanks for spotting that. Do `KC.RT` (and `KC.LT` for that matter) actually work? They're not defined in `keys.py`.

alex-ries:
> Thanks for spotting that. Do `KC.RT` (and `KC.LT` for that matter) actually work? They're not defined in `keys.py`.

Actually they do not even though there are no errors, for completeness I tried the other short hand as well, but that resulted in errors. I can confirm that the full versions LEFT_ANGLE_BRACKET,KC.RIGHT_ANGLE_BRACKET work fine

xs5871:
I would almost have suggested to add `KC.RT` and `KC.LT` to the key name definitions, but... `KC.LT` collides with the layer tap key. For consistencies sake it's probably best to remove both RT and LT from the docs, if there are no objections.

alex-ries:
I personally would like to see some shorthand to replace it :)

xs5871:
There are `KC.LABK` and `KC.RABK`, which are as shorthand as you can get while still being disambiguative.

alex-ries:
That makes sense I tested those and did they did not work initially. I triple checked and it is fine 😊

kdb424:
Any chance you'll update the PR to reflect this discussion? If not, I'll open a PR to resolve it as described.

xs5871:
Resolved by #632.

Issue #585, opened 2022-09-14
pystack exhausted when holdtap and combos are used together


ald0o:
Here is the debug trace:
```
Traceback (appels les plus récents en dernier) :
  Fichier "main.py", ligne 209, dans <module>
  Fichier "kmk/kmk_keyboard.py", ligne 440, dans go
  Fichier "kmk/kmk_keyboard.py", ligne 500, dans _main_loop
  Fichier "kmk/kmk_keyboard.py", ligne 271, dans _process_timeouts
  Fichier "kmk/modules/combos.py", ligne 170, dans <lambda>
  Fichier "kmk/modules/combos.py", ligne 272, dans on_timeout
  Fichier "kmk/modules/combos.py", ligne 287, dans send_key_buffer
  Fichier "kmk/kmk_keyboard.py", ligne 208, dans resume_process_key
  Fichier "kmk/kmk_keyboard.py", ligne 188, dans pre_process_key
  Fichier "kmk/kmk_keyboard.py", ligne 196, dans process_key
  Fichier "kmk/keys.py", ligne 493, dans on_release
  Fichier "kmk/modules/holdtap.py", ligne 148, dans ht_released
  Fichier "kmk/modules/holdtap.py", ligne 192, dans ht_activate_tap
  Fichier "kmk/kmk_keyboard.py", ligne 208, dans resume_process_key
  Fichier "kmk/kmk_keyboard.py", ligne 188, dans pre_process_key
  Fichier "kmk/kmk_keyboard.py", ligne 194, dans process_key
  Fichier "kmk/keys.py", ligne 479, dans on_press
RuntimeError: pystack exhausted
```

This happens when quickly pressing a combo key then a holdtap key belonging to a combo (provided the two keys fail to form a combo together). It can be triggered by pressing twice the same key, for instance, provided it's both a holdtap key and a combo component.

This issue is rather new and did not happen with master head from August 2022 (but I don't know precisely what is the first revision that triggers the crash).

ald0o:
It is not clear to me that there is an infinite recursion (well, I failed to find what could cause it). Might it be that CircuitPython's stack is just so small that it cannot handle 16 nested calls?

If there is indeed no infinite recursion, the cure could be to limit the stack height to a constant (independent of the numbers of modules using `resume_process_key`). This could be achieved by a standard trick such as using an event queue instead of handling the next module immediately in a nested call.

[...]

PR #615, opened 2022-10-08
fix pystack exhaust during resume_process_key.

xs5871:
resolves: #585

Instead of handling resumed key events in a deep stack, buffer them until the next main loop iteration. New resume events that may be emitted during handling of old resumes are prepended to that buffer, i.e. take precedence over events that happen deeper into the buffer/event stack. Logical replay order is thus preserved.

This obviously constitutes an architectural change to the way we handle buffered or deferred key events and is open for debate.
In my opinion it's cleaner and easier to debug than handling key events from wherever and whenever in the code, especially from within anonymous timeout callbacks. No more `keyboard._send_hid()` littered throughout modules. No more multiple key events during processing timeouts. One API to rule them all! (ok, well, almost)

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -5,6 +5,7 @@ except ImportError:
 
 from supervisor import ticks_ms
 
+from collections import namedtuple
 from keypad import Event as KeyEvent
 
 from kmk.consts import UnicodeMode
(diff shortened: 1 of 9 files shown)


kdb424 (approved):
I love this. I can't see why we wouldn't have done this in the first place. Seems much better designed 👍

PR #619, opened 2022-10-10
Fix behavior for hold-release-tap sequences

prokopiff:
There is a case with holdtap usage that feels broken or at least very sluggish.
For example, when the same key is used for `Space` and `Shift`, I would often find myself not having a space after a question mark, because the key was held for typing `?` and wasn't properly processed after it was released and quickly tapped again.
Same happens when I type in numbers on the layer activated by holding `Enter` and then try to submit entered values by tapping `Enter` again.


xs5871:
Are you using the repeat feature? Because that sounds like the repeat feature. You're kind of breaking half of it with that code.

prokopiff:
> Are you using the repeat feature? Because that sounds like the repeat feature. You're kind of breaking half of it with that code.

I use it a lot and it still works

xs5871:
I just submitted a unittest for the repeat feature in #620.
Your code does make the test fail. -- Under the assumption that you want the repeat to only apply to events that resolve to taps, but not to interrupts or holds, we can add a selector for that. Let me know if that's what you're looking for.

prokopiff:
I do believe that this is the correct behavior
- when hold is right after tap or hold it should repeat, because this is the only way to get the same as just holding a standard key
- when tap is right after hold it has to produce tap. I have many examples supporting it in the keymap I've been using with QMK for almost two years:
  - I use shift and space on same key, so when I'm done holding shift I want to be able to quickly get a space without waiting tap timeout
  - I use enter key for layer that has numbers, so when I enter some numbers in a spreadsheet I don't want to wait tap timeout to tap enter and start entering in the cell below
  - I use backspace for layer that has symbols, so when I enter a wrong symbol I want to delete it quickly without waiting tap timeout.

And now after I made this change I have exactly what I want, no issues for two days now.

xs5871:
Thanks for the confirmation. We'll make the repeat option multi-variate, so you can enable repeat for taps, but disable it for holds. Shouldn't take longer than a day or two.

xs5871:
@prokopiff if the generalized solution in #622, that you suggested, works for you; let us know, then we can close this PR.

prokopiff:
@xs5871 yes, it works great, thanks!

PR #621, opened 2022-10-14
fix typo in zodiark kb.py import

xs5871:


kdb424 (approved):
Whoops. This is going to need some work other than that, but thanks for being me to fixing that upstream at least.

PR #625, opened 2022-10-22
Enhancement global pointing device

xs5871:
We're currently handling pointing device HIDs completely independent and agnostic of one another.
That requires duplicate code and when using more than one PD HID module, they interfere with each other.

This is an attempt to consolidate and build a central API for axis movement.
Mouse keys can be handled just like keyboard and consumer control keys. Unfortunately that doesn't work with axes, because they have continuous states (the amount of movement), not binary (pressed/released). I'm still undecided about the exact implementation details and open to suggestions.

### DIFF
--- a/kmk/hid.py
+++ b/kmk/hid.py
@@ -5,6 +5,7 @@ from micropython import const
 from storage import getmount
 
 from kmk.keys import FIRST_KMK_INTERNAL_KEY, ConsumerKey, ModifierKey
+from kmk.utils import clamp
 
 try:
     from adafruit_ble import BLERadio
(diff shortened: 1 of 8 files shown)


kdb424:
I'm in favour of this change. Obviously it needs tested and finished, but over all, good thought.

xs5871:
@claycooper apologies... this is a big one.

claycooper (commented):
mouse_keys.py has some commented out print statements. Do you want to clean them up before merging or leave them for future testing?

xs5871:
From what I see it's the other way around. Old print comments have been removed.

claycooper (approved):
You're right. I switched apps before the final review and it must have swapped the order of the commits on me. Sorry about that.

PR #627, opened 2022-10-25
Update debounce docs

kdb424:
There was confusion about this, so hopefully this comment will help.

### DIFF
--- a/docs/en/scanners.md
+++ b/docs/en/scanners.md
@@ -30,7 +30,7 @@ class MyKeyboard(KMKKeyboard):
             row_pins=self.row_pins,
             # optional arguments with defaults:
             columns_to_anodes=DiodeOrientation.COL2ROW,
-            interval=0.02,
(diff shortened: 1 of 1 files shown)


docs/scanners.md
```diff
@@ -30,7 +30,7 @@ class MyKeyboard(KMKKeyboard):
             row_pins=self.row_pins,
             # optional arguments with defaults:
             columns_to_anodes=DiodeOrientation.COL2ROW,
-            interval=0.02,
+            interval=0.02,  # Debounce time
```
klardotsh, line 33: time in what unit? seconds? milliseconds? something else?

kdb424: One would assume ms I suppose. Unspecified. "Unit's of time" though for sure based on upstream's documentation https://learn.adafruit.com/key-pad-matrix-scanning-in-circuitpython/advanced-features

PR #628, opened 2022-10-25
docs(readme): remove scary banner about maintainership, reword chat blurb

klardotsh:


kdb424 (approved):
Approving this as long as we keep the looking for maintainers pinned or otherwise noted.

PR #647, opened 2022-11-21
fix: Include Z and 0 in CapsWords ranges

DBendit:
Before, the allowed alphabet of characters for CapsWords was exclusive of the letter Z, and the allowed number set was exclusive of the number 0.

Now, those characters are included, so using them won't interrupt CapsWords.


xs5871:
Makes sense and the unit test is much appreciated. The linter wants some formatting changes still.

DBendit:
My first thought was that this should be in the readme, and, heh, it already is :) I'll get it fixed up. Thanks for approving the workflow run.

xs5871 (approved):

Issue #655, opened 2022-11-29
Included boot.py incompatible with CircuitPython 8.0.0-beta3 and newer


DBendit:
https://blog.adafruit.com/2022/10/20/circuitpython-8-0-0-beta-3-released/

> The property `supervisor.runtime.next_stack_limit` replaces `supervisor.set_next_stack_limit()`

As a result of this change, the included boot.py produces the following error on startup:

```
Adafruit CircuitPython 8.0.0-beta.4 on 2022-10-30; WeAct Studio Pico with rp2040
Board ID:weact_studio_pico
UID:1262522004251630
boot.py output:
Traceback (most recent call last):
  File "boot.py", line 3, in <module>
AttributeError: 'module' object has no attribute 'set_next_stack_limit'
```

It should be pretty straightforward to inspect the supervisor module to see whether or not it has the `set_next_stack_limit()` method and use that to determine which means to use to set the stack limit.

Alternatively, is setting the stack limit necessary? In http://kmkfw.io/docs/boot, the example modified boot.py doesn't include the stack size modification at all, and at least in my own experience, works just fine. Would it make more sense to remove it altogether?

PR #656, opened 2022-11-30
fix: support CircuitPython 8.0.0-beta3

DBendit:
CircuitPython 8 changes the API of the `supervisor` module. This fix accounts for the old and new APIs without requiring any other configuration changes.

Fixes https://github.com/KMKfw/kmk_firmware/issues/655

Tested this on a pi pico with both 7.3.3 and 8.0.0-beta4.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -1,3 +1,6 @@
 import supervisor
 
-supervisor.set_next_stack_limit(4096 + 4096)
+if callable(getattr(supervisor, 'set_next_stack_limit', None)):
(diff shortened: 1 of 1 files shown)


xs5871:
In my opinion we should comment that out. It's an unreliable hack.
We've been working on reducing memory use and fragmentation; coincidentally there's #660, which further reduces our memory problems.

xs5871:
obsoleted by #698

PR #657, opened 2022-11-30
Adding Sub-Module Tri-layer-State to Layer module

doesntfazer:
# TRI_LAYER_STATE

```python
from kmk.modules.layers import Layers, Trilayerstate
keyboard.modules.append(Layers())
keyboard.modules.append(Trilayer())
```

#Keycodes

|Key         |Description                                                                    |
|-----------------|--------------------------------------------------------------------------|
|`KC.RAISE`      |Switches the default layer until the next time the keyboard powers off |
|`KC.LOWER`      |Momentarily activates layer, switches off when you let go              |

# Behavior

Tri-state layers is a feature that allows you to access a third layer by activating two other layers.

**Base layer is layer 0
LOWER is layer 1 
RAISE is 2 
ADJUST is 3**
****The only way to change this is to go into the module and adjust accordingly ***

# Without Tri Layer State
For most usecases, the way you would create a "Tri_Layer_State would be to have KC.MO(x) KC.MO(y) and KC.MO(z) on opposing layer keys. This works out fine in a lot of use cases. The issue you run into is if you press X, Then Y, to get to Z; then you release X, and continue holding Y, it stays in Z. To release from this layer you have to release Y.

# With Tri Layer State
With Tri_Layer_state you can set opposing LOWER and RAISE keys. With the example above, when you release X it will switch back to the state of Y. Each key acting independently.
With the way this is implimented in KMK (unlike in QMK) you can still access layer 3 indipendently.

# Limitations
There are some drawbacks at the moment, this does not support the layertap function at the moment. 

# Example code
```python
import board

from kb import KMKKeyboard

from kmk.keys import KC, make_key
from kmk.modules.trilayerstate import Layers

keyboard = KMKKeyboard()

keyboard.modules.append(Layers())
keyboard.modules.append(Trilayer())

keyboard.keymap = [
    [ # QWERTY
KC.A,     KC.B,     KC.C,
KC.LOWER, KC.SPACE, KC.RAISE,

     ],
    [ # LOWER
KC.N1,    KC.N2,    KC.N3,
KC.LOWER, KC.SPACE, KC.RAISE,

     ],
    [ # RAISE
KC.EXLM,  KC.AT,    KC.HASH,
KC.LOWER, KC.SPACE, KC.RAISE,

     ],
     [ # ADJUST
KC.F1,    KC.F2,    KC.F3,
KC.LOWER, KC.SPACE, KC.RAISE

     ]
]
```


kmk/modules/trilayerstate.py
```diff
@@ -0,0 +1,211 @@
...
+    
+    
+    
+    
+    
+    
+    
+    
+    
+
```
DBendit, line 210: That's a whole lotta newlines!

kmk/modules/trilayerstate.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+
+
+class LayerKeyMeta:
+    def __init__(self, layer, kc=None):
+        self.layer = layer
+        self.kc = kc
+
+
+class Layers(HoldTap):
```
DBendit, line 38: Would it make more sense to make this subclass the existing layers module, sort of like how ModTap subclasses HoldTap?

kmk/modules/trilayerstate.py
```diff
@@ -0,0 +1,211 @@
+'''One layer isn't enough. Adds keys to get to more of them'''
```
DBendit, line 1: Would be nice to see unit tests for this new module. At the very least, it'd illustrate how it's supposed to work.

DBendit (commented):
Also, make sure to run the linter as described in the project readme.

kmk/modules/trilayerstate.py
doesntfazer: I was thinking the same thing. That's sort of how I built it to begin with. I'll update tomorrow

docs/en/tri_layer_state.md
```diff
@@ -0,0 +1,111 @@
...
+# Without Tri Layer State
+For most usecases, the way you would create a "Tri_Layer_State would be to have KC.MO(x) KC.MO(y) and KC.MO(z) on opposing layer keys. This works out fine in a lot of use cases. The issue you run into is if you press X, Then Y, to get to Z; then you release X, and continue holding Y, it stays in Z. To release from this layer you have to release Y.
+
+# With Tri Layer State
+With Tri_Layer_state you can set opposing LOWER and RAISE keys. With the example above, when you release X it will switch back to the state of Y. Each key acting independently.
+With the way this is implimented in KMK (unlike in QMK) you can still access layer 3 indipendently.
+
+
+
+# Example code
```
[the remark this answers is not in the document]
doesntfazer: Absolutely

kmk/modules/trilayerstate.py
doesntfazer: Fixed.

kmk/modules/trilayerstate.py
doesntfazer: > Would it make more sense to make this subclass the existing layers module, sort of like how ModTap subclasses HoldTap?

Alright. I added as a subclass of layer. You're right, it does make more sense.

xs5871:
This functionality should be more generic, less hard-coded, and probably even part of the existing layer module.
We could have a list of layer combos that automatically resolve to a result layer, iff all of the "match" layers are active. That'll allow us to have more than one combo layer, more than two match layers, and many more special functions, such as changing the meaning of a layer switch depending on what default layer is active.
Example:
```
layers.combos = [
   # tri-layer:
   (match=(1,2), result=3),
   # quad-layer:
   (match=(1,2,4), result=5),
   # changing layer for different base layers:
   (match=(0,4), result=6),
   (match=(1,4), result=7),
   (match=(2,4), result=8),
]
```

doesntfazer:
Closing PR. Will make a new one for when I make this more modular.

PR #659, opened 2022-12-03
Fix cc hid report

### LINKED ISSUE #651: Support for 2-byte consumer control codes
Maybe I'm missing something but I get an error when trying to send the key `0x192`:
>   File "kmk/kmk_keyboard.py", line 491, in go
>   File "kmk/kmk_keyboard.py", line 557, in _main_loop
>   File "kmk/kmk_keyboard.py", line 106, in _send_hid
>   File "kmk/hid.py", line 113, in create_report
>   File "kmk/hid.py", line 183, in add_key
> OverflowError: value must fit in 1 byte(s)

According to evtest on another keyboard the calculator button send `0x192` - that fits to this description:
http://www.freebsddiary.org/APC/usb_hid_usages.php

According to qmk firmware the `KC_CALC` sends a `0xb2` but no matter if I create the key with `Key(code=178)` or `ConsumerKey(code=178)` it does not open a calculator while `0x192` seems to work in windows and linux.

Beside that it seems to be possible to send multi byte keys for other keyboards. Why KMK can't do it? 

I'm not really sure how to change the code to create a PR. When I understand it correct add_key iterates over a string byte for byte and if the byte is `0x00` it replaces the byte with the `key.code` - as my code is above 255 it fails...

My approach would be to just remember the index where to place it and replace it (I'm not coming from python so here is pseudo code):
```
replace <length of key.code> bytes in where_to_place starting from idx with key.code
```

But did I mention that I have no clue what is happening here as that is the first hid for me? it seems that `where_to_place` has a specific length and it could be that there is no more bytes left...

xs5871:
resolves #651
depends on #625

### DIFF
--- a/kmk/hid.py
+++ b/kmk/hid.py
@@ -200,13 +200,14 @@ class AbstractHID:
     def add_cc(self, cc):
         # Add (or write over) consumer control report. There can only be one CC
         # active at any time.
-        self._cc_report[1] = cc.code
+        memoryview(self._cc_report)[1:3] = cc.code.to_bytes(2, 'little')
         self._cc_pending = True
 
     def remove_cc(self):
         # Remove consumer control report.
-        if self._cc_report[1]:
-            self._cc_report[1] = 0x00
+        report = memoryview(self._cc_report)[1:3]
+        if report != b'\x00\x00':
+            report[:] = b'\x00\x00'
             self._cc_pending = True
 
     def add_pd(self, key):


kmk/hid.py
```diff
@@ -193,15 +190,40 @@ def add_key(self, key):
...
+    def add_cc(self, cc):
+        # Add (or write over) consumer control report. There can only be one CC
+        # active at any time.
+        memoryview(self._cc_report)[1:3] = cc.code.to_bytes(2, 'little')
+        self._cc_pending = True
+
+    def remove_cc(self):
+        # Remove consumer control report.
+        if self._cc_report[1]:
+            self._cc_report[1] = 0x00
```
tflori, line 208: should we not set all bytes to 0?

xs5871: That is absolutely correct; good catch.

xs5871:
No idea why the tests are failing. They don't even use CC.

tflori:
the thing is that the keys don't get removed now and the keys are send as pressed again.. `report = 0x0000` did not change the report.

added some debug messages:
```python
report = memoryview(self._cc_report)[1:3]
print(f'remove key: ${bytes(report)}')
if report != 0x0000:
    report = 0x0000
    print(f'removed: ${bytes(memoryview(self._cc_report)[1:3])}')
```

generates output like:
```
remove key: $b'\xea\x00'
removed: $b'\xea\x00'
```

the result is that the key never gets released

xs5871:
Thanks for looking into it.
I don't quite understand the "new" problem you mentioned in #651, but the latest push may also fix that.

tflori:
now it works! happy to help. thx

claycooper (approved):

PR #660, opened 2022-12-05
Reduce key dictionary memory footprint

xs5871:
Instead of indexing `Key` objects that have multiple names by each individual name, index by the set of names. This reduces the size of the default key dictionary by a factor of between 2 and 3, and as result also reduces reallocations/defragmentation.
Instead of instantiating all module/extension keys by default, append them to the `maybe_key`-generator list.

### DIFF
--- a/kmk/extensions/international.py
+++ b/kmk/extensions/international.py
@@ -1,35 +1,42 @@
 '''Adds international keys'''
 from kmk.extensions import Extension
-from kmk.keys import make_key
+from kmk.keys import KC, make_key
 
 
 class International(Extension):
     '''Adds international keys'''
 
     def __init__(self):
-        # International
-        make_key(code=50, names=('NONUS_HASH', 'NUHS'))
-        make_key(code=100, names=('NONUS_BSLASH', 'NUBS'))
-        make_key(code=101, names=('APP', 'APPLICATION', 'SEL', 'WINMENU'))
+        KC._generators.append(self.maybe_make_media_key)
 
-        make_key(code=135, names=('INT1', 'RO'))
-        make_key(code=136, names=('INT2', 'KANA'))
-        make_key(code=137, names=('INT3', 'JYEN'))
-        make_key(code=138, names=('INT4', 'HENK'))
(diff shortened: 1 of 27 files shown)


xs5871:
@klardotsh : maybe this fell through. Would you like to state an opinion?

mali1741:
I noticed there's still code that's not adapted to maybe-make-key things, like sequences for example.

xs5871:
If by sequences you're refering to `send_string`, correct. That is due for a refactor. If I missed some of the existing `KC.xxx` references, let me know.

xs5871:
Ah, ok. Fixed it.

claycooper:
Should we be concerned about the return of `make_key` changing the return type? I only see [one instance](https://github.com/KMKfw/kmk_firmware/blob/master/user_keymaps/klardotsh/klarank_featherm4.py#L83) in the codebase where the return of `make_key` is saved for later use in the keymap so it may not be a big deal except for people with custom modules/handlers in downstream forks.

xs5871:
That's an interesting point. Concerned, no,
* it's not really a publicly documented method,
* in the past it was a wrapper around key object instanciation that automatically indexed those keys in the global dictonary `KC`,
* the keymap you pointed out is a meme example and doesn't even use the auto-index functionality; `make_key` isn't even necessary there.

The key object generation code has gone through some radical changes in the last year. The code was even more convoluted then it is now. It's still easy to lose sight of how everything fits together.
While I don't see concern for compatibility reasons, that specific line of code isn't a nice solution. We could move the tuple return from `make_key` into the generators. This PR isn't the end of key code refactoring, and `make_key` may go away entirely at some point.

claycooper (commented):
Code ran without crashing, lowered memory as expected, and looks like a good move forward for improving keys.

claycooper (approved):

xs5871:
So, I did some further tests and investigations, and it appears that some of my assumptions about memory management are not correct. I'll demote this PR to draft for now, but probably close it and open a new one with a better solution.

PR #662, opened 2022-12-25
Removed duplicate quickpin.md and moved the newer version from docs to docs/en

moritz-john:
Hello,
I noticed that `quickpin.md` (as the only topic!) exists two times in the documentation:

1) in `kmk_firmware/docs/`:  https://github.com/KMKfw/kmk_firmware/blob/master/docs/quickpin.md

2) in `kmk_firmware/docs/en/`: https://github.com/KMKfw/kmk_firmware/blob/master/docs/en/quickpin.md

The first one is newer and added links to all the boards listed under `Supported footprints/boards`

___

I replaced current `quickpin.md` in `kmk_firmware/docs/en/` with the newer version from `kmk_firmware/docs/` after confirming in the [KMK Zulip chat](https://kmkfw.zulipchat.com/#narrow/stream/358347-general/topic/quickpin.2Emd.20exists.20twice.20in.20docs.3F/near/317449326) that having two `quickpin.md` files is not intentional.


xs5871 (approved):
Thanks!

PR #664, opened 2022-12-29
Update layers.md

SivanSrimanPC:
L87 to L90

1)the function declaration was missing the self argument, so added it.

2)i changed the color changing method call to just changing the hue parameter of the rgb object... this works perfectly with the breathing animation

### DIFF
--- a/docs/en/layers.md
+++ b/docs/en/layers.md
@@ -87,7 +87,8 @@ class Layers(_Layers):
 	def after_hid_send(self, keyboard):
 		if keyboard.active_layers[0] != self.last_top_layer:
 			self.last_top_layer = keyboard.active_layers[0]
-			rgb.hue = self.last_top_layer
+			rgb.set_hsv_fill(self.hues[self.last_top_layer], 255, 255)
+			rgb.hue = self.hues[self.last_top_layer]
 
 keyboard.modules.append(Layers())
 ```
--- a/docs/en/layers.md
+++ b/docs/en/layers.md
@@ -84,10 +84,10 @@ class Layers(_Layers):
 	last_top_layer = 0
 	hues = (4, 20, 69)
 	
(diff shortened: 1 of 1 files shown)


xs5871:
> 2)i changed the color changing method call to just changing the hue parameter of the rgb object... this works perfectly with the breathing animation

But it doesn't work with the static animation, because that one doesn't periodically update the LEDs.

SivanSrimanPC:
> > 2)i changed the color changing method call to just changing the hue parameter of the rgb object... this works perfectly with the breathing animation
> 
> But it doesn't work with the static animation, because that one doesn't periodically update the LEDs.

oh then i guess we can add this as a separate example for animations then
but for both cases the self argument needs to be passed

xs5871:
Easy fix: leave the `rgb.set_hsv_fill(...)` line in. I agree with the rest of the PR.

docs/en/layers.md
```diff
@@ -84,10 +84,10 @@ class Layers(_Layers):
 	last_top_layer = 0
 	hues = (4, 20, 69)
 	
-	def after_hid_send(keyboard):
+	def after_hid_send(self, keyboard):
 		if keyboard.active_layers[0] != self.last_top_layer:
 			self.last_top_layer = keyboard.active_layers[0]
-			rgb.set_hsv_fill(self.hues[self.last_top_layer], 255, 255)
+			rgb.hue = self.last_top_layer
```
xs5871, line 90:
```suggestion
			rgb.set_hsv_fill(self.hues[self.last_top_layer], 255, 255)
			rgb.hue = self.hues[self.last_top_layer]
```

xs5871 (changes requested):

xs5871:
There's a pending change request still open. Any updates?

xs5871 (approved):

PR #666, opened 2022-12-29
Adding ComboLayers to Layers Module (See #658)

doesntfazer:
## Combo Layers

Combo Layers is when you hold down 2 or more KC.MO() or KC.LM() keys at a time, and it goes to a defined layer.

By default combo layers is not activated. You can activate combo layers by adding this to your `main.py` file.  
The combolayers NEEDS to be above the `keyboard.modules.append(Layers(combolayers))`

```python
combo_layers = {
  (1, 2): 3,
   }
keyboard.modules.append(Layers(combo_layers))
```

In the above code, when layer 1 and 2 are held, layer 3 will activate. If you release 1 or 2 it will go to whatever key is still being held, if both are released it goes to the default (0) layer.  
You should also notice that if you already have the layers Module activated, you can just add combolayers into `(Layers())`

You can add more, and even add more than 2 layers at a time.

```python
combo_layers = {
  (1, 2): 3,
  (1, 2, 3): 4,
  }
```

## Limitations

Because of the way this feature works, you have to do the combo's in numerical order on the left side of the colon.

Example `(1, 2): 3` works. `(2, 1): 3` would not work. You can on the other hand do `(3, 4): 2` because the smaller number is to the right of the colon.  
You can't go to layer 0. so something like `(1, 2): 0` doesn't work.

### Limitations with more than 2 combo layers

Although adding more than 2 layers does work, (`(1, 2, 3): 0` for example) the functionality you get where it goes to whatever layer is still being held when another key is released, doesn't work.  
If you release the 3rd one, it will usually go to the second key press's layer if you continue to hold.  
Additionally, if you do use more than one layer combo, you should put anything that has more than 2 at the bottom.  
for example the following

```python
combo_layers = {
  (1, 2): 3,
  (1, 3): 4,
  (2, 3): 6,
  (1, 2, 3): 5,
  }
```

## Fully Working Example code

Below is an example of a fully working keypad that uses combo layers.

```python
print("Starting")

import board

from kmk.kmk_keyboard import KMKKeyboard
from kmk.keys import KC

combo_layers = {
  (1, 2): 3,
keyboard.modules.append(Layers(combo_layers))


keyboard = KMKKeyboard()


keyboard.keymap = [
    [ #Default
    KC.A,     KC.B  KC.C  KC.D,
    KC.E,     KC.F  KC.G  KC.H,
    KC.MO(1), KC.J, KC.K, KC.MO(2),
    ],
    [ #Layer 1
    KC.N1,    KC.N2, KC.N3, KC.N4,
    KC.N5,    KC.N6, KC.N7, KC.8,
    KC.MO(1), KC.N9, KC.N0, KC.MO(2),
    ],
        [ #Layer 2
    KC.EXLM,  KC.AT,   KC.HASH, KC.DLR,
    KC.PERC,  KC.CIRC, KC.AMPR, KC.ASTR,
    KC.MO(1), KC.LPRN, KC.RPRN, KC.MO(2),
    ],
        [ #Layer 3
    KC.F1,   KC.F2, KC.F3,  KC.F4,
    KC.F5,   KC.F6, KC.F7,  KC.F8,
    KC.MO(1) KC.F9, KC.F10, KC.MO(2)
    ]
    
]

if __name__ == '__main__':
    keyboard.go()
```


doesntfazer:
@xs5871 Here it is. And I got it to work with at least Quad Layer States, but the way the code is written it should work with as many as you would want.

xs5871:
This is certainly going into the right direction. There are a couple of minor things, mostly code style, that are immediately apparent to me.
In any case, we definitely want a unit test for this functionality. You don't necessarily have to do that yourself, I can chip in.

kmk/modules/layers.py
```diff
@@ -36,9 +36,13 @@ def __init__(self, layer, kc=None):
 class Layers(HoldTap):
     '''Gives access to the keys used to enable the layer system'''
 
-    def __init__(self):
+    def __init__(
+        self,
+        combolayers=None,
```
xs5871, line 41: general naming convention for variables is snake_case: `combo_layers`.

kmk/modules/layers.py
```diff
@@ -143,3 +165,41 @@ def _print_debug(self, keyboard):
         # debug(f'__getitem__ {key}')
         if debug.enabled:
             debug(f'active_layers={keyboard.active_layers}')
+
+    def get_combo_layer(self, keyboard):
+
+        acvlyrs = []
```
xs5871, line 171: Please use full english words for variable names. The 4 bytes of saved space aren't worth the inconvenience of decyphering it.

kmk/modules/layers.py
```diff
@@ -143,3 +165,41 @@ def _print_debug(self, keyboard):
...
+
+        # If the active keyboard.active_layers value is greater than 0, then add it to the acvlyrs list
+        if len(keyboard.active_layers) > 0:
+            for i in keyboard.active_layers:
```
xs5871, line 175: similare to the above: give it a proper, recognisable name (unless there's a reasonably common convention for a abreviation).
```suggestion
            for layer in keyboard.active_layers:
```

xs5871 (commented):
Here are some of coding style changes.

doesntfazer:
> This is certainly going into the right direction. There are a couple of minor things, mostly code style, that are immediately apparent to me. In any case, we definitely want a unit test for this functionality. You don't necessarily have to do that yourself, I can chip in.

I will for sure need some help with a unit test. I will also update the coding style as soon as I get some time. I appreciate it!

kmk/modules/layers.py
doesntfazer: fixed.

kmk/modules/layers.py
doesntfazer: Fixed

kmk/modules/layers.py
doesntfazer: Fixed this as well.

PetrStanik (approved):
``

doesntfazer:
@xs5871 coding style changes have been fixed. Wondering what needs to be done for the unit test. Never done one before.

xs5871:
Haven't forgotten about this, just didn't find the time to really look into unit tests for this.

xs5871:
Update: I started with the unit-test, found and fixed some problems, and now the build/test env is giving me a hard time.

xs5871:
Everytime you add tests and verify code, the development time quadruples. I had to make some tiny adjustments to your code.
Noteworthy fixes:
  * we can't assume that the default layer is layer number 0
  * we can't reorder or sort the layer list
  * code should also work with momentary, toggle, switch default, mod layer
  * we can't add functionality that introduces bugs which require a firmware restart (specifically: more than one combo active at a time)
 
Check it out in my private repo: [doesntfazer_combolayers](https://github.com/xs5871/kmk_firmware/tree/doesntfazer_combolayers)
You can merge that onto your PR.

doesntfazer:
> Everytime you add tests and verify code, the development time quadruples. I had to make some tiny adjustments to your code. Noteworthy fixes:
> 
> *   we can't assume that the default layer is layer number 0
> *   we can't reorder or sort the layer list
> *   code should also work with momentary, toggle, switch default, mod layer
> *   we can't add functionality that introduces bugs which require a firmware restart (specifically: more than one combo active at a time)
> 
> Check it out in my private repo: [doesntfazer\_combolayers](https://github.com/xs5871/kmk_firmware/tree/doesntfazer_combolayers) You can merge that onto your PR.

I appreciate you working on this. I've been sick with Covid. I will have to take a look at adding this to my PR when I am feeling better.

xs5871:
Any progress? I'd like to merge this at some point.

doesntfazer:
> Any progress? I'd like to merge this at some point.

Just took your changes and merged them. Sorry for the delay. Covid kicked my butt and then I forgot about it. Thanks for the reminder. Should be ready.

xs5871 (approved):

PR #668, opened 2023-01-02
add trackball rotation to example in docs

schaeggi:
Added how to set the rotation angle of the pimoroni trackball breakout to the existing example in the docs. 
It took me a while to find out, so I thought this information could be useful for others as well.


xs5871 (approved):
Seems reasonable enough.

PR #670, opened 2023-01-07
OS_LCTL_LSFT should be set in example code

tamelion:
The Chord example references OS_LCTL_LSFT which is not set in the code above it.


xs5871 (approved):
Good catch, thanks!

PR #672, opened 2023-01-08
Update peg_oled_display.py

regicidalplutophage:
It's been bugging me how this extension created i2c object with board.SDA and board.SCL.
My MCU board of choice doesn't have SDA and SCL in it's board module, yet somehow it works as long as SDA and SCL are defined in keyboard and i2c isn't(srsly, why do example boards all have i2c = board.I2C in their keyboard class? Lulu has it twice!)

Again, it somehow works without changes to the oled extension, I just fail to see how.
Anyway, here's a small tweak that also works, also as long as there's no i2c object, but it seems more correct?

### DIFF
--- a/kmk/extensions/peg_oled_display.py
+++ b/kmk/extensions/peg_oled_display.py
@@ -123,9 +123,9 @@ class Oled(Extension):
     def on_runtime_disable(self, sandbox):
         return
 
-    def during_bootup(self, board):
+    def during_bootup(self, keyboard):
         displayio.release_displays()
-        i2c = busio.I2C(board.SCL, board.SDA)
+        i2c = busio.I2C(keyboard.SCL, keyboard.SDA)
         self._display = adafruit_displayio_ssd1306.SSD1306(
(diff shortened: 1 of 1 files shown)


xs5871:
I agree with the name change, but because of a different reason: the object passed to `during_bootup` method is a keyboard object, not the board object from the CP builtin modules. That can be very confusing indeed. The reason why other examples use `board.SDA` and `board.SCL` is because those are aliases mapped to the respective numbered IO pins in CP for most boards. I could see how that would be less of a hurdle than reading a datasheet; personally I prefer pin numbers, because then I know which of the potentially multiple I2C buses I'm using, especially when hand-wiring.
In my opinion the I2C pins or optionally an initialized I2C interface should be passed during initialization, like in other extensions or modules. On the other hand, `during_bootup` is a good place to initialize peripherals and we currently don't have a mechanism that would allow doing that for interfaces with multiple endpoints (like I2C).

xs5871 (approved):

PR #673, opened 2023-01-10
fix: LED static mode toggling

Kryomatix:
Static animation now gets set to standby instead of none. The off() function has been moved so that it only only activates when toggled, similar to RGB extension

### DIFF
--- a/kmk/extensions/led.py
+++ b/kmk/extensions/led.py
@@ -230,7 +230,7 @@ class LED(Extension):
     def _key_led_tog(self, *args, **kwargs):
         if self.animation_mode == AnimationModes.STATIC_STANDBY:
             self.animation_mode = AnimationModes.STATIC
-        
+
         if self._enabled:
             self.off()
         self._enabled = not self._enabled


xs5871:
Change looks unproblematic enough that I'd be willing to yolo merge, once the linter is happy.

xs5871 (approved):

joeclarkin:
I guess I may not know the purpose of toggle but this still doesn't seem to work. 

If it's enabled and the led has a certain brightness, toggle will turn it off and set _enable to False. So now the led is off (internal brightness/duty cycle of 0).  Now toggle again and _enable is set to True, brightness is still 0 and can't be turned on again because nothing turns the brightness back up. How is that toggling?

What needs to happen is that if enabled and toggle, grab current brightness and store it. Then turn brightness to 0. After this, if toggled again, set brightness to stored value. This would work in a minimal way. 

I personally would expect that if brightness is 0, toggle would turn it on regardless of other state. So enabled+brightness 0 should propably toggle to last not 0 or a default brightness.

Kryomatix:
As far as I understand it, it already does cache brightness?

>    def off(self):
    self.set_brightness(0)

When it gets toggled off, this function gets run, setting the brightness to 0.

When it gets toggled back on, an animation mode gets run

>  def effect_static(self):
        self.set_brightness(self._brightness)
        # Set animation mode to standby to prevent cycles from being wasted
        self.animation_mode = AnimationModes.STATIC_STANDBY

It sets the brightness to self._brightness, which is the variable used to store the default brightness. 
From what I can tell at the moment, when it toggles on, it gets set to default brightness.

If you are desiring it to maintain any brightness adjustments after you toggle off and then on, then I think that the **step_brightness()** and **set_brightness()** would need to set **self.brightness** to whatever value they are passing to the **self.set_brightness()** function. I think that would help to preserve the brightness after toggling

I hope that helps

Issue #669, opened 2023-01-07
Stringy extension doesn't handle anything but KC.key.


CupricReki:
Looks like the stringy extension needs some love. The extension doesn't handle anything but KC.keys. 

https://github.com/KMKfw/kmk_firmware/blob/3238857e1e715b83edd8b249fc0ef2664b11dca4/kmk/extensions/stringy_keymaps.py#L20

Additionally, it doesn't use the Debug utility so it doesn't honor `StringyKeymaps.debug_enable = True`

xs5871:
What else should it handle? If I understand the conversation on zulip, you'd want it handle callable keys, like `KC.MO(1)`? That is almost certainly never going to happen. Callable keys "generated on the fly", i.e. aren't stored in the `KC` dictionary. There's no way to look them up, unless you put them in `KC` manually and reference them by an alias. It's also not reasonable to generate them via a string, because you'd have to write a parser that can distinguish between regular keys and callable ones, and I don't that's worth it. Stringy keymaps are considered more of a toy than core functionality anyway.

Agree with the debug issue though.

CupricReki:
The issue is that anything besides `KC.key` gets replaced with `KC.NO`. It makes the extension useless if you plan on using any  "argument" keys.

I can see how writing a parser with recursion would be time consuming. Would it be possible to just ignore anything that's not `KC.key` such as `KC.MO(1)`? I understand that nested keys would have to be non-stringy (`KC.key` instead of the stringy `key`) but it would at least make the extension usable for most applications.

xs5871:
Could you please provide a code example for illustration? The extension should only replace string objects in the keymap, not key objects. Mixing strings and regular key objects should already be handled correctly.
Defaulting to `KC.NO` for strings that don't correspond to a `KC.key` is the sensible thing to do, otherwise you'd crash the firmware when pressing that key.

CupricReki:
Well I'm a fool. My mistake was using expecting `'MO(1)'`  or `MO(1)` to work. `KC.MO(1)` works fine. Just as expected...

PR #676, opened 2023-01-13
Added more examples to the Stringy Keymaps documentation

moritz-john:
Following this issue: 
https://github.com/KMKfw/kmk_firmware/issues/669
[Zulip Chat Topic](https://kmkfw.zulipchat.com/#narrow/stream/358347-general/topic/Do.20layers.20MO.281.29.20work.20in.20StringyKeymaps.3F)

I felt the need to update the [Stringy Keymaps documentation](http://kmkfw.io/docs/extension_stringy_keymaps/).
In it's current form the documentation does not mention that it's possible to mix and match string-based key references and standard keycodes.

I added an example showcasing this.

___

The sentence
> Enables referring to keys by 'NAME' rather than KC.NAME.

may lead users to incorrectly assume it applies to layer keycodes like `KC.MO(layer)`, resulting in confusion as it needs to still be referenced using the notation of e.g. `KC.MO(1)` rather than `MO(1)`.

I also added a note for this.

### DIFF
--- a/docs/en/extension_stringy_keymaps.md
+++ b/docs/en/extension_stringy_keymaps.md
@@ -32,4 +32,4 @@ keyboard.extensions.append(stringyKeymaps)
 It should be noted that these are **not** ASCII. The string is **not** what
 will be sent to the computer. The examples above have no functional difference.
 
-When utilizing layer keycodes, such as `KC.MO(layer)`, it's not possible to use a string like `'MO(layer)'` instead employ the standard notation of e.g. `KC.MO(1)` in your keymap.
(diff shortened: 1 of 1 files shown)


xs5871:
There's a pending change request still open.

moritz-john:
Thank your for your reply.
> There's a pending change request still open.

May I ask which pending change request you are talking about?

xs5871 (changes requested):

xs5871:
I may have marked it as a question instead of a request for change. My bad.

moritz-john:
No problem, I applied your suggested change I think.
It's the first time I have to deal with requested changes so I apologize if I make a mistake.

xs5871 (approved):
Perfect, thank you and apologies for the confusion!

PR #681, opened 2023-01-22
Re-formatted layer key creation to be consistent

claycooper:
Key creation formatting in layer module wasn't consistent across keys in the module or with key creation in other parts of the codebase.


xs5871:
It is, however, consistent with the linters format rules. Your proposed change, though nicer to the human eye, unfortunately is not.

xs5871 (approved):
Clever use of punctuation.

claycooper:
Thanks for the merge. I'm not good with git or linting and was thinking that was a test commit. I wanted to attempt a squash merge once I got the linter to succeed.

xs5871:
The linter trips up everyone, no matter the experience. I could have rebase-squashed your commits in the github ui, but was a bit too trigger-happy on the merge button.

PR #687, opened 2023-01-24
Added Reviung34

daysgobye:
Also created Gtips folder and put all of gtips's boards in said folder.

I have built and tested this keyboard with this code.


xs5871:
Since there hasn't been any progress: PR has to pass tests before any review.
@daysgobye, you're not a new contributer... we have this exact same discussion _every single time_.

kdb424:
Looks like this can be closed as fixes are already in from #777

PR #689, opened 2023-01-25
added draculad keyboard

daysgobye:
added and tested draculad with blok controller.


MangoIV:
Hi, it would be nice if the information provided was correct. If it’s indeed Draculad which you ported, I’d like to ask you to mention my gh repo and provide proper information on where it’s sold. 

If it’s not Draculad, then please change the name.

kdb424:
Looks like this can be closed as it was fixed in #767

PR #692, opened 2023-01-31
 fixed tg4x and tested now works

daysgobye:


boards/tg4x/kb.py
```diff
@@ -7,25 +7,79 @@
...
+        52,
+        53,
+    ]
+    # all keys
+    #          coord_mapping = [
+    #  0 ,1 ,2 ,3 ,4 ,5 ,6 ,28,29,30,31,32,33,
+    #  7 ,8 ,9 ,10,11,12,13,35,36,37,38,39,
+    #  14,15,16,17,18,19,20,42,43,44,45,46,
+    #  21,22,23,25,26,50,51,52,53
+    #      ]
```
xs5871, line 85: In the past people have put
```python
# flake8: noqa
# fmt: off
```
into their `kb.py` files to disable linting for `coord_mapping` and format it the way it is in the comment. We can tell the linter to make an exception in general, just like we do for `main.py` (implemented in #710).

PR #695, opened 2023-02-08
Split KeyAttrDict.__cache into memory friendly partitions.

xs5871:
Instead of relying on the uncontrollable availability of a big chunk of contiguous memory for key caching, we can manually fragment the cache into reasonably small partitions. This will get rid of the out-of-memory-during-reallocation issues when `KC.__cache` grows. That means we can also finally purge the stack size cap from `boot.py`.

### DIFF
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -400,43 +400,59 @@ KEY_GENERATORS = (
 
 
 class KeyAttrDict:
-    __cache = {}
+    # Instead of relying on the uncontrollable availability of a big chunk of
+    # contiguous memory for key caching, we can manually fragment the cache into
+    # reasonably small partitions. The partition size is chosen from the magic
+    # values of CPs hash allocation sizes.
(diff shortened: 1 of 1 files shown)


claycooper (approved):
Looks good. Was the partition size of 37 a SWAG or was there some investigation that went into choosing it?

xs5871:
I tested with values between 8 and 73. There's not much difference between 37/47/59, but the exact optimal number depends on the number of keys to be stored. The smaller the partition, the better the chance of successfull allocation, and the less "wasted" space if it's just filled up over the last reallocation threshold. The bigger the partition, the faster the search and the smaller the cache list.
There is no particular reason for 37. I've been looking into the key deduplication again (no reason to have 5 aliases for key, if you only ever use one). If that's going somewhere, I'll probably test again. Maybe 97 is a better choice. Fits 60% keyboards (probably among the most commen) + ~30 extra keys.

PR #700, opened 2023-02-13
Without this, when I set keyboard.keymap I have random memory error.

checksummaster:
I'm not sure if I do a good thing, but without it, I can't run it on nice!nano without a memory error.
I need to bring back supervisor in boot.py and add a gc.collect() somewhere in keys construction.

In my code I initialize split keyboard with Bluetooth +
Add leds and oled display +
Add code for multimedia and mouse key.

I have two scanner (matrix and RotaryioEncoder)

Then set keymap with base, raise, lower set (64+4 key) and sometimes I have memory error. 

After this modification, just before calling keyboard.go,  gc.mem_free() report 15k

Note: I try to add a new board, avalanche (https://github.com/vlkv/avalanche)
Later (After fixing the memory problem, I will make a pull for the board.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -1,0 +1,3 @@
+import supervisor
+
+supervisor.set_next_stack_limit(4096 + 4096)
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -10,6 +10,7 @@ from kmk.consts import UnicodeMode
 from kmk.key_validators import key_seq_sleep_validator, unicode_mode_key_validator
 from kmk.types import UnicodeModeKeyMeta
 from kmk.utils import Debug
+import gc
 
 # Type aliases / forward declaration; can't use the proper types because of circular imports.
 Keyboard = object
@@ -424,6 +425,7 @@ class KeyAttrDict:
         if len(self.__cache[-1]) >= self.__partition_size:
             self.__cache.append({})
         self.__cache[-1][name] = key
+        gc.collect()
         return key
 
     def __getattr__(self, name: str):


xs5871:
First off: thank you for the report and looking into it.
What is the exact error you get and what do you mean by "random"?

Here's why I disagree with the proposed solution:
The requirements for contiguous memory for the keycode dictionary have already been drastically decreased. If you're getting allocation errors, then there's probably some piece of code that runs before the key dict allocation and heavily fragments memory.
The key dict is usually the last data structure that gets filled, so that's why it seems that memory problems show up during the keymap construction, and we're already running out of options to optimize that.
Spamming gc collection calls during key allocation as a good luck charm doesn't really fix the underlying issue, i.e. that no other part of KMK has gotten any attention in regards to memory management. Sooner or later we have to start looking into other parts of the code as well. Unfortunately, proper memory profiling on stock CP is nigh impossible. Just looking at free memory doesn't cut it.
The stack size limitation can cause and has repeatedly caused stack frame exhaustions. That should only be a last resort thing, if you're 100% certain that you're running your MCU on the very edge and need 10k more heap space. Not something we want as a  recommended default.

checksummaster:
I agree with you; spamming gc.collect everywhere is not a real solution.
But doing one before attributing keys was not enough for me. I was on the edge.

As you said, we should check before to save some memory.
I fix it by writing my own code for OLED display. 
The line "from adafruit_display_text import label" takes 8k.  (I think is because of its load font ?!?)
Now I use only images bitmap. And I have more than 4 sections (I will share the code later).
I think we can use text somehow because the "boot loader" of micropython writes some text on my display when I reboot, we should just use the same thing but I don't know how to do that yet.

I'm a little bit new to micropython (or just python).

If you don't mind, I will close it. Thank's

PR #704, opened 2023-02-16
Feature oneshot stacking

### LINKED ISSUE #427: Automatic Modifier Chaining for OneShot Keys

### LINKED ISSUE #535: Oneshot doesn't work with key taps sent on key release
The oneshot module holds down the given key until interrupted by another keypress, but it only releases when the other key releases, which is when holdtap and co send the tap key. As a result, the oneshot key is released just as the other key is being sent so modifiers do not get applied. The key is released on line 32:
https://github.com/KMKfw/kmk_firmware/blob/a1d8a990f131dc0c2a30e601983b3730a4790dc7/kmk/modules/oneshot.py#L32
I think this could be resolved by moving the release into a timeout, or by force-sending the key buffer or hid report or whatever before the release, but I am still not entirely sure about how all the internals work so that is why this is an issue instead of a pull request.

This is reproducable by using `KC.OS(KC.LSHIFT)` and `KC.HT(KC.A, KC.B)`. Tapping the oneshot key, then tapping the holdtap key produces a lowercase 'a', instead of 'A'.

xs5871:
resolves #427
resolves #535

Oneshot keys automatically stack and reset timeouts. I'm not going to implement an option to toggle that off, unless there's a real world application / request.

### DIFF
--- a/kmk/modules/oneshot.py
+++ b/kmk/modules/oneshot.py
@@ -1,9 +1,13 @@
 from kmk.keys import make_argumented_key
 from kmk.modules.holdtap import ActivationType, HoldTap, HoldTapKeyMeta
+from kmk.utils import Debug
+
+debug = Debug(__name__)
 
 
-def oneshot_validator(kc, tap_time=None):
-    return HoldTapKeyMeta(tap=kc, hold=kc, prefer_hold=False, tap_time=tap_time)
+class OneShotKeyMeta(HoldTapKeyMeta):
+    def __init__(self, kc, tap_time=None):
+        super().__init__(tap=kc, hold=kc, prefer_hold=False, tap_time=tap_time)
 
 
 class OneShot(HoldTap):
@@ -12,16 +16,31 @@ class OneShot(HoldTap):
     def __init__(self):
         super().__init__()
         make_argumented_key(
-            validator=oneshot_validator,
+            validator=OneShotKeyMeta,
             names=('OS', 'ONESHOT'),
             on_press=self.osk_pressed,
             on_release=self.osk_released,
         )
 
     def process_key(self, keyboard, current_key, is_pressed, int_coord):
-        '''Release os key after interrupting keyup.'''
+        '''Release os key after interrupting non-os keyup, or reset timeout and
+        stack multiple os keys.'''
+        send_buffer = False
+
         for key, state in self.key_states.items():
             if key == current_key:
+                continue
+
+            if isinstance(current_key.meta, OneShotKeyMeta):
+                keyboard.cancel_timeout(state.timeout_key)
+                if key.meta.tap_time is None:
+                    tap_time = self.tap_time
+                else:
+                    tap_time = key.meta.tap_time
+                state.timeout_key = keyboard.set_timeout(
+                    tap_time,
+                    lambda k=key: self.on_tap_time_expired(k, keyboard),
+                )
                 continue
 
             if state.activated == ActivationType.PRESSED and is_pressed:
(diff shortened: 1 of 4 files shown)


regicidalplutophage:
Does not work as expected.
I created momentary oneshot layer. I can combine oneshot mods on that layer with each other and anything on oneshot and base layers, but not with something on another layer. I think entering another layer "fires off" oneshot mods.

regicidalplutophage:
Tested the latest commit, yep, that solves it

xs5871:
Nice. Any volunteers to provide an update to the documentation?

regicidalplutophage (approved):
It does what it says on the tin 👍

PR #707, opened 2023-02-18
Enhancement: Makefile

xs5871:
Couple of minor enhancements to the `Makefile` that make my work easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,65 +102,44 @@ reset-board:
 	@echo "===> Rebooting your board (safe to ignore file not found errors)"
 	@-timeout -k 5s 10s $(PIPENV) run ampy -p /dev/ttyACM0 -d ${AMPY_DELAY} -b ${AMPY_BAUD} run util/reset.py
 
-ifdef MOUNTPOINT
-$(MOUNTPOINT)/kmk/.copied: $(shell find kmk/ -name "*.py" | xargs -0)
-	@echo "===> Copying KMK source folder"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -109,17 +109,31 @@ $(MOUNTPOINT)/kmk/.copied: $(shell find kmk/ -name "*.py" | xargs -0)
...
 endif
 
-copy-board: $(MOUNTPOINT)/kb.py
-$(MOUNTPOINT)/kb.py: $(BOARD)
-	@echo "===> Copying your board to kb.py"
-	@rsync -rh $(BOARD) $@
+ifdef MOUNTPOINT
+copy-board:
+	@echo "===> Copying your board to $(MOUNTPOINT)"
+	@rsync -urh $(BOARD)/*.py $(MOUNTPOINT)/
```
claycooper, line 131: Do you want to include a check that BOARD is defined here as well?

xs5871: "Want": no. I find that whole `ifdef` target guarding clunky and inelegant. I'd be surprised if there isn't a better way to do that and meant to look into that as well at some point. As there has been interest in a cross-compilation build chain recently it would make sense to revisit the Makefile.
I mean, you're not wrong, a similar guard clause is used for keymaps. I'll do it, but wish to express my disatisfaction with that design pattern.

claycooper: I agree, nested `ifdef`s are ugly.

claycooper (approved):

PR #709, opened 2023-02-20
Building mpy files.

daysgobye:
So I was doing some testing on how to improve boot times and ram use. I found the number one thing we can do is move kmk to mpy files. Being that it seems to be undocumented on how to do so with our make file (altho it works) I thought I would try and make it easy. 

In my testing just moving kmk to mpy files:
cut boot times in half 2.84106s->1.25781s

So whats in my pr?
mpy-cross-7: the adafruit distribution of mpy-cross at 7.0.0 
mpy-cross-8: the adafruit distribution of mpy-cross at 8.0.0
compiled/ : folder for mpy files
a new docker file that willl build kmk to .mpy files into the compiled folder
docker-compose.yml to handle passing in compiled folder and kmk its self (for quick local testing)
and last but not least is a new build workflow that will build on push and and create build artifacts for downloads.

Now this all works, but I think we could make it better by moving these build artifacts to a release. I have not done so because I dont know the best way for us to do so. We would end up with a lot of releases I think thats ok but maybe people dont agree. 
in my other projects it only does so on a push with a tag eg "7.9" that could be an option but I dont know if we can tag prs? then some one would have to make a push to master and tag the push.


xs5871:
Our documentation could profit from a section about general performance improvements, tips and tricks, including how to use the existing cross compilation build target. There's already a PR on the way that improves the build + copy-to-board process (see #707). In general I approve of recommending to use pre-compiled bytecode.

We're not going to add binary blobs to the repository though. That's in general not a great idea, not to mention that publishing copies of the mpy-cross binaries would be questionable from a licensing point of view to say the least, and "version controlling" bytecode that changes with every single commit is a bit silly. There's also no plan to have frozen releases any time soon.

PR #710, opened 2023-02-20
Add linter format exception for `kb.py`

xs5871:
Apply the the same linter format exceptions to `kb.py` as for `main.py`.


claycooper (approved):
Makes sense.

PR #711, opened 2023-02-20
Feature more mouse keys

xs5871:
Add keys for mouse buttons 4 and 5. CPs default pointing device HID report doesn't support more than that.

### DIFF
--- a/docs/en/mouse_keys.md
+++ b/docs/en/mouse_keys.md
@@ -15,6 +15,8 @@ keyboard.modules.append(MouseKeys())
 | `KC.MB_LMB`               | Left mouse button                    |
 | `KC.MB_RMB`               | Right mouse button                   |
 | `KC.MB_MMB`               | Middle mouse button                  |
+| `KC.MB_BTN4`              | mouse button 4                       |
+| `KC.MB_BTN5`              | mouse button 5                       |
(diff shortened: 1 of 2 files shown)


claycooper (approved):

ald0o:
I noticed this too late. But aren't buttons 4/5 more well known as "previous"/"next"? (as in "go back to previous place"/"return to next one"). At least this is how most web browsers, file managers, document viewers, ... act upon the press of this couple of buttons.

Also names Button 4 and Button 5 are particularly obscure for X11 users, since this environment does not number those buttons as 4 and 5 but rather as 8 and 9 (4/5/6/7 being reserved for vertical and horizontal wheels reinterpreted as discrete events).

For these two reasons I believe these two buttons deserve some more understandable aliases (such as BTN_PRV and BTN_NXT).

xs5871:
Well, yes, but also no.
Mouse buttons 4/5 are commonly mapped by modern OS's to something "prev/next", but that is not necessarily the case. Strictly speaking, that is also true for mouse buttons left/right/middle = 1/2/3.
This is one of the very few instances were I agree with QMKs naming decisions. We can add aliases though.
The fact that X11 does its own confusing thing, which it is objectively wrong about, doesn't count in opinion. It's been shooting itself in the foot for decades, let's not encourage it.

PR #714, opened 2023-02-22
Removed erroneous extra ".__iter__()" in KeyAttrDict iterator

ald0o:
I propose to remove the extra ".__iter__()" in KeyAttrDict iterator: iterating elements in the partition already implicitly calls its __iter__ method.

Actually when I use the dictionary iterator with this extra call, it triggers an error since an iterator does not have an iterator. This is now no longer the case.

For the story, I had to use this iterator to copy the dictionary and stumbled upon this issue.
Most likely this iterator is not being used in current code (or if it is, how could it be working?).

### DIFF
--- a/kmk/keys.py
+++ b/kmk/keys.py
@@ -435,7 +435,7 @@ class KeyAttrDict:
 
     def __iter__(self):
         for partition in self.__cache:
-            for name in partition.__iter__():
+            for name in partition:
                 yield name
 
     def __setitem__(self, name: str, key: Key):


xs5871:
For context:
The `KeyAttrDict.__iter__()` is used by the unit tests when debug output is on. Previously, when there was only a single dict as cache, it made sense to pass back the cache iterator instead of implementing the loop in python. That did not work in Circuitpython (and wasn't meant to), because unlike CPython, Circuitpython doesn't implement the `__iter__` method at all.

So, while this works "as intended", if you have a use for that functionality in KMK, I don't object to the change.

xs5871 (approved):

ald0o:
Well, I do copy the keymap in CircuitPython for a not very good reason (I'd probably achieve my goal with another method if I had to start over). Nonetheless it's certainly a good thing that all the firmware code can be executed inside the environment it is primarily targeting!

Issue #649, opened 2022-11-22
Pimoroni Trackball starts as clicked upon connection


slabua:
I've been having an issue with the trackball. Whenever I connect the keyboard to the pc, the trackball starts in the clicked status. Once I click on it it gets released and I can start using the keyboard normally.
Has anyone encountered the same problem?

xs5871:
No, but the pointing device HID handling is due for a refactor anyway, see #625.

slabua:
@xs5871 thank you! I will keep my source up to date and see what happens.

PR #715, opened 2023-02-24
Refactor pimoroni trackball

xs5871:
Various memory and performance improvements.
resolves: #649

### DIFF
--- a/kmk/modules/pimoroni_trackball.py
+++ b/kmk/modules/pimoroni_trackball.py
@@ -6,6 +6,7 @@ from micropython import const
 
 import math
 import struct
+from adafruit_pixelbuf import PixelBuf
 
 from kmk.keys import AX, KC, make_argumented_key, make_key
 from kmk.kmktime import PeriodicTimer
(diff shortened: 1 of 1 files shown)


claycooper (approved):

xs5871:
@regicidalplutophage you ment to ask some questions about optimization. This would be a good opportunity, before we merge.

regicidalplutophage:
> @regicidalplutophage you ment to ask some questions about optimization. This would be a good opportunity, before we merge.

My reluctance to review this stemmed from struggling to understand a good chunk of changes. I think gaining some experience and going through commits one by one helped

regicidalplutophage (approved):
I still don't understand the maths, but I'm fine with it.

xs5871:
Assuming you mean the rotation matrix: that's a common way to represent 2D rotations (or higher dimensions in computer graphics, though slightly modified). Fundamentally it's just trigonometry expressed in the terms of linear algebra. The important part is that we calculate the rotation coefficients only once during initialization, instead of every cycle. It breaks down to reducing the amount of floating point calculations, because they're usually not done in hardware.
Let me know if that was somewhat of a satisfactory breakdown, or if I missed your question completely.

regicidalplutophage:
The scary trig was moved to ```during_bootup```, got it

xs5871:
Not scary, just misunderstood.

PR #723, opened 2023-02-28
Add Piantor to boards

cairnm:


regicidalplutophage (approved):
Looks good!
I handle pins for right and left sides similarly in my own direct pin split.
I wonder if this method should be mentioned somewhere in the docs...

boards/beekeeb/piantor/kb.py
```diff
@@ -0,0 +1,52 @@
...
+import digitalio
+
+from storage import getmount
+
+from kmk.kmk_keyboard import KMKKeyboard as _KMKKeyboard
+from kmk.scanners.keypad import KeysScanner
+
+# VBus pin config
+VBUS_PIN = board.VBUS_SENSE  # RPi Pico
+# VBUS_PIN = board.GP29      # WeAct RP2040 + resistors on Piantor PCB
```
One-Nub, line 11: As a person with the WeAct on a Piantor, `board.GP29` results in an error (attribute not found), it appears as `board.A3` instead.

cairnm: Thanks! I've updated the comment to board.A3 instead of board.GP29

hxse:
When does this pull merge?

xs5871 (changes requested):

PR #722, opened 2023-02-26
oled improvement revival

after-ephemera:
This is meant to be strict revival of https://github.com/KMKfw/kmk_firmware/pull/548 so we can land these improvements. I do not have a kyria but [I expect to use the powersave disable feature on my lily58](https://kmkfw.zulipchat.com/#narrow/stream/358347-general/topic/oled.20powersave.20mode). 

If any features are still up for debate per the previous PR, I'm more than happy to try to pare it down to just the powersaving ones that I need. Thanks to @janlindblom for doing the heavy lifting here.

PR #726, opened 2023-03-02
Oled improvement

regicidalplutophage:
Hope @after-ephemera doesn't mind me taking over the  #722.

### DIFF
--- a/kmk/extensions/oled.py
+++ b/kmk/extensions/oled.py
@@ -147,8 +147,8 @@ class Oled(Extension):
             self.updateOLED(sandbox)
         return
 
-    def after_matrix_scan(self, keyboard):
-        if keyboard.matrix_update or keyboard.secondary_matrix_update:
+    def after_matrix_scan(self, sandbox):
+        if sandbox.matrix_update or sandbox.secondary_matrix_update:
             self.timer_time_reset()
             return
 
--- a/kmk/extensions/oled.py
+++ b/kmk/extensions/oled.py
@@ -1,0 +1,136 @@
+import busio
+import gc
+
+from kmk.handlers.stock import passthrough as handler_passthrough
+from kmk.keys import make_key
+import adafruit_displayio_ssd1306
+import displayio
+import terminalio
+from adafruit_display_text import label
+
+from kmk.extensions import Extension
+
+DISPLAY_OFFSET = 10
+
+
+class OledData:
+    def __init__(
+        self,
+        labels=None,
+    ):
+        if labels != None:
+            self.data = labels
+
+
+class Oled(Extension):
+    def __init__(
+        self,
+        views,
+        oWidth=128,
+        oHeight=32,
+        flip: bool = False,
+        device_address=0x3C,
+        brightness=0.8,
+    ):
+        displayio.release_displays()
+        self.rotation = 180 if flip else 0
+        self._views = views.data
+        self._width = oWidth
+        self._height = oHeight
+        self._prevLayers = 0
+        self._device_address = device_address
+        self._brightness = brightness
+        gc.collect()
+
+        make_key(
+            names=('OLED_BRI',), on_press=self._oled_bri, on_release=handler_passthrough
+        )
+        make_key(
+            names=('OLED_BRD',), on_press=self._oled_brd, on_release=handler_passthrough
+        )
+
+    @staticmethod
+    def oled_text_entry(x=0, y=0, text="", layer=None):
+        return {
+            0: label.Label(
+                terminalio.FONT,
+                text=text,
+                color=0xFFFFFF,
+                x=x,
+                y=y + DISPLAY_OFFSET,
+            ),
+            1: layer,
+        }
+
+    @staticmethod
+    def oled_image_entry(x=0, y=0, image="", layer=None):
+        odb = displayio.OnDiskBitmap(image)
+        return {
+            0: displayio.TileGrid(
+                odb, pixel_shader=odb.pixel_shader, x=x, y=y + DISPLAY_OFFSET
+            ),
+            1: layer,
+        }
+
+    def render_oled(self, layer):
+        splash = displayio.Group()
+        self._display.show(splash)
+        print(f"views={self._views}, layer={layer}")
+        for view in self._views:
+            if view[1] == layer or view[1] == None:
+                splash.append(view[0])
+        gc.collect()
+
+    def updateOLED(self, sandbox):
+        self.render_oled(sandbox.active_layers[0])
+        gc.collect()
+
+    def on_runtime_enable(self, sandbox):
+        return
+
+    def on_runtime_disable(self, sandbox):
+        return
+
+    def during_bootup(self, board):
+        displayio.release_displays()
+        i2c = busio.I2C(board.SCL, board.SDA)
+        self._display = adafruit_displayio_ssd1306.SSD1306(
+            displayio.I2CDisplay(i2c, device_address=self._device_address),
+            width=self._width,
+            height=self._height,
+            rotation=self.rotation,
+            brightness=self._brightness,
+        )
+        self.render_oled(0)
+        return
+
+    def before_matrix_scan(self, sandbox):
+        if sandbox.active_layers[0] != self._prevLayers:
+            self._prevLayers = sandbox.active_layers[0]
+            self.updateOLED(sandbox)
+        return
+
+    def after_matrix_scan(self, sandbox):
+        return
+
+    def before_hid_send(self, sandbox):
+        return
+
+    def after_hid_send(self, sandbox):
+        return
+
+    def on_powersave_enable(self, sandbox):
+        return
+
+    def on_powersave_disable(self, sandbox):
+        return
+
+    def _oled_bri(self, *args, **kwargs):
+        self._display.brightness = (
+            self._display.brightness + 0.1 if self._display.brightness < 0.9 else 1.0
+        )
+
+    def _oled_brd(self, *args, **kwargs):
+        self._display.brightness = (
+            self._display.brightness - 0.1 if self._display.brightness > 0.1 else 0.1
+        )
--- a/kmk/extensions/oled.py
+++ b/kmk/extensions/oled.py
@@ -116,7 +116,7 @@ class Oled(Extension):
 
     def during_bootup(self, board):
         displayio.release_displays()
-        i2c = busio.I2C(board.SCL, board.SDA)
+        i2c = busio.I2C(keyboard.SCL, keyboard.SDA)
         self._display = adafruit_displayio_ssd1306.SSD1306(
             displayio.I2CDisplay(i2c, device_address=self._device_address),
             width=self._width,
--- a/kmk/extensions/oled.py
+++ b/kmk/extensions/oled.py
@@ -65,19 +65,23 @@ class Oled(Extension):
         device_address=0x3C,
         brightness=0.8,
         brightness_step=0.1,
-        portrait: bool = False,
+        dim_time = 30,
+        off_time = 60,
     ):
         displayio.release_displays()
         self.rotation = 180 if flip else 0
-        #if portrait: self.rotation = self.rotation + 90
         self._views = views.data
-        self._width = width if not portrait else height
-        self._height = height if not portrait else width
+        self._width = width
+        self._height = height
         self._prevLayers = 0
         self._device_address = device_address
         self._brightness = brightness
         self._brightness_step = brightness_step
         self._timer_start = ticks_ms()
+        self._dark = False
+        self._go_dark = False
+        self._dim_time_ms = dim_time * 1000
+        self._off_time_ms = off_time * 1000
         gc.collect()
 
         make_key(
@@ -87,8 +91,9 @@ class Oled(Extension):
             names=('OLED_BRD',), on_press=self._oled_brd, on_release=handler_passthrough
         )
 
-    def render_oled(self, layer):
+    def render_oled(self, layer, *args, **kwargs):
         splash = displayio.Group()
+        splash.hidden = False if not self._dark else True
 
         for view in self._views:
             if view[3] == layer or view[3] is None:
@@ -145,18 +150,22 @@ class Oled(Extension):
         if sandbox.active_layers[0] != self._prevLayers:
             self._prevLayers = sandbox.active_layers[0]
             self.updateOLED(sandbox)
+        elif self._dark != self._go_dark:
+            self._dark = self._go_dark
+            self.updateOLED(sandbox)
         return
 
-    def after_matrix_scan(self, sandbox):
-        if sandbox.matrix_update or sandbox.secondary_matrix_update:
+    def after_matrix_scan(self, keyboard):
+        if keyboard.matrix_update or keyboard.secondary_matrix_update:
             self.timer_time_reset()
-            return
+        return
 
     def before_hid_send(self, sandbox):
         return
 
     def after_hid_send(self, sandbox):
         self.dim()
+        return
 
     def on_powersave_enable(self, sandbox):
         self._display.brightness = (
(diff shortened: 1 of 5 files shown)


regicidalplutophage:
@after-ephemera 
I think this is mostly done. Latest tweak made it pretty stable.
I will open it for reviews after making timed dimming and blacking out functions optional, meanwhile you can test it.
I'm not sure I can make waking up faster, but at least display waking up doesn't seem to mess anything up.

regicidalplutophage:
Due to OLED modules having limited brightness range, a better strategy for saving power would be to dim them sooner instead of reducing their brightness.
Variables ```powersave_dim_time``` and ```powersave_off_time``` are now exposed in constructor.

Text entry constructor has been changed in a way I think is more intuitive and easier to explain.

regicidalplutophage:
Ready for review!

xs5871:
Acknowledged. I'll get to it soon-ish, probably within the next 2 days.

after-ephemera:
Thanks, ack here too. I have been busy in the middle of a lot of life things but I'll come back around to testing soon!

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+class OledData:
+    def __init__(
+        self,
+        entries=None,
+    ):
+        if entries is not None:
+            self.data = entries
+
+    @staticmethod
+    def oled_text_entry(
```
xs5871, line 30: I don't get why this is a static method in a class that is basically just a list. Why not
```python
Oled(views = [
    OledTextEntry(...),
    OledImageEntry(...),
    ...], ...
)
```

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        direction='LTR',
+        line_spacing=0.75,
+        inverted=False,
+        layer=None,
+        side=None,
+    ):
+        _x = 0
+        _x_anchor = 0.0
+        _y_anchor = 0.0
+        if x_anchor == 'left' or x_anchor == 'L':
```
xs5871, line 45: I'd recommend to stick to just single letters, saves space and code, and matches the `direction` option from the adafruit lib.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+            2: y,
+            3: side,
+            4: layer,
+            5: OledEntryType.TXT,
+            6: _x_anchor,
+            7: _y_anchor,
+            8: direction,
+            9: line_spacing,
+            10: inverted,
+        }
```
xs5871, line 82: Why is this a dictionary? It should at least be a list, better a named tuple or object, so you can put names to the entries and not have to refer to the entries by some magic number.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        self._go_dark = False
+        self._powersave = False
+        self._dim_time_ms = dim_time * 1000 if dim_time else None
+        self._off_time_ms = off_time * 1000 if off_time else None
+        self._powersave_dim_time_ms = (
+            powersave_dim_time * 1000 if powersave_dim_time else None
+        )
+        self._powersave_off_time_ms = (
+            powersave_off_time * 1000 if powersave_off_time else None
+        )
```
xs5871, line 137: Default to `0` instead of `None` would spare you from the whole if-else stuff.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        make_key(
+            names=('OLED_BRD',), on_press=self._oled_brd, on_release=handler_passthrough
+        )
+
+    def render_oled(self, layer, *args, **kwargs):
+        splash = displayio.Group()
+
+        for view in self._views:
+            if self._dark is False:
+                if view[4] == layer or view[4] is None:
```
xs5871, line 152: Flatten out those guards. General rule of thumb: if your indentation has more than 4 levels, it probably needs a refactor.
```suggestion
            if self._dark:
                return
            if view[4] is not None or view[4] != layer:
                return
```

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+                    elif view[5] == OledEntryType.IMG:
+                        splash.append(
+                            displayio.TileGrid(
+                                view[0],
+                                pixel_shader=view[0].pixel_shader,
+                                x=view[1],
+                                y=view[2],
+                            )
+                        )
+        gc.collect()
```
xs5871, line 176: Did you verify that this gc call is necessary? `gc.collect()` is really, really slow for one, and at this point of the code I don't see you freeing up any memory anyway.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        self._display.show(splash)
+
+    def updateOLED(self, sandbox):
+        self.render_oled(sandbox.active_layers[0])
+        gc.collect()
```
xs5871, line 181: There may be some memory beeing freed here, but still. Verify that the code won't run without.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        flip_right: bool = False,
+        device_address=0x3C,
+        brightness=0.8,
+        brightness_step=0.1,
+        dim_time=None,
+        off_time=None,
+        powersave_dim_time=10,
+        powersave_off_time=60,
+    ):
+        displayio.release_displays()
```
xs5871, line 115: why is this here?

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        i2c = busio.I2C(keyboard.SCL, keyboard.SDA)
+        self._display = adafruit_displayio_ssd1306.SSD1306(
+            displayio.I2CDisplay(i2c, device_address=self._device_address),
+            width=self._width,
+            height=self._height,
+            rotation=180 if self._flip else 0,
+            brightness=self._brightness,
+        )
+        self.render_oled(0)
+        return
```
xs5871, line 214: unnecessary return statement

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def before_matrix_scan(self, sandbox):
+        self.dim()
+        if self._dark != self._go_dark or sandbox.active_layers[0] != self._prevLayers:
+            self._dark = self._go_dark
+            self._prevLayers = sandbox.active_layers[0]
+            self.updateOLED(sandbox)
+        return
```
xs5871, line 222: unnecessary return statement

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def after_matrix_scan(self, keyboard):
+        if keyboard.matrix_update or keyboard.secondary_matrix_update:
+            self.timer_time_reset()
+        return
```
xs5871, line 227: unnecessary return statement

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        displayio.release_displays()
+        i2c = busio.I2C(keyboard.SCL, keyboard.SDA)
+        self._display = adafruit_displayio_ssd1306.SSD1306(
+            displayio.I2CDisplay(i2c, device_address=self._device_address),
+            width=self._width,
+            height=self._height,
+            rotation=180 if self._flip else 0,
+            brightness=self._brightness,
+        )
+        self.render_oled(0)
```
xs5871, line 213: Is `during_bootup` the right place to start rendering? Shouldn't we wait for the rest of the firmware to be initialized?

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+    def after_hid_send(self, sandbox):
+        return
+
+    def on_powersave_enable(self, sandbox):
+        self._powersave = True
+        return
+
+    def on_powersave_disable(self, sandbox):
+        self._powersave = False
+        return
```
xs5871, line 241: unnecessary return statement

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def _oled_bri(self, *args, **kwargs):
```
xs5871, line 243: You can give this a proper full name. The `*args, **kwargs` doesn't do anything. Do we still forcefully pass empty argument lists to key handlers?
```suggestion
    def _oled_brightness_increase(self):
```

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        self._display.brightness = (
+            self._display.brightness + self._brightness_step
+            if self._display.brightness + self._brightness_step <= 1.0
+            else 1.0
+        )
```
xs5871, line 248: There's a `clamp` function in `kmk.utils` for that.
```suggestion
        self._display.brightness = clamp(self._display.brightness + self._brightness_step, 0, 1)
```

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        self._brightness = self._display.brightness  # Save current brightness
+
+    def _oled_brd(self, *args, **kwargs):
+        self._display.brightness = (
+            self._display.brightness - self._brightness_step
+            if self._display.brightness - self._brightness_step >= 0.1
+            else 0.1
+        )
+        self._brightness = self._display.brightness  # Save current brightness
```
xs5871, line 257: Same as `_oled_bri`.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+                self._display.brightness = (
+                    self._display.brightness / 2
+                    if self._display.brightness > 0.5
+                    else 0.1
+                )
+            else:
+                self._display.brightness = self._brightness
+                self._go_dark = False
+        elif self._off_time_ms is not None and not check_deadline(
+            ticks_ms(), self._timer_start, self._off_time_ms
```
xs5871, line 280: Check out `kmk.kmktime.PeriodicTimer`. Doesn't have a reset yet, but we can add that.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        ):
+            self._go_dark = True
+        elif self._dim_time_ms is not None and not check_deadline(
+            ticks_ms(), self._timer_start, self._dim_time_ms
+        ):
+            self._display.brightness = (
+                self._display.brightness / 2 if self._display.brightness > 0.5 else 0.1
+            )
```
xs5871, line 288: Again with the inline ifs. Makes it harder to read and doesn't even save loc, because the line gets too long.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        self._views = views.data
+        self._width = width
+        self._height = height
+        self._prevLayers = 0
+        self._device_address = device_address
+        self._brightness = brightness
+        self._brightness_step = brightness_step
+        self._timer_start = ticks_ms()
+        self._dark = False
+        self._go_dark = False
```
xs5871, line 128: If you call one `go_dark`, you might as well call the other `is_dark`.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        powersave_off_time=60,
+    ):
+        displayio.release_displays()
+        self._flip = flip
+        self._flip_left = flip_left
+        self._flip_right = flip_right
+        self._views = views.data
+        self._width = width
+        self._height = height
+        self._prevLayers = 0
```
xs5871, line 122: Plural name for singular data.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+        self._split = split
+
+        make_key(
+            names=('OLED_BRI',), on_press=self._oled_bri, on_release=handler_passthrough
+        )
+        make_key(
+            names=('OLED_BRD',), on_press=self._oled_brd, on_release=handler_passthrough
+        )
+
+    def render_oled(self, layer, *args, **kwargs):
```
xs5871, line 147: Method order: We usually prefer the required/inherited extension methods first, then privat ones. Also:
```suggestion
    def render_oled(self, layer):
```

xs5871 (changes requested):
First round! Yes there will likely be more than one.

Usually I'd ask to split the extension from the keyboard config, because those are different topics and should be organized in individual PRs. We can overlook that this time, there's enough to go over already.

kmk/extensions/oled.py
```diff
@@ -0,0 +1,291 @@
...
+    IMG = 1
+
+
+class OledData:
+    def __init__(
+        self,
+        entries=None,
+    ):
+        if entries is not None:
+            self.data = entries
```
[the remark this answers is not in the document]
regicidalplutophage: It was inherited from the PR I'm basing this off of.
Personally, to me this naming scheme makes a lot of sense:
```entries``` are single labels that we're plopping on the screen and relates to the way we use ```OledData``` class constructor;
```data``` is the collection of labels that we treat as one chunk, it matches the class ```OledData``` and a singular ```data``` consists of plural ```entries```.
I don't like the name ```views``` either, but there is some merit to that. To me it originally suggested that it's something like one frame, not in the sense of 1 frame of animation, but one screen's worth of stuff. Definitely the least justifiable name, could easily be ```data``` or ```entries``` as in data unpacked into singular entries.

File: kmk/extensions/oled.py (hunk 2 above)
Comment (regicidalplutophage) [reply kmk/extensions/oled.py:45]: Something left from the earlier iteration that I thought won't hurt. Removing that leaving only single letters.

File: kmk/extensions/oled.py (hunk 1 above)
Comment (regicidalplutophage) [reply kmk/extensions/oled.py:30]: Again, something inherited, will fix that.

File: kmk/extensions/oled.py (hunk 4 above)
Comment (regicidalplutophage) [reply kmk/extensions/oled.py:137]: Somehow I didn't think of doing that. Will change that.

File: kmk/extensions/oled.py (hunk 5 above)
Comment (regicidalplutophage) [reply kmk/extensions/oled.py:152]: That's something I just learnt, thanks!

File: kmk/extensions/oled.py (hunk 6 above)
Comment (regicidalplutophage) [reply kmk/extensions/oled.py:176]: In fact I tried running it without ```gc.collect()``` and it went fine, but then I thought if the original PR had it, it probably had a reason to include it. I had suspicion one of them were redundant and was actually hoping you'd point that out in review.

File: kmk/extensions/oled.py (hunk 8 above)
Comment (regicidalplutophage) [reply kmk/extensions/oled.py:115]: An artifact from the base PR, removed!

File: kmk/extensions/oled.py (hunk 12 above)
Comment (regicidalplutophage) [reply kmk/extensions/oled.py:213]: If fits the following criteria:
Runs without our input so that we render the initial state of OLED without waiting for changes.
Runs only once so we aren't drawing initial state to immediately change it.

I could put it in ```before_matrix_scan``` with some variable that flips to true after the initial state has been rendered but then it will run additional check on each cycle. Am I missing something?

File: kmk/extensions/oled.py (hunk 4 above)
Comment (regicidalplutophage) [reply kmk/extensions/oled.py:137]: Currently, accounting for '''None''' gives an option to set up OLED in s way that it never dims/turns off. You could make an argument that ```0``` should do that or there should be another parameter for that. I'm not sure what would be the best.

File: kmk/extensions/oled.py (hunk 5 above)
Comment (regicidalplutophage) [reply kmk/extensions/oled.py:152]: I managed to get rid of ```self._dark``` and ```self._go_dark``` and all the associated checks.
The
```
            if view[4] is not None or view[4] != layer:
                return
```
Doesn't really work as I want it to keep going through ```self._views``` even if the clause returns ```False```

kmk/extensions/oled.py
regicidalplutophage: Changed that. 👍

kmk/extensions/oled.py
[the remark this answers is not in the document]
xs5871: It's confusing when the excact same data structure is refered to by many names _within a single file of code_. Personally, I think just `data` as a variable name is so void of meaning, its semantic inference is on par with `my_var`.
You can also name it `labels`, `sprites`, `tiles`, `widget`, ... or any sensible combination of that.

kmk/extensions/oled.py
xs5871: Choosing `0` for infinite timeout is a common design pattern.

kmk/extensions/oled.py
xs5871: I see, it only updates on layer changes.

One thing though: `0` isn't necessarily the active layer on boot.

regicidalplutophage: You're right. Defaulting prev. layer to ```None``` allows me to remove the first rendering from bootup.

Issue #719, opened 2023-02-25
Suggestion: Decouple (RGB) Keycodes & Libs from Extensions


moritz-john:
Hello,
Recently I created two KMK firmware.
One for the [KLOR](https://github.com/moritz-john/kmk-config-klor/tree/master/firmware) and one for the [Splaytoraid](https://github.com/moritz-john/kmk-config-splaytoraid/tree/master/firmware) keyboard.
 While creating both firmware I faced the same issue:

Let's say I want to ship my keymap with as many features as possible.
E.g.: I'd like to put `KC.RGB_TOG` on a layer by default.

I create my `kb.py` & `main.py` file and host them on GitHub.
Someone downloads both files, puts them on their keyboard and...
_it's not working..?_

Turns out the person did not use LEDs in their keyboard build (or just skimmed over the firmware files) so they decided not to uncommend the LED code in `main.py`:

```py
# from kmk.extensions.RGB import RGB
# rgb = RGB(keyboard.rgb_pixel_pin, num_pixels=22)
# keyboard.extensions.append(rgb)
```
which will output the following error:
`ValueError: Invalid key: RGB_TOG` [via picocom]

Your average user will only notice one thing: **my keyboard is not working.**
They will not be someone who has picocom installed and  be able to troubleshoot this issue.

That's my first reason for suggesting to decouple RGB keycodes from the actual extension.

Make it so that e.g.: `KC.RGB_TOG` does nothing when there is no acutal RGB code running.
Add an error message for e.g.: picocom.
But don't stop the whole keyboard from operating at all. 
That's an absolutely in-transparent design decision for the average end user.

___

Second point:

Let's say this user did some reading and managed to uncomment:
```py
from kmk.extensions.RGB import RGB
rgb = RGB(keyboard.rgb_pixel_pin, num_pixels=22)
keyboard.extensions.append(rgb)
```
well, the keyboard is still not working..?
`ImportError: no module named 'neopixel'` [via picocom]

Turns out the user forgot to add `neopixel.mpy` to their libs folder.

Because as soon as the `from kmk.extensions.RGB import RGB` gets imported, it tries to import [neopixel](https://github.com/KMKfw/kmk_firmware/blob/master/kmk/extensions/rgb.py#L113).
Once again this is completly invisible to your average end user. 
All they know is that their keyboard is not working.

Make it so that `RGB` only _tries_ to import `neopixel`. If it's not found throw an error message for picocom without stopping the whole keyboard from operating. 
[...]

PR #727, opened 2023-03-06
Add Splaytoraid to boards

moritz-john:


moritz-john:
I have decided to temporarily withdraw my pull request and instead, plan to submit a new one in the future. 
Given that work is already underway on https://github.com/KMKfw/kmk_firmware/issues/719, I believe it would be more beneficial to ship my keymap with additional functions enabled by default once the work is completed.

PR #730, opened 2023-03-06
Add boilerplate method for board configuration at boot time

xs5871:
Just an idea how to streamline (read: user-proof) common configurations that happen at boot time.
It this form, it would still have to be manually called in `boot.py`:
```python
import board
from kmk.bootcfg import bootcfg

# Check single pin connected to ground, maybe hardware on-board or direct pin
# Always enable serial console.
bootcfg(board.GP0, no_cdc=False)
# Check a matrix key at position whatever. Diode orientation is fixed from first to second parameter.
# Enable midi.
bootcfg(board.GP9, board.GP0, no_midi=False)
```

### DIFF
--- a/docs/en/boot.md
+++ b/docs/en/boot.md
@@ -1,65 +1,161 @@
 # boot.py
-`boot.py` lives in the root of your keyboard when mounted as a storage device. 
-There is a more detailed explanation in the [circuit python docs](https://docs.circuitpython.org/en/latest/README.html), 
+`boot.py` lives in the root of your keyboard when mounted as a storage device.
+There is a more detailed explanation in the [circuit python docs](https://docs.circuitpython.org/en/latest/README.html),
 however there are some common use cases for your keyboard listed here.
 
-## Hiding device storage
-You can hide your device from showing up as a USB storage by default (this can be overridden 
-at startup if desired, per the example code further down this page).
+
+## KMKs builtin boot configurator
+
+KMK ships with a handy boot configuration function that does all the hard work
+for you.
+The interface may change in the future, but there is a safety mechanism in
+place: if anything goes wrong, it'll boot into a mountable and debuggable
+configuration.
+
+
+###  Signature
 
 ```python
-storage.disable_usb_drive()
+from kmk.bootcfg import bootcfg
+
+bootcfg(
+    # required:
+    sense: [microcontroller.Pin, digitalio.DigitalInOut],
+    # optional:
+    source: Optional[microcontroller.Pin, digitalio.DigitalInOut] = None,
+    boot_device: int = 0,
+    cdc: bool = True,
+    consumer_control: bool = True,
+    keyboard: bool = True,
+    midi: bool = True,
+    mouse: bool = True,
+    storage: bool = True,
+    usb_id: Optional[tuple[str, str]] = None,
+    **kwargs,
+) -> bool
+```
+All optional parameters are set to reflect common Circuipython defaults, however
+they may differ from board specific defaults.
+
+#### Sense and Source
+`sense` and `source` both accept either uninitialized `Pin`s or `DigitalInOut`
+instances for maximum flexibility.
+The boot configuration is only applied if `sense` reads `True` or "high", and
+skipped if it reads `False` or "low".
+
+
+#### boot_device
+Boot HID device configuration for `usb_hid`, see the [usb_hid documentation](https://docs.circuitpython.org/en/latest/shared-bindings/usb_hid/index.html#usb_hid.enable)
+for details.
+
+
+#### cdc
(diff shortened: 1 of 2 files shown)


regicidalplutophage:
This idea decouples boot.py from the keyboard class, which is amazing.

I had a similar, but contradictory idea of ```boot,py``` pulling matrix position of ```KC.RESET``` so that once user presses it they can choose to hold it while keyboard reboots to enable serial and storage which is what I've set up on my own board manually.

xs5871:
Refactored `bootcfg`, ready for review.
* fewer double negatives
* all optional parameters default to CP defaults
* add more options for management of HID devices
* add documentation

kmk/bootcfg.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # Entries for cdc (REPL) and storage are intentionally evaluated last to
+    # ensure the board is debuggable, mountable and rescueable, in case any of
+    # the previous code throws an exception.
+    if not cdc:
+        import usb_cdc
+
+        usb_cdc.disable()
+
+    if storage:
```
regicidalplutophage, line 76: Should be ```if not storage:``` I believe.

kmk/bootcfg.py
```diff
@@ -0,0 +1,81 @@
...
+    pass
+
+import digitalio
+import microcontroller
+import usb_hid
+
+
+def bootcfg(
+    sense: [microcontroller.Pin, digitalio.DigitalInOut],
+    source: Optional[microcontroller.Pin, digitalio.DigitalInOut] = None,
```
regicidalplutophage, line 13: Calling it ```source``` is counter-intuitive, as diode would point away from it.

regicidalplutophage (changes requested):

kmk/bootcfg.py
xs5871: I get what you mean. It's the "source" of the test voltage, and that name is commonly used as the opposite of a sense input.
Maybe I didn't explain it well enough in the documentation. I'm open to suggestions.

regicidalplutophage: Most people only interact with diodes by the means of their firmware config and thus are only aware of their abstract direction.

I think ```sense``` and ```source``` should be described separately then, like:
```sense``` the GPIO pin of directly wired key, or anode side of a key in matrix;
```source``` the GPIO pin of cathode side of a key in matrix.

Cathode and anode aren't terms commonly used in keyboard community, but they are easily googlabe.

Comment (xs5871) [reply kmk/bootcfg.py:13]: The whole diode point is something I intentionally avoided in the parameter description (let alone the technical terms of anode and cathode) and tried to explain only the basic eletrical meaning, leaving diode direction and direct wiring to the examples.
If you think that's something that should be mentioned in both places, or at least mentioning the examples in the parameter description, I'll have another go at it.

Comment (regicidalplutophage) [reply kmk/bootcfg.py:13]: The way I see it:
Requiring user to know what ```low``` is is bad;
Requiring used to reference an example is also bad, examples should be strictly supplementary IMO;
Dropping terms like anode and cathode is bad also, but a quick google search yields a picture with a neat arrow pointing away from anode and to cathode. Maybe include that diagram? ```anode ->|- cathode```

Comment (xs5871) [reply kmk/bootcfg.py:13]: Because `sense` and `source` are topology agnostic. The boot config pin isn't necessarily in a matrix, probably even a directly wired button on the MCU board.
I don't want to clutter the interface with lots of special cases that add complexity down the line.
Your pseudocode is not wrong, it's just incomplete and blows up in size if _all_ scenarios have to be handled.
This is as simple an interface as I can think of, and it's really not difficult to understand in my opinion -- if the documentation is improved.

Anyway, there's now more detailed documentation with configuration reference table.

File: docs/en/boot.md [hunk 3]
```diff
@@ -1,65 +1,178 @@
...
+maximum flexibility.
+It's the "source" of the test voltage to be read by the sense pin.
+If `source` is an uninitialized `Pin`, it'll be configured as a "low" output; it
+wont be further configured if it is a `DigitalInOut`.
+
+Common matrix and direct pin configurations (see also the examples below):
+|diode_orientation |sense pin  |source pin |
+|`COL2ROW`         |column     |row        |
+|`ROW2COL`         |row        |column     |
+|direct pin        |direct pin |`None`     |
```
regicidalplutophage, line 61: I like the new explanation but this table isn't displayed right

regicidalplutophage (changes requested):

kmk/bootcfg.py
regicidalplutophage: Alright, that's fair.

regicidalplutophage (approved):

PR #736, opened 2023-03-09
Add deinit method to modules and extensions

xs5871:
Adds a `deinit` method that gets called when `keyboard.go()` terminates. Allows to implement cleanup, like hardware deinit, or writing persistant state to flash.


regicidalplutophage (commented):
love it!

PR #740, opened 2023-03-13
Fix layer tap interrupted

### LINKED ISSUE #739: tap_interrupted=True parameter does not work in context of LayerTap[BUG]
**Describe the bug**

The combination of ```prefer_hold=True``` and ```tap_interrupted=True``` is an extremely useful setting that allows to both minimize typos and doesn't feel as sluggish to use as ```prefer_hold=False```.
This behavior is default in ZMK which is en exemplary firmware in terms of "ergonomic" features.
Unfortunately, when applied in context of LayerTap, whether through ```KC.MT``` or ```KC.HT```, pressing and releasing a key within ```tap_time``` produces weird results.

Sidenote:
Since ```KC.MT``` is now just an alias for ```KC.HT```, I think it should be deprecated completely.

**To Reproduce**
* Setup and configuration of the affected parts of the firmware (avoid copy-pasting the entire configuration if possible)

```
TEST1 = KC.MT(KC.F, KC.MO(3), prefer_hold=True, tap_interrupted=True, tap_time=1000)
TEST2 = KC.HT(KC.F, KC.MO(3), prefer_hold=True, tap_interrupted=True, tap_time=1000)
TEST3 = KC.HT(KC.F, KC.LSFT, prefer_hold=True, tap_interrupted=True, tap_time=1000)
```

* Setup and configuration of peripherals
* Input: keys pressed, ...

1.
- ```TEST2``` is pressed
- ```KC.F``` is pressed
- ```KC.F``` is released within ```tap_time```
- ```TEST2``` is released
Result: stuck ```f```

2.
- ```TEST2``` is pressed
- ```TEST3``` is pressed
- ```TEST3``` is released within ```tap_time```
- ```TEST2``` is released
Result: ```f```, which is the tap output of ```TEST3``` located on the same layer as ```TEST2```

```TEST2```, ```TEST3``` and ```KC.F``` are all located on the same layer.
```TEST1``` is omitted for brevity as it produces the same result as ```TEST2```

**Expected behavior**
Expected results are:
1. ```enter``` located on layer 3
2. ```backslash``` located on layer 3

**Debug output**
If applicable, add debug output from the serial console to help explain your problem.

**Additional context**

xs5871:
resolves #739 and adds unit test for that scenario.

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -197,7 +197,7 @@ class KMKKeyboard:
             key = ksf.key
 
             # Handle any unaccounted-for layer shifts by looking up the key resolution again.
-            if ksf.int_coord in self._coordkeys_pressed.keys():
+            if ksf.int_coord is not None:
                 key = self._find_key_in_map(ksf.int_coord)
 
             # Resume the processing of the key event and update the HID report
--- a/kmk/modules/holdtap.py
+++ b/kmk/modules/holdtap.py
@@ -83,6 +83,11 @@ class HoldTap(Module):
             if state.activated != ActivationType.PRESSED:
                 continue
 
+            # holdtap isn't interruptable, resolves on ht_release or timeout.
+            if not key.meta.tap_interrupted and not key.meta.prefer_hold:
+                append_buffer = True
+                continue
+
             # holdtap is interrupted by another key event.
             if (is_pressed and not key.meta.tap_interrupted) or (
                 not is_pressed and key.meta.tap_interrupted and self.key_buffer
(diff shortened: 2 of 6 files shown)


regicidalplutophage:
That only fixes case 1 where we press a non-holdtap key within the tap time of a LayerTap key.
Does not fix case 2 where we press a holdtap within the tap time of modtap: the output is still not layer shifted and both keys produce tap output.

regicidalplutophage:
The #741 is not what caused the problem in the second case, although not specifying ```tap_time``` changed output from tapped key to held key. The expected output is layer shifted key.

regicidalplutophage (approved):
It doesn't fix layertap completely, but somewhat improves it. I think it can be merged for what it is.

xs5871:
How about now?

regicidalplutophage:
Nope, sorry

xs5871:
Did you set `tap_interrupted=False`? I tripple checked and all the unit tests and hardware testing show that it's working properly in any combination of scenarios.

regicidalplutophage (approved):
In combination with #743 resolves issues #739 and #741

Issue #741, opened 2023-03-14
Specifying tap_time for an individual keys prevents holdtap keys from tap interrupting each other [BUG]


regicidalplutophage:
**Describe the bug**
```
holdtap = HoldTap()
holdtap.tap_time = 500

LCTL = KC.HT(KC.F, KC.LCTRL, prefer_hold=True, tap_interrupted=True)
RCTL = KC.HT(KC.J, KC.RCTRL, prefer_hold=True, tap_interrupted=True)
```
This works like one would think it would.
```
LCTL = KC.HT(KC.F, KC.LCTRL, prefer_hold=True, tap_interrupted=True, tap_time = 500)
RCTL = KC.HT(KC.J, KC.RCTRL, prefer_hold=True, tap_interrupted=True, tap_time = 500)
```
This doesn't

**To Reproduce**
```
LCTL = KC.HT(KC.F, KC.LCTRL, prefer_hold=True, tap_interrupted=True, tap_time = 500)
RCTL = KC.HT(KC.J, KC.RCTRL, prefer_hold=True, tap_interrupted=True, tap_time = 500)
```
- Hold ```RCTL```, tap ```LCTL``` within tap time.
- The output would be ```f```
 
**Expected behavior**
Expected output should be ```ctrl+f```

**Debug output**
If applicable, add debug output from the serial console to help explain your problem.

**Additional context**

PR #743, opened 2023-03-15
Fix some old holdtap inconsistencies

xs5871:
Working on #740 I noticed a few inconsistencies with stacked holdtaps.


regicidalplutophage (approved):
In combination with #740 resolves issues #739 and #741

PR #752, opened 2023-03-23
Refactored the Layers Module

doesntfazer:
I was working out some issues with the combo's layers when I found that most of the code for the layers module was horribly inefficient. Rewrote almost everything. Will need a unit test if someone would like to do that?


doesntfazer:
Not sure why it failed testing, as it passed linting.

doesntfazer:
Found some glaring issues after this commit.

PR #760, opened 2023-04-16
Pull up resistor support in digitalio MatrixScanner

macdude95:
## Background
I am currently working on a split keyboard project that uses a PCF8575 I/O expander. I ran into an issue while using the digitalio MatrixScanner because the scanner currently hard-codes the `pull` value in the `pin.switch_to_input(...)` call. Many I/O expanders don't support Pull-down resistors, so I was blocked in my project. 

## Description of change
I made modifications to the `digitalio.py` file that allows the user to configure which type of pull resistor is being used. I added an optional `resistor_pull` parameter to the MatrixScanner constructor that defaults to the original hard-coded value of `digitalio.Pull.DOWN`. This value is used in the constructor as well as in the `scan_for_changes(...)` method in order to determine which values to expect from the pins being read.

### DIFF
--- a/kmk/scanners/digitalio.py
+++ b/kmk/scanners/digitalio.py
@@ -130,11 +130,10 @@ class MatrixScanner(Scanner):
                         row = oidx
                         col = iidx
 
-                    pressed = (
-                        not new_val
-                        if self.resistor_pull is digitalio.Pull.UP
-                        else new_val
-                    )
+                    if self.resistor_pull is digitalio.Pull.UP:
+                        pressed = not new_val
+                    else:
+                        pressed = new_val
                     self.state[ba_idx] = new_val
 
                     any_changed = True
--- a/kmk/scanners/digitalio.py
+++ b/kmk/scanners/digitalio.py
@@ -37,13 +37,13 @@ class MatrixScanner(Scanner):
         # https://github.com/adafruit/Adafruit_CircuitPython_MCP230xx/blob/3f04abbd65ba5fa938fcb04b99e92ae48a8c9406/adafruit_mcp230xx/digital_inout.py#L33
 
         if self.diode_orientation == DiodeOrientation.COLUMNS:
-            self.outputs = [
+            self.anodes = [
                 x
                 if x.__class__.__name__ == 'DigitalInOut'
                 else digitalio.DigitalInOut(x)
                 for x in cols
             ]
-            self.inputs = [
+            self.cathodes = [
                 x
                 if x.__class__.__name__ == 'DigitalInOut'
                 else digitalio.DigitalInOut(x)
@@ -51,13 +51,13 @@ class MatrixScanner(Scanner):
             ]
             self.translate_coords = True
         elif self.diode_orientation == DiodeOrientation.ROWS:
-            self.outputs = [
+            self.anodes = [
                 x
                 if x.__class__.__name__ == 'DigitalInOut'
                 else digitalio.DigitalInOut(x)
                 for x in rows
             ]
-            self.inputs = [
+            self.cathodes = [
                 x
                 if x.__class__.__name__ == 'DigitalInOut'
                 else digitalio.DigitalInOut(x)
@@ -66,6 +66,15 @@ class MatrixScanner(Scanner):
             self.translate_coords = False
         else:
             raise ValueError(f'Invalid DiodeOrientation: {self.diode_orienttaion}')
+        
+        if self.pull == digitalio.Pull.DOWN:
+            self.outputs = self.anodes
+            self.inputs = self.cathodes
+        elif self.pull == digitalio.Pull.UP:
+            self.outputs = self.cathodes
+            self.inputs = self.anodes
+        else:
+            raise ValueError(f'Invalid pull: {self.pull}')
 
         for pin in self.outputs:
             pin.switch_to_output()
--- a/kmk/scanners/digitalio.py
+++ b/kmk/scanners/digitalio.py
@@ -79,7 +79,7 @@ class MatrixScanner(Scanner):
 
         self._key_count = self.len_cols * self.len_rows
         initial_state_value = (
-            b"\x01" if self.pull is digitalio.Pull.UP else b"\x00"
+            b'\x01' if self.pull is digitalio.Pull.UP else b'\x00'
         )
         self.state = bytearray(initial_state_value) * self.key_count
 
--- a/kmk/scanners/digitalio.py
+++ b/kmk/scanners/digitalio.py
@@ -17,7 +17,7 @@ class MatrixScanner(Scanner):
     ):
         self.len_cols = len(cols)
         self.len_rows = len(rows)
-        self.resistor_pull = resistor_pull
+        self.pull = pull
         self.offset = offset
 
         # A pin cannot be both a row and column, detect this by combining the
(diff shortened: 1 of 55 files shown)


kmk/scanners/digitalio.py
```diff
@@ -11,11 +11,13 @@ def __init__(
         cols,
         rows,
         diode_orientation=DiodeOrientation.COLUMNS,
+        resistor_pull=digitalio.Pull.DOWN,
```
xs5871, line 14:
```suggestion
        pull=digitalio.Pull.DOWN,
```
Keep naming scheme consistent with `digitalio`.

kmk/scanners/digitalio.py
```diff
@@ -125,15 +130,19 @@ def scan_for_changes(self):
                         row = oidx
                         col = iidx
 
-                    pressed = new_val
+                    pressed = (
+                        not new_val
+                        if self.resistor_pull is digitalio.Pull.UP
+                        else new_val
+                    )
```
xs5871, line 137:
```suggestion
                    if self.resistor_pull is digitalio.Pull.UP:
                        pressed = not new_val
                    else:
                        pressed = new_val
```
An inline-if is harder to read and doesn't safe LOC compared to a regular if-statement.
alternatively:
```
pressed = new_val
if self.resistor_pull is digitalio.Pull.UP:
    pressed = not pressed
```

Also, I think it should be `if pull == digitalio.Pull.UP`, since semantically we're comparing values, not object identities.

xs5871 (changes requested):
Definitely a missing feature.

kmk/scanners/digitalio.py
macdude95: Good call on readability. Will incorporate that change.

For enum comparison, see this documentation:
>Enumeration members are compared by identity:
>```
>>>> Color.RED is Color.RED
>True
>```

xs5871 (changes requested):
You have to replace `resistor_pull` with `pull` everywhere for it to work, and also in the docs.

macdude95:
> You have to replace resistor_pull with pull everywhere for it to work, and also in the docs.

Sorry I totally intended to do this, but I forgot. I think I have everything fixed now.

regicidalplutophage:
If you're unsure what tests want from you, you can
```
make test
make fix-isort fix-formatting
```
in a codespace

regicidalplutophage (approved):

regicidalplutophage (changes requested):
There's a bug:
In this current iteration of the scanner diode direction determines whether a set of pins acts as inputs or outputs. This is not enough as now outputs can either source or sink current and depending on which it is, diode must point to or from it.

What can be done is:
- Diode direction determines whether a set of pins is meant to high or low;
- Pull direction determines whether a high or low set is inputs or outputs.

kmk/scanners/digitalio.py
```diff
@@ -69,14 +71,15 @@ def __init__(
...
+            pin.switch_to_input(pull=self.pull)
 
         self.rollover_cols_every_rows = rollover_cols_every_rows
         if self.rollover_cols_every_rows is None:
             self.rollover_cols_every_rows = self.len_rows
 
         self._key_count = self.len_cols * self.len_rows
-        self.state = bytearray(self.key_count)
+        initial_state_value = b'\x01' if self.pull is digitalio.Pull.UP else b'\x00'
+        self.state = bytearray(initial_state_value) * self.key_count
```
parallacks, line 82: Is there a reason you're using a bytearray here? I tried something like `self.state = ([self.pull is digitalio.Pull.UP for i in range(self.key_count)])` in place of line 81 and 82 and it seemed to have the same behavior.

parallacks: I should add, I'm not saying one or the other is right. I'm new to this repo and actively testing this with an MCP23018 IO chip. I've been fiddling with things trying to understand the code.

macdude95:
> If you're unsure what tests want from you, you can
> 
> ```
> make test
> make fix-isort fix-formatting
> ```
> 
> in a codespace

Thanks for the help. Sorry i'm fairly new to codespaces, kmk, and even github pull requests in general. I'll look into the bugs you are calling out.

kmk/scanners/digitalio.py
macdude95: I honestly didn't put any thought into it. This pull request is focused on something else, so I decided to leave it at a bytearray for consistency.

macdude95:
> There's a bug: In this current iteration of the scanner diode direction determines whether a set of pins acts as inputs or outputs. This is not enough as now outputs can either source or sink current and depending on which it is, diode must point to or from it.
> 
> What can be done is:
> 
> * Diode direction determines whether a set of pins is meant to be on anode or cathode side;
> * Pull direction determines which set of pins is inputs and which is outputs as well as where inputs and outputs are pulled and driven.

Honestly, as a relatively new keyboard dev, I'm struggling to wrap my head around this.
> * Pull direction determines which set of pins is inputs and which is outputs as well as where inputs and outputs are pulled and driven.

This part makes sense, and I can make the following change to check Pull instead of DiodeDirection to determine inputs and outputs (I might have it backward, I haven't tested yet):
```
        if self.pull == digitalio.Pull.Down:
            self.outputs = [
                x
                if x.__class__.__name__ == 'DigitalInOut'
                else digitalio.DigitalInOut(x)
                for x in cols
            ]
            self.inputs = [
                x
                if x.__class__.__name__ == 'DigitalInOut'
                else digitalio.DigitalInOut(x)
                for x in rows
            ]
            self.translate_coords = True
        elif self.pull == digitalio.Pull.Up:
            self.outputs = [
                x
                if x.__class__.__name__ == 'DigitalInOut'
                else digitalio.DigitalInOut(x)
                for x in rows
            ]
            self.inputs = [
                x
                if x.__class__.__name__ == 'DigitalInOut'
                else digitalio.DigitalInOut(x)
                for x in cols
            ]
            self.translate_coords = False
        else:
            raise ValueError(f'Invalid Pull: {self.pull}')
```

However, this part is confusing me:
> * Diode direction determines whether a set of pins is meant to be on anode or cathode side;

What would this change look like? Can you help me out a bit?

regicidalplutophage:
Sorry, I could've explained it better.
In short, whether a set of pins is an input or an output depends BOTH on ```diode_orientation``` and ```pull```.
I'll propose changes in a bit.

regicidalplutophage (approved):

regicidalplutophage:
This should be good. I don't have the hardware, but you can test it.
By the way, have you had issue with key chatter when using this?

xs5871 (approved):
Good enough. The digitalio scanner wasn't pretty in the first place, and this doesn't make it much worse, but it adds crucial functionality -- which is the important part.
Good work and thank you both.

macdude95:
Sorry for the delayed response. No i have not had any issues with key chatter with this as long as I make sure to put pull-up resistors on my input pins when setting the new "pull" value to "up"

regicidalplutophage:
So... You're adding discrete resistors along with the pullups inside the MCP?

macdude95:
>So... You're adding discrete resistors along with the pullups inside the MCP?

I'm not using the `mcp23017`. I'm using the `pcf8575`. I apologize for my lack of hardware knowledge, but maybe the PCF lacks the internal pull-up resistors that the MCP has. I could not get it to work without external pull-up resistors.

Also, I just finally got around to manually testing this. I think it all works great, but the cathodes/anodes are reversed i'm pretty sure.

regicidalplutophage:
Thank you for clarifying this. The PCF8575 does have actual pull up resistors that are always active. Your use of external pullups indicates the internal pullups of PCF are too weak to reliably return to key-down state, which is consistent with what Parallacks experienced with MCP.

Regarding diode direction, the diode direction would indeed be flipped after the changes I made.
As long as diode "arrow" points to a column in ```ROW2COL``` or "''ROWS''';
Or to a row in ```COL2ROW``` or '''COLUMNS''',
it works as intended.
If that's not the case and I made some mistake, I don't see it yet...

PR #764, opened 2023-04-20
Added quickpin support for Frood RP2040 Microcontroller

jasonj2232:
Added quickpin support for Frood RP2040 Microcontroller by 42keebs (https://github.com/piit79/Frood)

Differs from other controllers supported by quickpin because it has 5 extra pins on the bottom like the Elite Pi. Unlike Elite Pi, it is supported by circuitpython, although the firmware should work on the Elite Pi and similar microcontrollers as well.

### DIFF
--- a/kmk/quickpin/pro_micro/frood.py
+++ b/kmk/quickpin/pro_micro/frood.py
@@ -1,0 +1,33 @@
+import board
+
+pinout = [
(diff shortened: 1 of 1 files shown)


xs5871 (approved):
Thanks!

PR #778, opened 2023-05-01
Lily58 patches

hjw:
There are a few glaring problems in the lily58 related files included in the repository. The coord_mapping was wrong and the example keyboard.keymaps had errors, both in the default(boards) folder as well as the user_keymaps folder.

I fixed the files in the boards folder... at least as far as the maps were concerned and I added warning comments to the user contributed files.


xs5871:
When I look at that picture of a lily58, I don't see 5 keys in the bottom row.

hjw:
The key which you are seeing, that looks like it's on row 3 or 4 is actually wired on row 5.

xs5871:
Sorry for the late response. The layout of the `keymap` is supposed to be representative of the visual layout of the keyboard, not necessarily the electrical. In turn, the "old" keymap would be the preferred way and the `coord_mapping` needs adjusting to reflect that.

hjw:
ok. i will fix the mapping so that it works with a visually correct keymap and also look for someplace in the docs/wiki to note that this is how it should be done. (The release mapping does not work properly with the visually correct keymap)

xs5871 (approved):

xs5871 (approved):

PR #800, opened 2023-05-21
build zip of .mpy files with GitHub Action, resolve #270

### LINKED ISSUE #270: Use Github Actions to automatically generate mpy files of kmk core files
It's understood by most  in the CP community that libraries can be pre-compiled to mpy files and that would make the bootup of our keyboards faster as our controllers won't need to compile the core kmk code.

I have a ["kmk builder" repo](https://github.com/jpconstantineau/KMK_Builder) I created to generate the mpy files of all core files of KMK.  At least from a proof of concept, it's workable...  My current CP 7.0.0 board uses it.  I haven't measured how much faster bootup time is but it feels faster...

As I don't know all the details about the project, would a PR be welcome to get a GHA workflow to automatically generate the kmk mpy files into an artifact when a PR is merged in?

rrotter:
This is a fairly trivial GH Action that automatically runs `make compile` and uploads the zip (for an example of what this looks like, see Artifacts [here](https://github.com/rrotter/kmk_firmware/actions/runs/5035251976)). As configured this will only run on merge/push to the master branch. Only supports CircuitPython 8, would be easy to add other versions if need be.

This was requested in #270.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -1,0 +1,44 @@
+name: Build
+on:
+  push:
+    branches: ['main', 'master']
+env:
+  mpy-cross-path: "https://adafruit-circuit-python.s3.amazonaws.com/bin/mpy-cross/"
+  mpy-cross-name: "mpy-cross.static-amd64-linux-"
+  mpy-cross-8-release: "8.0.5"
+jobs:
+  Compile:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "A ${{ github.event_name }} event trigged build of branch ${{ github.ref }}, repository ${{ github.repository }}"
+      - name: Check out repository code
+        uses: actions/checkout@v3
+
+      - name: Cache mpy-cross
+        id: cache-mpy-cross
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-mpy-cross-8
+        with:
+          path: ${{ github.workspace }}/mpy-cross
+          key: ${{ env.mpy-cross-name }}${{ env.mpy-cross-8-release }}
+      - if: ${{ steps.cache-mpy-cross.outputs.cache-hit != 'true' }}
+        name: install mpy-cross
+        run: |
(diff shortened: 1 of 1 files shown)


xs5871:
Couple of questions:
* We would need to mention in documentation where to find and download that for it to be usefull to anyone, right?
* Building on _every_ push seems a bit wastefull, but I guess we could design some kind of check to only build and save if the source subtree got updated?
* I'm not familiar enough with github actions. Is it common practice to pull dependencies (i.e. download mpy-cross) for every build? Or is there a persistant container / venv solution?

rrotter:
- Yes, if this is merged we should add some help text explaining what these builds are and how to find/use them. The artifacts _are publicly discoverable_ (once there were several builds it would look much like [the ZMK build workflow](https://github.com/zmkfirmware/zmk/actions/workflows/build.yml)). I'll be happy to add something to the docs if that's the only blocker. Though, that might be better placed in a second PR. There would be some question where to put it and how to stage it (i.e. could be anything from making mpy files the new recommended install for new users, to a footnote on a page about nrf52 and it's small flash, etc)
- That would be tricky. The action only has the current commit checked out, no view into the history. Also one would need to check if `/kmk` changed since the last build, not the prev commit (since a push could contain any number of commits). I'm not sure if you're concerned about wasting CPU or storage space but:
  - To be clear this isn't on _every push_ it's set to run on every push _to the master branch_.
  - This uses 6-10 seconds of CPU time (compared to the Test action that takes about 90 seconds to run, and isn't restricted to one branch).
  - The zip is under 100K, and GitHub deletes them after 90 days (their policy, not my setting), and AFAICT storage is either limited to 500MB or unlimited because it's a public repo (the docs are a little confusing on this point). By my back-of-the-envelope calculations it would take over 50 builds every day to exhaust the 500MB quota before the 90 day timeout deletes the builds anyway.
- Good point. I updated the workflow to cache mpy-cross.

Latest force push cleans up the naming of the Action, the naming of the zip (uses short sha and not the full sha), and caches mpy-cross as mentioned above.

xs5871 (approved):
Looking forward to the doc PR discussion.

PR #804, opened 2023-05-22
fixed kmk AttributeError: 'PeriodicTaskMeta'

blazingh:
fixed kmk AttributeError: 'PeriodicTaskMeta' object has no attribute 'restart'
i was having problems onb my rp2040 board while using the mouse_keys
added the a restart function the PeriodicTaskMeta class in the scheduler file fixed my isue, i hope this helps


xs5871:
addressed by #806

PR #807, opened 2023-05-22
Fix create_task

b2ox:
Fixed error (PeriodicTaskMeta has no ph_key) in get_due_task() when create_task argument is PeriodicTaskMeta.


xs5871:
I'll look into it.

xs5871 (approved):
Yes, that's something we missed. Hasn't been used in code so far.

PR #823, opened 2023-06-09
Ignore all HoldTap related keys as Oneshot interrupts

xs5871:
I think there was a reason we didn't do this from the start. Can't remember it and all the unit tests pass, so...
Leaving this as a draft for now?


regicidalplutophage:

PR #833, opened 2023-07-05
Update lab68.py

r2-d2222:
make this old file work with newer versions of kmk which do not use the digitalio scanner by default.


xs5871:
Linter is unhappy: you can run `make test` to get the result locally and `make fix-isort fix-formatting` to do some automatic repairs.

PR #837, opened 2023-07-11
Fix sequences.py

regicidalplutophage:


xs5871 (approved):
Although I'm a bit reluctant to fall back on the legacy `add_key`, `remove_key`, `tap_key` interface, I guess we won't do better until we factored out all the sequencing/macro code into a module.
Especially `tap_key` will introduce weird side effects, because all the releases are handled only after every other key in the sequence is looped over.

PR #842, opened 2023-07-27
fix: exception when using combos with match_coord

scorp200:
when using combos with `match_coord = True` and debugging enabled `__repr__` would throw an exception.


xs5871 (approved):
Thanks!

PR #843, opened 2023-08-02
Remove 'here's  in the english docs

### LINKED ISSUE #713: Replace "here" links in the docs
There are a lot of "here" links in the docs, which make it impossible to skim the docs and see were these links are actually pointing to.
Example: Instead of "You can have a look [here](http://pcbheaven.com/wikipages/How_Key_Matrices_Works/)" -> "You can have a look at this explanation [how key matrices work](http://pcbheaven.com/wikipages/How_Key_Matrices_Works/".

grasegger:
I made some changes to hopefully close #713


xs5871 (approved):
Awesome, thanks!
We usually squash and/or rebase before merging. But of course you're totally free to do that on your end and force push if you want.

Issue #634, opened 2022-10-27
Provide an issue template that helps separate live support from actual known+reproable project issues


klardotsh:
This is IMO a failing of GitHub's UX, where "Issues" does not clearly mean "for end-users" or "for maintainers", but in general, we fall on the "for maintainers" end of the opinion spectrum here, and it's a bad first experience with the project to open an issue and immediately be told "well, thanks, but also, this is the wrong place, try Matrix/Discord".

Tangentially related is #437, as without *somewhere* that isn't an unending stream of text to sort through, we're missing an important mode of communication. I have more+new thoughts on that that I'll share on that issue next.

PR #849, opened 2023-08-06
Add some ble details to docs in regards to controller compatibility

### LINKED ISSUE #785: BLE on a pico w
I updated my pico w with the most recent circuit python version and I tried to use BLE with no success.

Has anyone had success with this?

For my implementation I followed the docs http://kmkfw.io/docs/ble_hid#enabling-ble

grasegger:
This PR is very opinionated, but as far as I can see it is out of scope for kmk to implement support for ble on certain, so I would suggest to 

close #785 

with this PR. I am gonna follow up with a PR to #634 to point questions about hardware support to zulip, but I did not want to polute this PR with two issues.

### DIFF
--- a/docs/en/Officially_Supported_Microcontrollers.md
+++ b/docs/en/Officially_Supported_Microcontrollers.md
@@ -62,3 +62,5 @@ What you'll need to have at minimum
 - CircuitPython
 - 256KB of flash storage
 - HID over USB and/or Bluetooth.
+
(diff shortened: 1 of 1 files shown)


xs5871:
Do you mean 785? 778 doesn't seems to be related.
I'd remove the recommendation for nrf. There are plenty of RP2040 based boards that support BLE. Pretty much anything that's not designed by the Raspberry Pi people.

xs5871 (approved):

PR #852, opened 2023-08-07
Splitkb kyria rev2 port

dasmikko:
This is a working kb.py for the rev2 board

### DIFF
--- a/boards/kyria/kb_v2.py
+++ b/boards/kyria/kb_v2.py
@@ -1,0 +1,67 @@
+from kmk.kmk_keyboard import KMKKeyboard as _KMKKeyboard
+from kmk.quickpin.pro_micro.sparkfun_promicro_rp2040 import pinout as pins
+from kmk.scanners import DiodeOrientation
+from kmk.scanners import intify_coordinate as ic
+from storage import getmount
+
+isLeftSide = False if str(getmount('/').label)[-1] == 'R' else True
+leftColPins = (
+        pins[6],
+        pins[7],
+        pins[8],
+        pins[9],
+        pins[10],
+        pins[11],
+        pins[12],
+        pins[13],
+    )
+leftRowPins = (pins[17], pins[16], pins[15], pins[14])
+
+rightColPins = (
+        pins[10],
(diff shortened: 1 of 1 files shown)


boards/kyria/kb_v2.py
```diff
@@ -0,0 +1,67 @@
+from kmk.kmk_keyboard import KMKKeyboard as _KMKKeyboard
+from kmk.quickpin.pro_micro.sparkfun_promicro_rp2040 import pinout as pins
+from kmk.scanners import DiodeOrientation
+from kmk.scanners import intify_coordinate as ic
```
xs5871, line 4: unused; remove this line.

boards/kyria/kb_v2.py
```diff
@@ -0,0 +1,67 @@
...
+from storage import getmount
+
+isLeftSide = False if str(getmount('/').label)[-1] == 'R' else True
```
xs5871, line 7: This only works if the split sides were set up to use the mountpoint name to determine left and right.
The universal and future-proof way would be to request the "split-side" of a board from the `split` module. That's not technically your responsibility. The split module doesn't yet offer that functionality as, for example, a utility function. It should though, and I'm using this opportunity to point that out.
I'm proposing to factor out the split side/location heuristic into a public auxiliary function within the split module. Then you can call that sometime during the initialization to figure out what column and row pins to use.

boards/kyria/kb_v2.py
```diff
@@ -0,0 +1,67 @@
...
+        pins[9],
+        pins[10],
+        pins[11],
+        pins[12],
+        pins[13],
+    )
+    row_pins = (pins[17], pins[16], pins[15], pins[14])
+    diode_orientation = DiodeOrientation.COL2ROW
+    data_pin = pins[1]
+    rgb_pixel_pin = pins[0]
```
xs5871, line 60: These members are legacy only and should not be used in new contributions. These pins are to be provided to the split and rgb modules directly.

xs5871 (changes requested):

PR #386, opened 2022-04-09
Refactor: move matrix code

xs5871:
Move the old KMK matrix scanning code into the scanner module collection and give it a nice name.

PR #853, opened 2023-08-10
Fix DiodeOrientation imports from matrix module

coolavery:
`DiodeOrientation` was moved into `kmk.scanners` in #386. Ran into an import that wasn't updated in the Sofle config, and found one more instance of it in another board config.


xs5871 (approved):
Thanks!

PR #855, opened 2023-08-18
Add Pico quickpin

Allstreamer:
Create Raspberry PI pico quickpin based on the [Pico Docs](https://www.raspberrypi.com/documentation/microcontrollers/raspberry-pi-pico.html). (Also tested on Pico H)


xs5871 (changes requested):
Pi Pico doesn't have a pro micro footprint. It would need a seperate footprint directory.

Allstreamer:
Not sure what would be the best structure.
Right now it's compatible with both the pico H & pico W so i was thinking of naming it pico but that would make the import:
```python
from kmk.quickpins.pico.pico import pinout
```
which feels a bit weird, 
i'm thinking of making it top level or making a raspberry pi folder?
IDK not sure if i should just keep the name as pico.

For Refrence:
```python
# Pico folder
from kmk.quickpin.pico.pico import pinout

# Top Level
from kmk.quickpin.pico import pinout

# Raspi folder
from kmk.quickpin.raspi.pico import pinout
# or
from kmk.quickpin.raspberry.pico import pinout
# or 
from kmk.quickpin.raspberrypi.pico import pinout
```

xs5871:
Of course it looks weird. quickpin is meant to translate physical pin names / mappings from different board layouts. You're translating from the pi pico to the pi pico, which is kind of pointless imo... but if you think that's a good addition, then `kmk.quickpin.pico.pico` it is.

kdb424:
I've seen some pico clones, but AFIK, I haven't seen any that change the pinout's to the point of having a use for quickpin. The only use I can see for this is not having to find the pinout with an image for people reading the code, but it wouldn't have the intended primary use of quickpin.

xs5871:
I'll close this as stale.

PR #858, opened 2023-08-19
New spellcheck

grasegger:
Continuing the work on #425 

I made the script posix compliant, dropped the bash dependency and let implemented the shellcheck suggestions, thus all the problems @klardotsh talked about in the original PR should be taken care of.

I also updated the markdown search to exclude Japanese docs and made it work with the new docs structure.


xs5871 (changes requested):
Rename it "spellcheck.sh", that should do it for now.

xs5871 (approved):

Issue #859, opened 2023-08-20
[BUG] keybow2040: RGB causing infinite loop 


JimMadge:
**Describe the bug**
When using the example keybow2040 configuration, current CircuitPython and current kmk library, button presses have no effect. After a while CircuitPython soft resets.

Disabling the RGB, or forcing an exit from the `_process_timeouts` loop is a workaround.

**To Reproduce**
* Firmware 85658f6105a2ef2f2de63fa7850dfd851cc0cbd0
* Unmodified [keybow2040 files](https://github.com/KMKfw/kmk_firmware/tree/4b5d6a3133ae5de51918242a3dcbb33c4742c6b0/boards/pimoroni/keybow_2040)
* [Adafruit CircuitPython IS31FL3731 3.3.9](https://github.com/adafruit/Adafruit_CircuitPython_IS31FL3731/releases/tag/3.3.9)
* Plug in keyboard

**Expected behavior**
Keyboard responds to button presses.

**Debug output**
```
75001 kmk.keyboard: Unexpected error: cleaning up

Code stopped by auto-reload. Reloading soon.
soft reboot
```
repeatedly (different numbers)

**Additional context**

- When using the unmodified keybow2040 configuration, the RGB cycles colours but key presses have no effect.
- If you disable RGB (by commenting )
  https://github.com/KMKfw/kmk_firmware/blob/4b5d6a3133ae5de51918242a3dcbb33c4742c6b0/boards/pimoroni/keybow_2040/code.py#L15
  key presses work as expected
- Trying to find out what was happening, I found that `_process_timouts()` never returns.
  The loop does not end.
  https://github.com/KMKfw/kmk_firmware/blob/4b5d6a3133ae5de51918242a3dcbb33c4742c6b0/kmk/kmk_keyboard.py#L275-L277
  If you force this loop to exit early (say, by calling return after 10 iterations), key presses again work, but the RGB stops cycling.

So, I suspect that calls to update the RGB LEDs are filling the task queue and preventing the HID functions in `go` from being called.

I'm not very familiar with async but I'm happy to help debug if you have any suggestions.

JeffreyStocker:
I am having the same issue. I just started to play with RGB module and noticed that the keypad would become unresponsive to key presses when one of the more intricate RGB sequences were selected.

Ones tried
- KC.RGB_MODE_SWIRL
- KC.RGB_MODE_BREATHE_RAINBOW
- KC.RGB_MODE_RAINBOW

[...]

PR #864, opened 2023-08-23
increased RBG efficiency

JeffreyStocker:
Increased RGB efficiency by sending LED updates as a single command rather than updating individual updates
Separates LED auto write and manual LED updates into separate properties
Changed default auto write property to False to auto use more efficient LED update method

### DIFF
--- a/kmk/extensions/rgb.py
+++ b/kmk/extensions/rgb.py
@@ -105,9 +105,10 @@ class RGB(Extension):
         effect_init=False,
         reverse_animation=False,
         user_animation=None,
-        disable_auto_write=False,
+        disable_auto_write=True, #sets auto_write flag for newly created NeoPixal Objects
         pixels=None,
         refresh_rate=60,
+        manual_updates = False #sets if User must manually trigger updates with self.show()
     ):
         self.pixel_pin = pixel_pin
         self.num_pixels = num_pixels
@@ -132,6 +133,7 @@ class RGB(Extension):
         self.disable_auto_write = disable_auto_write
         self.pixels = pixels
         self.refresh_rate = refresh_rate
+        self.manual_updates = manual_updates
 
         self.rgbw = bool(len(rgb_order) == 4)
 
@@ -297,9 +299,6 @@ class RGB(Extension):
                     break
                 index -= len(pixels)
 
-            if not self.disable_auto_write:
-                pixels.show()
-
     def set_rgb_fill(self, rgb):
         '''
         Takes an RGB or RGBW and displays it on all LEDs/Neopixels
@@ -307,8 +306,6 @@ class RGB(Extension):
         '''
         for pixels in self.pixels:
             pixels.fill(rgb)
-            if not self.disable_auto_write:
-                pixels.show()
 
     def increase_hue(self, step=None):
         '''
@@ -456,6 +453,10 @@ class RGB(Extension):
             else:
                 self.off()
 
+            if not self.manual_updates:
+                #show pixals when not using manual updates
+                self.show()
+
     def _animation_step(self):
         self._substep += self.animation_speed / 4
         self._step = int(self._substep)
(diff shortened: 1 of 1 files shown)


xs5871:
I don't see how this improves anything. Can you go more into detail where the current code does "individual updates"?

JeffreyStocker:
Specifically, it is the property auto_write for NeoPixels and PixelBuf. 
`https://github.com/KMKfw/kmk_firmware/blob/642dbff31ba9913fcf9d69a0671359fe1a3f21d1/kmk/extensions/rgb.py#L209C5-L219C1`

auto_write set to True causes the board to send updates to the leds whenever any of the LED RGB values are updated. This means if the code is changing multiple LEDs in a animation step (such as RGB_MODE_RAINBOW or RGB_MODE_SWIRL) the code will send out a ton of updates which will clog up system resources. By turning auto_write to False, making updates, and the showing the update, the code will only send a single update to the LEDs. 

For example. as I stated in #859, when I implemented this on my Adafruit Macro pad (which only has 12 LEDs) I was limited to about 30 updates/second, before I started to get key press issues. By implementing my code, i was now up to 300ish. I would expect this to be even more noticeable performance gain for full keyboards with more LEDs.

xs5871:
So, as it turns out we did not set the `auto_write` of the PixelBuffers correctly.
Take a look at #874. That should have the same effect as your suggestion, but without any refactoring.
I would greatly appreciate your feedback on that PR.

JeffreyStocker:
No, you did use auto_write correctly. But the module's code confuses manual updates with RGB updating on writes. It is a good idea to keep some sort of manual update in the code in case someone makes more complicated RGB sequences. 

Also just deleting that property left in a ton of references to that property which has broken at least the breathing animation, probably more which I didn't test. The left over code also still caused too many updates, which means there was not any performance gains.

xs5871:
> No, you did use auto_write correctly. But the module's code confuses manual updates with RGB updating on writes. It is a good idea to keep some sort of manual update in the code in case someone makes more complicated RGB sequences.

`disable_auto_write` was probably intended to control `auto_write` (i.e. "manual updates") by the original author.
1. It did that only if the extension is initialized with GPIO pins, but not for a list of PixelBuffers.
2. Regardless of what the user set it to, it will be changed in the swirl and knight animations.
Both these things can result in inconsistent or at least unexpected behavior.

> Also just deleting that property left in a ton of references to that property which has broken at least the breathing animation, probably more which I didn't test. The left over code also still caused too many updates, which means there was not any performance gains.

Would you mind pointing out those dangling references? At what point are there still superflous updates?
I did check everything on hardware and can't reproduce what you found.

kdb424:
Auto writes were, in fact, disabled by default intentionally. If you look at animations, you'll see that you want to write out the entire animation that's to be processed, and force an update so that the animation shows up all at the same time. It was quite unintuitive to have to turn it off every time you wanted to batch things. I personally tried (on much weaker MCU's than we have these days) to see if there was a substantial performance difference in visuals or keyboard feel as I didn't have the tooling then to profile, and came to the conclusion that auto writes were, at least at that time, not more efficient, so I optimized for "set everything, then tell it to go all at once. Different users may have a different use case, but explaining the "how I got here" if that helps.

xs5871:
In order to go forwards I'll merge #874 and close this PR, leaving the initial issue open. It seems there're still some unaddressed performance issues.

JeffreyStocker:
I finally had some time to dig in a bit and I think I see what is happening

1) First you are right, there are no dangling bits of code. When I read your changes I thought that you had completely removed disable_auto_write from __init__, and just hadn't removed all the references, 

2) The code changes definitely helped, but as you said there are some more performance to be had. Mainly I'm seeing some of more complicated animation updating the pixels twice, much better than the per write than before. I'll poke at it and then do a separate merge

3) It looks like the Breathing problem I am having is a separate problem that has to do with #865 and was confusing the issue. I'll poke at that later as this a separate issue.

Because of the way disable_auto_write is being using use to control pixel updates rather than auto_writes on PixBufs, I think it should be renamed, but it isn't a big issue.

PR #885, opened 2023-09-18
use disable_auto_write with complex animations

JeffreyStocker:
Minor increase in RGB efficiency by preventing the more complex animations from updating the LEDs twice. Once in increase_hue and again in set_hsv or set_hsv_fill


xs5871:
`increase_hue` doesn't update the pixel buffer.

xs5871:
In addition: I was thinking about removing `disable_auto_update` altogether (#890).

JeffreyStocker:
yea that works. I'm just going to close this out then

PR #891, opened 2023-09-21
Added extra construction support to Encoder.py to support pullup direction from Code.py

Bob837217:
Added support to Encoder.py to allow GPIOEncoder to accept tuple for pin_button construction.  This includes allowance for pull resistor (pullup/pulldown) customisation directly from code.py.  Additionally the change preserves state generation semantics for pin_button.get_value(self) and depends upon the pull direction.  The default pull direction remains as digitalio.Pull.UP

An example config from code.py to customise the pin button pull up direction:
encoder_handler.pins = ((board.VOLC_A, board.VOLC_B, (board.VOLC_PUSH, True, digitalio.Pull.DOWN), False, 2),)

Tested this out on my own code.py.

### DIFF
--- a/kmk/modules/encoder.py
+++ b/kmk/modules/encoder.py
@@ -112,7 +112,15 @@ class BaseEncoder:
 
 
 class GPIOEncoder(BaseEncoder):
-    def __init__(self, pin_a, pin_b, pin_button=None, is_inverted=False, divisor=None, button_pull=digitalio.Pull.UP):
+    def __init__(
+        self,
+        pin_a,
+        pin_b,
+        pin_button=None,
+        is_inverted=False,
+        divisor=None,
+        button_pull=digitalio.Pull.UP,
+    ):
         super().__init__(is_inverted)
 
         # Divisor can be 4 or 2 depending on whether the detent
@@ -162,6 +170,7 @@ class EncoderPin:
         if digitalio.Pull.UP != io.pull:
             result = not result
         return result
+
 
 class I2CEncoder(BaseEncoder):
     def __init__(self, i2c, address, is_inverted=False):
--- a/kmk/modules/encoder.py
+++ b/kmk/modules/encoder.py
@@ -112,7 +112,7 @@ class BaseEncoder:
 
 
 class GPIOEncoder(BaseEncoder):
-    def __init__(self, pin_a, pin_b, pin_button=None, is_inverted=False, divisor=None):
+    def __init__(self, pin_a, pin_b, pin_button=None, is_inverted=False, divisor=None, button_pull=digitalio.Pull.UP):
         super().__init__(is_inverted)
 
         # Divisor can be 4 or 2 depending on whether the detent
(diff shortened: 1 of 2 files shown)


kmk/modules/encoder.py
```diff
@@ -122,6 +122,7 @@ def __init__(self, pin_a, pin_b, pin_button=None, is_inverted=False, divisor=Non
         self.pin_a = EncoderPin(pin_a)
         self.pin_b = EncoderPin(pin_b)
         self.pin_button = (
+            EncoderPin(*pin_button) if isinstance(pin_button, tuple) else
```
xs5871, line 125: Make that a proper named argument instead of overloading `pin_button` with a tuple.

kmk/modules/encoder.py
```diff
@@ -150,12 +152,14 @@ def prepare_pin(self):
...
-            self.io.pull = digitalio.Pull.UP
+            self.io.pull = self.pull
         else:
             self.io = None
 
     def get_value(self):
-        return self.io.value
+        return (
+            self.io.value if digitalio.Pull.UP == self.io.pull else not self.io.value
+        )
```
xs5871, line 162: if an inline ternary causes a linebreak, then please use a regulare `if` clause instead.

xs5871 (changes requested):

kmk/modules/encoder.py
Bob837217: This is not possible with how you have defined the pins config initially.  My concern is that there are users who have all kinds of config specified in their code.py as anonymous tuples of fixed length.  If I introduce an extra named argument, while this code would work - it would break your users configs in code.py.

Here is my config from code.py before any change to my PR:
`encoder_handler.pins = ((board.VOLC_A, board.VOLC_B, (board.VOLC_PUSH, True, digitalio.Pull.DOWN), False, 2),)
`
And here is my config after I introduce this proper named argument:
`encoder_handler.pins = ((board.VOLC_A, board.VOLC_B, None, (board.VOLC_PUSH, True, digitalio.Pull.DOWN), False, 2),)`

See how I have extended the length of this pins config tuple by one extra argument which is "None" for the original "pin_button" argument?
This will break users configs who have the regular tuple.  Because your code expects a tuple in this way and passes this directly into the constructor of GPIOEncoder like this:
`new_encoder = GPIOEncoder(*pins)
`

This is why I went with overloading "pin_button" with a tuple.
Is there anything else that may be done to fix this?  I don't want to break compatibility with user configs.  And my original PR does not break compatibility.

kmk/modules/encoder.py
Bob837217: resolved with [549e96d](https://github.com/KMKfw/kmk_firmware/pull/891/commits/549e96dd1c163418b86c2cc55019634ebb3ff000)

kmk/modules/encoder.py
xs5871: Here's a mockup of what I had in mind:
```python
class GPIOEncoder(BaseEncoder):
    def __init__(self, pin_a, pin_b, pin_button=None, is_inverted=False, divisor=None, button_pull=digitalio.Pull.UP):
        # ...
        if pin_button:
            self.pin_button = EncoderPin(pin_button, button_type=True, button_pull=button_pull)
        else:
            self.pin_button = None
```

Can be used like this:
```
encoder_handler.pins = (
    (board.GP0, board.GP1, board.GP2, False, 2, digitalio.Pull.DOWN),
    (board.GP3, board.GP4, board.GP5),
)
```

Doesn't break backwards compatibility, although we have to break it eventually to get rid of some of the old garbage code -- like guessing a constructor from the contents of a tuple, instead of using the proper constructor in the first place.

xs5871:
On another note: new features require documentation for a merge.

kmk/modules/encoder.py
Bob837217: Implemented, tested and documented the change

xs5871 (approved):

PR #897, opened 2023-09-26
fix adns9800 module: add missing inputs to AX.X.move and AX.y.move

db-ok:
Added missing keyboard input arguments to the AX.X.move and AX.Y.move function calls.

I have an adns9800 module running on a Pi Pico which I have tested with this fix on, and it's working in my local copy.

I had gotten the adns9800 module working with an older version of kmk, but the latest version of kmk was throwing errors when moving the trackball on the adns9800, and this change fixes the issue for me.

### DIFF
--- a/kmk/modules/adns9800.py
+++ b/kmk/modules/adns9800.py
@@ -203,10 +203,10 @@ class ADNS9800(Module):
                 delta_y *= -1
 
             if delta_x:
-                AX.X.move(delta_x)
+                AX.X.move(keyboard, delta_x)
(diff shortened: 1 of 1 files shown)


db-ok:
I put kmk onto my Pi Pico with an adns9800 sensor module that was working with an older version of kmk, but the latest version of kmk was throwing errors when moving the trackball on the adns9800, and this change fixes the issue for me.

xs5871 (approved):
Thanks!

PR #903, opened 2023-10-21
Fix off() method in RGB

regicidalplutophage:
Since the animate() method doesn't call `self.show()` when RGB is toggled off, it has to be called in `self.off()` for the latter to produce desired effect.


Xoffio:
this fixes issue #904

xs5871 (approved):

PR #912, opened 2023-11-09
Add automatic debug logs

xs5871:
Automatically enable debug output if a serial monitor is connected to the serial interface.

### DIFF
--- a/docs/en/debugging.md
+++ b/docs/en/debugging.md
@@ -1,11 +1,13 @@
 # Debugging
-Debug will output most of the useful state to the console. This can be enable in your firmware
-by setting this in your keymap. NOTE that it will be slower, so only enable this when you
-need debugging.
+KMK's debug output is written to CircuitPython's serial console -- the one that's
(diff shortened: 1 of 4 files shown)


regicidalplutophage (approved):
Cautiously approving this.

xs5871:
Boot configuration doesn't really affect this. If the serial console is disabled, then of course no debug messages are processed. If, as by default, the serial endpoint is connected: if and only if a serial monitor is actually connected (picocom, screen, etc.) debug messages are enabled. They are disabled again once the monitor is properly disconnected/closed.
There's one stumbling block where a serial monitor can be closed without resetting the connection (in picocom that's ctrl-q vs. ctrl-x) -- but that'll only happen if debugging was enabled intentionally in the first place. That's a reasonable trade-off for the convenience in my opinion and shouldn't affect any casual users.
Does this explanation address your concerns? I'll wait with the merge for your feedback.

regicidalplutophage:
U-huh, I misunderstood it then.
All good, no issues 👍

Issue #913, opened 2023-11-15
[BUG] Statusled brightness change crashes


minordemon:
When changing the brightness for the statusled the extension crashes with an unknown variable error (_brlghtness)

I will open a merge request to fix the wrong variable name myself.

PR #914, opened 2023-11-15
#913 fix statusled extension variable name typo

minordemon:
Fix for issue 913.

Renamed self._brightness to self.brightness


xs5871 (changes requested):
Github changed something in the UI and I can't review individual code segments and suggest changes anymore. Very cool.
Line 64: There's one missed `_brightness` that needs renaming.

xs5871 (approved):
Looks good.

minordemon:
Wow, I failed at doing a replace all.. <sigh>

Thanks for checking.

PR #915, opened 2023-12-04
[do not merge] Add nullbitsco tidbit board with pin mapping; doc updates; hex composer

patricksurry:
The hex_compose module is experimental keyboard-side composer for ascii/unicode characters which i'll use to drive a retro pc.   I can separate into a different PR if that's helpful

PR #916, opened 2023-12-04
Mapping for nullbits tidbit board with bit-c pro

patricksurry:
Split from https://github.com/KMKfw/kmk_firmware/pull/915 after review by board owner.  cc @jaygreco


jaygreco (approved):

xs5871:
GHA runs `make lint` for linting. The `fix-*` targets are for mostly automatic fixing of lint errors and are meant to be run manually before a commit. The `lint` and `fix` targets run different tools and have slightly different code coverage for some historical reason... There can sometimes be differences, this time it's something on your side though. Maybe your master branch or your tooling is out of date.

xs5871 (approved):

PR #918, opened 2023-12-04
[do not merge] Tidbit hexpad with keyboard-side hex character composition

patricksurry:
More complex tidbit hexpad build, with experimental keyboard-side ascii/unicode character composition module.
The idea is to eventually support a direct interface between keyboard and a retro-computer via SPI where it sends 8-bit ascii equivalents for keycodes.
 
Depends on https://github.com/KMKfw/kmk_firmware/pull/916


patricksurry:
I'll rebase and resubmit this once 916 is merged

PR #921, opened 2023-12-09
board hygiene for local mocking

patricksurry:
i'm working on an extension that imports a kmk board locally using mocks so I can reflect layout info and draw an SVG version with https://github.com/caksoylar/keymap-drawer.  I planned to test it on a bunch of KMK layouts, but found that many of them have errors which prevent importing them even in a mocked setup.  

So I added a unittest in `test_boards.py` that tries importing all 70+ boards/**/main.py in turn using `tests/mock_board.py`.  This turned up a bunch of simple issues that I fixed (capitalization etc), including some doc stuff.  The only one I don't know how to deal with is that several of the `dactyl` keyboards try to map an unknown AVR pin `B7`.   I ignore that for now but happy to hear if anyone knows the correct fix.

The test itself is a little slow (about 5s on my old macbook) so I guess could be optional.  Let me know what you think.


xs5871:
Sounds reasonable in general, but we should make the unit tests for contributions and the firmware core independent. Suggestion: two new make targets `unit-tests-boards` and `unit-tests-kmk` (I'm open to better names), make `unit-tests` pull in both as a catch-all, adjust github actions to only trigger if respective sources change, and reflect all that in the docs.

regicidalplutophage:

PR #924, opened 2023-12-25
Removed intify_coordinate from code, boards, docs

### LINKED ISSUE #922: [Cleanup] Remove `intify_coordinate` from code, boards, docs
Remove `intify_coordinate` from code, boards, docs:
It's a legacy mechanism that is obfuscating and potentially confusing, and for a long time was only kept for backwards compatibility.

NDalt:
Modified the board code that used intify_coordinates to use the same method as other boards to the best of my ability, removed the references to intify_coordinate in the "porting to kmk" documentation, and then removed the intify_coordinate method itself.
Resolves #922


xs5871 (approved):
That should do it. Thanks.

PR #926, opened 2023-12-29
Minor tweaks to the example in the Display extension doc

brianmarco:
A few tweaks to the example mentioned in the Display extension documentation.

- Moved the `device_address` argument from the `Display` instantiation to the `SSD1306` instantiation.  `Display` doesn't currently support this argument, but `SSD1306` does (https://github.com/KMKfw/kmk_firmware/blob/master/kmk/extensions/display.py#L137)
- Added missing comma to the arguments for the `Display` instantiation.

### DIFF
--- a/docs/en/Display.md
+++ b/docs/en/Display.md
@@ -197,10 +197,10 @@ layers = Layers()
 keyboard.modules.append(layers)
 
 i2c_bus = busio.I2C(board.GP21, board.GP20)
-display_driver = SSD1306(i2c=i2c_bus)
+display_driver = SSD1306(i2c=i2c_bus, device_address=0x3C)
 
 display = Display(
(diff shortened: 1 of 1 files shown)


xs5871 (commented):
Looks good. One thing: the `device_address` value is the same as the default. I think we should clarify that, i.e. comment it out as optional and add that `0x3C` is the default and only needs to be changed if necessary -- just to avoid confusion and duplicate code proliferation.

brianmarco:
Thanks for the review @xs5871 . I applied your feedback as suggested. I also applied the same feedback to the `width` argument for `Display`, and adjusted the `dim_target` argument to make it different from the default.

xs5871 (approved):
Cool, thanks.

PR #927, opened 2024-01-08
Fix digitalio matrix orientation

regicidalplutophage:
Fixing the bug I introduced earlier. Not especially pretty, but this scanner's code is very hard to follow in general.


xs5871 (approved):
Should be good.

edit: unit test is failing

PR #933, opened 2024-01-21
Remove non-English docs

claycooper:
The non-English documentation was not complete and undermaintained. The work put into initially creating the docs is appreciated however having inaccurate documentation is at the same level as having no documentation. With most browsers supporting built-in live translation there might no longer be a need for maintaining localized docs. If someone would like to provide and maintain translations then that work would be welcomed.

The English docs were left in the `en` directory to prevent breaking any links, at least for now.


regicidalplutophage (approved):
Yep...

klardotsh (approved):
Agreed all around. Thanks to all of the initial contributors of these translations, in many cases way back in the early days. They're appreciated!

PR #935, opened 2024-01-30
Add steno module

Cheetah26:
Adds a module for communicating with stenography software over serial.

### DIFF
--- a/docs/en/steno.md
+++ b/docs/en/steno.md
@@ -22,49 +22,49 @@ keyboard.modules.append(Steno())
 
 The following keys are created for use in your keymap:
 
-| Keycode   | Description   |
-|-----------|---------------|
-| STN_LS1   | S1-           |
-| STN_LS2   | S2-           |
-| STN_LT    | T-            |
-| STN_LK    | K-            |
-| STN_LP    | P-            |
-| STN_LW    | W-            |
-| STN_LH    | H-            |
-| STN_LR    | R-            |
-| STN_A     | A             |
-| STN_O     | O             |
-| STN_AS1   | * Top-left    |
-| STN_AS2   | * Lower-left  |
-| STN_AS3   | * Top-right   |
-| STN_AS4   | * Lower-right |
-| STN_E     | E             |
-| STN_U     | U             |
-| STN_RF    | -F            |
-| STN_RR    | -R            |
-| STN_RP    | -P            |
-| STN_RB    | -B            |
-| STN_RL    | -L            |
-| STN_RT    | -T            |
-| STN_RS    | -S            |
-| STN_RD    | -D            |
-| STN_RZ    | -Z            |
-| STN_N1    | Number bar 1  |
-| STN_N2    | Number bar 2  |
-| STN_N3    | Number bar 3  |
-| STN_N4    | Number bar 4  |
-| STN_N5    | Number bar 5  |
-| STN_N6    | Number bar 6  |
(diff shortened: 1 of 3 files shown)


kmk/modules/steno.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # flip a key's bit in the buffer
+    def _steno_press(self, key, *_):
+        idx = self._codes[key.code]
+        self._buffer[idx // 7] |= 0x80 >> (idx % 7 + 1)
+
+    # send all keys that were pressed, and reset the buffer
+    def _steno_release(self, *_):
+        usb_cdc.data.write(self._buffer)
+        self._initialize_buffer()
```
xs5871, line 77: There are two memory consumption optimization that are fairly obvious:
* reuse `self._buffer` instead of creating a new instance
* generate buffer index and code only once during key creation and encode them in `key.code`

That would look like this for example:
```python
class Steno(Module):
    def __init__(self):
        self._buffer = bytearray(6)
        self._initialize_buffer()

        for idx, key in enumerate(STENO_KEYS):
            make_key(
                code=((idx // 7) << 8) | (0x40 >> (idx % 7))
                names=(key,),
                on_press=self._steno_press,
                on_release=self._steno_release
            )

    def _initialize_buffer(self):
        self._buffer[:] = b'\x08\x00\x00\x00\x00\x00'

    # flip a key's bit in the buffer
    def _steno_press(self, key, *_):
        self._buffer[key.code>>8] = key.code & 0xff

    # send all keys that were pressed, and reset the buffer
    def _steno_release(self, *_):
        usb_cdc.data.write(self._buffer)
        self._initialize_buffer()
```

It also looks like you're sending the buffer for every key release, but only the first release sends any meaningful data. The other releases send what I assume is interpreted as empty. This could be further optimized by checking if the buffer contents are non-trivial (or using a flag) to cut down on superflous transmissions. I did not look into how the protocol works.

docs/en/steno.md
```diff
@@ -0,0 +1,71 @@
...
+keyboard.modules.append(Steno())
+```
+
+## Keys
+
+The following keys are created for use in your keymap:
+
+| Keycode   | Description   |
+|-----------|---------------|
+| STN_LS1   | S1-           |
```
xs5871, line 27: Keycode table formatting:
```suggestion
| `KC.STN_LS1`   | S1-           |
```

xs5871 (changes requested):
This looks interesting.

kmk/modules/steno.py
Cheetah26: Good points!

I've included your suggestions, and tested that everything is still working as expected.

docs/en/steno.md
```diff
@@ -0,0 +1,71 @@
...
+keyboard.modules.append(Steno())
+```
+
+## Keys
+
+The following keys are created for use in your keymap:
+
+| Keycode    | Description   |
+|------------|---------------|
+| `STN_LS1`  | S1-           |
```
xs5871, line 27:
```suggestion
| `KC.STN_LS1`  | S1-           |
```

You have to prefix with `KC.`.

xs5871 (changes requested):

docs/en/steno.md
Cheetah26: Didn't notice that, my bad

xs5871 (approved):

PR #936, opened 2024-01-30
Add data serial to `bootcfg` and rename console serial

xs5871:
Add data serial to bootcfg, rename console serial to distinguish between them.

Draft, because I have currently no clue why the spellchecker started complaining about a single code segment...


xs5871:
Ok, got it. The failed check was further down, but it also tagged every previous occurance.

regicidalplutophage (approved):

PR #941, opened 2024-02-23
Do not delete the keys pressed in a sequence

juru1234:
Remember the keys that were pressed before entering the sequence. E.g. when a sequence creates the German Umlaut "ä", holding shift should lead to "Ä". This is also the behavior in QMK.


juru1234:
Sorry, the patch was wrong. Only the line `keyboard.keys_pressed = set()` should be deleted.

PR #942, opened 2024-02-26
Add a key for replacing the top layer

xs5871:
Reference: [chat discussion](https://kmkfw.zulipchat.com/#narrow/stream/384078-KMK-development/topic/adding.20an.20extra.20way.20of.20switching.20layers)

### DIFF
--- a/kmk/modules/layers.py
+++ b/kmk/modules/layers.py
@@ -52,6 +52,9 @@ class Layers(HoldTap):
             on_release=self._mo_released,
         )
         make_argumented_key(
+            validator=layer_key_validator, names=('FD',), on_press=self._fd_pressed
+        )
+        make_argumented_key(
             validator=layer_key_validator, names=('DF',), on_press=self._df_pressed
         )
         make_argumented_key(
@@ -79,11 +82,17 @@ class Layers(HoldTap):
             on_release=self.ht_released,
         )
 
+    def _fd_pressed(self, key, keyboard, *args, **kwargs):
+        '''
+        Switches the top layer
+        '''
+        self.activate_layer(keyboard, key.meta.layer, idx=0)
+
     def _df_pressed(self, key, keyboard, *args, **kwargs):
         '''
         Switches the default layer
         '''
-        self.activate_layer(keyboard, key.meta.layer, as_default=True)
+        self.activate_layer(keyboard, key.meta.layer, idx=-1)
 
     def _mo_pressed(self, key, keyboard, *args, **kwargs):
         '''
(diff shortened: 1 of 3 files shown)


kmk/modules/layers.py
```diff
@@ -79,11 +82,17 @@ def __init__(
...
+        Switches the top layer
+        '''
+        self.activate_layer(keyboard, key.meta.layer, idx=0)
+
     def _df_pressed(self, key, keyboard, *args, **kwargs):
         '''
         Switches the default layer
         '''
-        self.activate_layer(keyboard, key.meta.layer, as_default=True)
+        self.activate_layer(keyboard, key.meta.layer, idx=-1)
```
regicidalplutophage, line 95: I just really want to say I really like this!

regicidalplutophage (approved):
:+1:

ronsmits:
so Using this change I expected that the base layer and the current layer are active, thet are not. It changes the default layer and it seems other layers are not active. Running it from thonny shows that indeed all other layers are deactivated:
```
200175719 kmk.keyboard: coordkeys_pressed={0: Key(code=1010, has_modifiers=None)}
200175725 kmk.modules.layers: active_layers=[1]
200175849 kmk.keyboard: <Event: key_number 0 released>: Key(code=1010, has_modifiers=None)
200175855 kmk.keyboard: coordkeys_pressed={}
200178848 kmk.keyboard: <Event: key_number 0 pressed>: Key(code=1011, has_modifiers=None)
200178854 kmk.keyboard: coordkeys_pressed={0: Key(code=1011, has_modifiers=None)}
200178860 kmk.modules.layers: active_layers=[2]
200178980 kmk.keyboard: <Event: key_number 0 released>: Key(code=1011, has_modifiers=None)
200178986 kmk.keyboard: coordkeys_pressed={}
200180793 kmk.keyboard: <Event: key_number 0 pressed>: Key(code=1012, has_modifiers=None)
200180797 kmk.keyboard: coordkeys_pressed={0: Key(code=1012, has_modifiers=None)}
200180803 kmk.modules.layers: active_layers=[0]
200180924 kmk.keyboard: <Event: key_number 0 released>: Key(code=1012, has_modifiers=None)
200180931 kmk.keyboard: coordkeys_pressed={}
```

xs5871:
`KC.FD` replaces the top most layer. If only the "base" layer (i.e. lowest layer) is active, then that layer is also the top layer -- in that case `KC.DF(...)` and `KC.FD(...)` have identical behavior. Btw just to clarify, there's no such thing as a default or current layer, those are colloquialisms; there's a list of active layers with key positions being resolved by layer order and key transparency.

I'm guessing you probably want to keep layer 0 as the lowest layer. If so, you want to add another layer on top first, and then switch out that top layer, then activating only layer 0 to complete the cycle. Example snippet for 1 base and 3 overlay layers:
```python
keymap = [
[KC.TG(1)],
[KC.FD(2)],
[KC.FD(3)],
[KC.TO(0)],
]
```

ronsmits:
Thank you very much for explaining this, it works now :)

PR #950, opened 2024-03-26
Keymap for Budgy Keyboard

JakobEdvardsson:
This config has both US-qwerty and Swedish-Colemak-DH in mind.  
The default keymap is US-qwerty, in order to switch to the Swedish-Colemak-DH keymap,
comment out the US-qwerty keymap and uncomment the Swedish-Colemak-DH keymap in the main.py file.


boards/budgy/qwerty/kb.py
xs5871, line 1: Why does qwerty need a seperate `kb.py`?

JakobEdvardsson: Will fix, did not mean to do that

xs5871 (commented):
If you're providing two keymaps, wouldn't it make more sense to have common `kb.py` and `main.py`, and split off only the keymaps as `keymap_sw.py` and `keymap_us.py` for example?

JakobEdvardsson:
> If you're providing two keymaps, wouldn't it make more sense to have common `kb.py` and `main.py`, and split off only the keymaps as `keymap_sw.py` and `keymap_us.py` for example?

You are right! Will make sure to do that instead!

JakobEdvardsson:
@xs5871 made the keymap more modular, all you have to do is to uncomment/comment out an import to change layout.

xs5871:
There are linter errors that need fixing

xs5871 (approved):

PR #957, opened 2024-04-05
Budgy, More advanced SWE layout and fixed US Qwerty

JakobEdvardsson:
Hi, I have made quite a few changes to the Swedish version of this layout.

I have also fixed the US Qwerty variant since I discovered an error in the keymap.


boards/budgy/main.py
```diff
@@ -26,22 +29,43 @@
...
+class Layers(_Layers):
+    last_top_layer = 0
 
+    def after_hid_send(self, keyboard):
+        if keyboard.active_layers[0] != self.last_top_layer:
+            self.last_top_layer = keyboard.active_layers[0]
+            if self.last_top_layer == 6:  # mouse layer
+                led.value = True
+            else:
+                led.value = False
```
xs5871, line 56: I'd recommend wrapping `activate_layer` and `deactivate_layer` instead of constantly polling with `after_hid_send`.
example:
```python
class Layers(_Layers):
    def activate_layer(self, keyboard, layer, idx=None)
        super().activate_layer(keyboard, layer, idx)
        if keyboard.active_layers[0] == 6:
            led.value = True
```

JakobEdvardsson: Thank you that makes a lot more sense! Will fix!

xs5871 (approved):

PR #960, opened 2024-04-09
Cleanup old timeout code

xs5871:
bit of code cleanup


claycooper (approved):
Looks good. I can't find anything in the history about when/why this was even used.

xs5871:
It's from before we had the scheduler.

PR #965, opened 2024-04-25
[peg_oled_display] Update outdated `show()` calls

after-ephemera:
These were [deprecated in circuitpython 9](https://github.com/adafruit/circuitpython/pull/8456) and need to be updated to use the `root_group` API to support latest versions.

`root_group` has been around for some time so although this is a breaking change for almost anyone using Peg + Blok. 

Anecdotally, my Lily58 was an early Blok build and the microcontroller I got from BoardSource had circuitpython 8.x.

### DIFF
--- a/kmk/extensions/peg_oled_display.py
+++ b/kmk/extensions/peg_oled_display.py
@@ -97,7 +97,7 @@ class Oled(Extension):
                 y=25,
             )
         )
-        self._display.show(splash)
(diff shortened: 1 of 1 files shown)


lucacri (approved):
I tried it and worked!

xs5871:
I'm closing the PR because this extension is being removed with #977. Please use the `display` extension instead.

PR #967, opened 2024-05-12
Re-implement sequences as macro in a module

xs5871:


xs5871:
Better macros. It's not *super* pretty, but it's good enough to be an improvement:
* It's a module. No unecessary code in core.
* Asynchronous. Doesn't lock the entire firmware, but it will buffer key events when a macro is running and replay them afterwards.
* Clean Interface. No more awkward double negative `no_press`/`no_release`.
* Unittestable.
* Unintentionally flexible. Effectively any code can be put in a macro now.
* Spellchecked documentation on the third try.

I'll purge the abomination that's the `sequences` code soonish. A couple of board configurations need refactoring, which I'm not looking forward to.

xs5871:
Sometimes clicking the right button is hard.

regicidalplutophage (approved):
I like this a lot!

anventia:
It seems like the docs have not been updated, and still shows the old Sequences, can this please be updated?

claycooper:
New docs have been pushed to the website.

PR #969, opened 2024-05-19
Implement automatic firmware reload and show error stacktrace

xs5871:
Trigger a supervisor reload on any exception, restarting the firmware; unless it's a keyboard interrupt in which case we explicitly want to drop into the REPL.
Replace the sparse debug error message with a stacktrace.

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -461,17 +461,28 @@ class KMKKeyboard:
                 debug_error(ext, 'deinit', err)
 
     def go(self, hid_type=HIDModes.USB, secondary_hid_type=None, **kwargs) -> None:
-        self._init(hid_type=hid_type, secondary_hid_type=secondary_hid_type, **kwargs)
         try:
+            self._init(
+                hid_type=hid_type,
+                secondary_hid_type=secondary_hid_type,
+                **kwargs,
+            )
             while True:
                 self._main_loop()
         except Exception as err:
-            debug_error(self, 'Unexpected error', err)
+            import traceback
+
+            traceback.print_exception(err)
         finally:
             debug('cleaning up...')
             self._deinit_hid()
             self.deinit()
             debug('...done')
+
+            if not debug.enabled:
+                import supervisor
+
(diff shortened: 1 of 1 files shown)


regicidalplutophage (commented):
Maybe `reload = True` should be an object in KMKKeyboard so that we can toggle this behavior in userspace?

xs5871:
It started out that way and then removed it. Couldn't think of a reason to ever disable this in "normal" use; I always rather have the keyboard reboot than manually unplug-replug it.
Only when debugging crashes would I maybe want to stop the autoreload occasionally, and then I have the serial console open anyways and can just look at the terminal history or hit ctrl-c (or disable reloading in code).

regicidalplutophage:
Well, ideally errors shouldn't happen, I'd argue if they are that's barely a normal use

xs5871:
Any sufficiently complex software has bugs, encountering them is part of the UX, as undesireable as that may be. That is beside the point though.

Can you give a realistic example where turning of the reload is an improvement to the user experience, discounting scenarios that require the serial console?

claycooper:
The only issue I can think of with auto-restarting is that a crash caused by an obscure, rare bug would get masked and attributed to an unrelated symptom however that's more of an item to note when troubleshooting or assisting someone and not a blocker to this PR.

I'll approve this PR and leave it up to @xs5871 to merge depending on the conversation's outcome with @regicidalplutophage.

claycooper (approved):

xs5871:
I wouldn't say it "masks" issues, it automates the recovery. Crashes should still be very noticable.
Our recommendation for figuring out any kind of unexpected behavior already is: check debug output. Doesn't really impact diagnostic procedure at all, in my opinion. Even with the current crash and drop to REPL you'd only know something died by checking the logs.

xs5871:
Here's an idea: we could couple the reload to the debug flag, which is already automagically enabled and disabled depending on an active connection to the serial console.

regicidalplutophage:
Yeah, that would be perfect

regicidalplutophage (approved):

PR #972, opened 2024-05-25
Update linting

xs5871:
After updating to python 3.12 my venv broke, there were dependency issues, updated linter rules failed, etc...
One thing lead to another, now we have proper linting for everything under `boards/*`, instead of blanket-ignoring random rules for reasons no one can remember.


claycooper (approved):
I can only imagine how determined you must have been to make this many changes.

xs5871:
Wasn't too bad, to be honest. Had an unexpectedly helpfull bout of editing zen and hyperfixation.

PR #977, opened 2024-06-06
Remove the `peg_oled_display` extension

xs5871:
Reasons for the removal:
* terrible performance
* poor code quality
* unmaintained / unsupported

Where applicable: replace with the officially supported `display` extension.


claycooper (approved):

claycooper:
Thank you.

regicidalplutophage:
I'm not against removing it, but maybe we should come up with an example that's basically the re-implementation of PEG OLED to use in affected boards.

xs5871:
All boards in our repository that used it are "ported" as part of the commit. To the best of my ability at least; they only displayed the top most layer anyway.

regicidalplutophage:
> they only displayed the top most layer anyway

I missed that detail about them.

LGTM!

regicidalplutophage (approved):

PR #978, opened 2024-06-07
Refactor sticky/oneshot keys

xs5871:
There was a thing that didn't work for me, and once again... as I came to realize, the current implementation based on modified hold-tap mechanism wasn't the best choice in the first place and increasingly difficult to work around.

* complete rewrite, basically from scratch
* rebranded to "sticky keys", because that is unarguably the more sensible name
* no more dependency on hold-tap
* actually helpful comments (I hope)
* works from tap dance, can even stack multiple sticky keys from a single tap dance
* even more tests
* new feature to stack with any other keys until all keys are released

The old `oneshot` module is still in place, can be cleaned up eventually.

### DIFF
--- a/docs/en/README.md
+++ b/docs/en/README.md
@@ -28,7 +28,7 @@ Before you look further, you probably want to start with our [getting started gu
 - [HoldTap](holdtap.md): Adds support for augmented modifier keys to act as one key when tapped, and modifier when held.
 - [Macros](macros.md): Adds macros.
 - [Mouse keys](mouse_keys.md): Adds mouse keycodes
-- [OneShot](oneshot.md): Adds support for oneshot/sticky keys.
+- [Sticky keys](sticky_keys.md): Adds support for sticky keys.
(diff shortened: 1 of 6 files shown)


regicidalplutophage:
Oneshot name is a [staple](https://github.com/qmk/qmk_firmware/blob/user-keymaps-still-present/users/callum/readme.md) when discussing small keymaps. I don't mind the module being called sticky keys, but can you make it clear the sticky keys are the same as oneshot?

I'm a heavy oneshot user with the implementation almost the same as Callum, so I'm going to test it before approving.
One feature I'd like is if a sticky keycode is being held for some time, it starts being treated as a normal keycode that's being held. Current implementation didn't mix well with holdtap, so going to test that too. The reason is modifier+mouse inputs are common in CAD packages and mouse being a different device has no effect on sticky/oneshot keys.

regicidalplutophage (approved):
I've been using this for a while and it's great!

PR #980, opened 2024-06-08
Remove key pre/post handlers in favor of macros

xs5871:
I have only once seen a legitimate use case for these handlers -- one where custom handlers would've been a bit more complicated, and that use case was so niche that I can't remember what it was.
I don't think this functionality and the associated overhead should be the default for *every single key*.
Also: macros can replace this functionality without requiring users to learn how to make custom keys from scratch.

### DIFF
--- a/docs/en/keys.md
+++ b/docs/en/keys.md
@@ -6,6 +6,8 @@ you're stumped: [`kmk/keys.py`](/kmk/keys.py).
 
 ---
 
+## Key Objects
+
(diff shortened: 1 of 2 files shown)


regicidalplutophage (approved):
Sad to see that chunk of text go, but it's for the better

xs5871:
At least the always topical shrek will be preserved for posterity in `user_keymaps/`.

PR #985, opened 2024-06-12
rgb_2d: 2D RGB effects extension.

Gadgetoid:
Add a 2D RGB effects extensions which allows for effects like Larson scanning or rainbow cycling to span the whole width/height of a board.

After making RGB fast on the Keybow 2040 I wanted some nice effects to run on it.

This is a very early prototype of a 2D RGB effects extension.

Features:

* Effects are pluggable by design
* Animations are wall-timed - `time.monotonic()` isn't great for this afaik but it works - to be framerate independent
* Effects are constant time, you can know how quick a colour cycle will finish, etc
* Reports framerate (potential, not actual) over debug

Limitations:

1. Supports only *one* PixelBuf or compatible, use an extension instance for each PixelBuf
2. All key binding and customisation functionality has been removed until the core is finished
3. Effects are functions which return an `animate` closure, they can set up some basic variables to store state, or accept config arguments but mostly only take a "t" variable from which RGB effects are algorithmically derived
4. Only tested on a 4x4 matrix powered by an RP2040

TODO:

* [ ] Add support for passing config args to effects, maybe via an effects stack which would allow a user to customise each effects finer points to their board and then just control basic stuff (brightness, hue, selected effect) via key bindings for the whole stack
* [ ] Add back all the key bindings
* [ ] Documentation
* [ ] Testing (anyone got a nice big RGB backlit board to run this on?)

### DIFF
--- a/kmk/extensions/rgb_2d.py
+++ b/kmk/extensions/rgb_2d.py
@@ -1,0 +1,345 @@
+from math import e, exp, pi, sin, floor, atan2
+from random import randint
+
+from kmk.extensions import Extension
+from kmk.handlers.stock import passthrough as handler_passthrough
+from kmk.keys import make_key
+from kmk.scheduler import create_task
+from kmk.utils import Debug, clamp
+import time
+
+debug = Debug(__name__)
+
+
+class Timing:
+    WALL = 0
+    FRAME = 1
+
+
+def hsv_to_rgb(hue, sat, val):
+    '''
+    Converts HSV values, and returns a tuple of RGB values
+    :param hue:
+    :param sat:
+    :param val:
+    :return: (r, g, b)
+    '''
+    if sat == 0:
+        return (val, val, val)
+
+    hue = 6 * (hue & 0xFF)
+    frac = hue & 0xFF
+    sxt = hue >> 8
+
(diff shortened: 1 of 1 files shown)


Gadgetoid:
I should probably also pay attention to the mentioned issues here: https://kmkfw.zulipchat.com/#narrow/stream/384078-KMK-development/topic/RGB.20Feature.3F

I don't believe animations really *need* start and stop methods, they should be as close to stateless - think pixel shader - as possible. They serve only to answer the question: "what colour should this LED be at this time."

rgb_2d is not trying to refactor rgb but there are probably few reasons why it couldn't *replace* it.

Key awareness is something I looked into but I don't know enough about the codebase to tackle it yet. I wanted animation effects that would respond to a keypress but without an LED->Key map and a position aware keypress handler that's a non starter.

xs5871:
Matrix animations, or even some kind of positional awareness of addressable LEDs would be a sensible addition. What I don't want to happen is yet another LED extension that shares 80% of the funcionality with the other 4.

xs5871:
rgb, peg_rgb_matrix, led, statusled all share at least *some* kind of functionality. The "other 4" was more of a "O(4)" estimate.

PR #987, opened 2024-06-12
Let copy-compiled remove old mpy files from device

xs5871:


claycooper:
Are people going to want to selectively copy compiled files? Or does our stance of not mixing and matching modules preclude that?

xs5871:
At least boards based on the nice!nano want to strip out unused modules and that is what we have been recommending in the past to get around limited flash. But those wouldn't use the bulk copy with `copy-compiled` target anyway.
Occasionally we remove or rename files. I would consider not cleaning up old "binaries" during an update process a bug, that's the rational.

claycooper (approved):

claycooper:
I agree with that stance.

PR #991, opened 2024-06-18
Adds nullbitsco/nibble mapping with multiplex keymatrix support

patricksurry:
This extends digitalio.MatrixScanner to support multiplexed output scanning which is used by nibble.

### DIFF
--- a/boards/nullbitsco/nibble/README.md
+++ b/boards/nullbitsco/nibble/README.md
@@ -1,14 +1,11 @@
-Keyboard mapping for the [nullbits tidbit](https://nullbits.co/tidbit/).
+Keyboard mapping for the [nullbits nibble](https://nullbits.co/nibble/).
 
 Copy `kb.py` and `main.py` to your top level CircuitPython folder beside the kmk folder.
-Edit the key mapping in `main.py` to match your build,
-for example the number and location of encoders and double-size keys.
+Edit the key mapping in `main.py` to match your keyboard layout.
 
 The Keyboard constructor supports a couple of optional arguments (see `kb.py`).
 
-If you're setting your tidbit up in landscape mode,
-with the USB connector at top right instead of top left, pass
-`landscape_layout=True`.
+TODO
 
 You can specify the active encoder positions by passing a list like
 `active_encoders=[0, 2]` which corresponds to the 1st and 3rd positions shown
@@ -24,11 +21,11 @@ Here's an example:
 from kb import KMKKeyboard
 from kmk.extensions.rgb import RGB, AnimationModes
 
-keyboard = KMKKeyboard(active_encoders=[0], landscape_layout=True)
+keyboard = KMKKeyboard(active_encoders=[0])
 
 rgb = RGB(
     pixel_pin=keyboard.pixel_pin,
-    num_pixels=8,
+    num_pixels=10,
     animation_mode=AnimationModes.BREATHING,
     animation_speed=3,
     breathe_center=2,
--- a/boards/nullbitsco/nibble/README.md
+++ b/boards/nullbitsco/nibble/README.md
@@ -3,33 +3,8 @@ Keyboard mapping for the [nullbits nibble](https://nullbits.co/nibble/).
 Copy `kb.py` and `main.py` to your top level CircuitPython folder beside the kmk folder.
 Edit the key mapping in `main.py` to match your keyboard layout.
 
-The Keyboard constructor supports a couple of optional arguments (see `kb.py`).
+The Keyboard constructor supports an optional `encoder` argument (see `kb.py`).
+See the sample `main.py` for an example of how to configure encoder
 
-TODO
-
-You can specify the active encoder positions by passing a list like
-`active_encoders=[0, 2]` which corresponds to the 1st and 3rd positions shown
-in [step 6](https://github.com/nullbitsco/docs/blob/main/tidbit/build_guide_en.md#6-optional-solder-rotary-encoder-led-matrix-andor-oled-display) of the build guide.
-The default is for a single encoder in either of the top two locations labeled 1
-in the build diagram, i.e. `active_encoders=[0]`.  Pass an empty list if you skipped
-adding any encoders.
-
-You can control the RGB backlights with the [RGB extension](http://kmkfw.io/docs/rgb).
-Here's an example:
-
-```python
-from kb import KMKKeyboard
-from kmk.extensions.rgb import RGB, AnimationModes
-
-keyboard = KMKKeyboard(active_encoders=[0])
-
-rgb = RGB(
-    pixel_pin=keyboard.pixel_pin,
-    num_pixels=10,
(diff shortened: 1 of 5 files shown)


boards/nullbitsco/nibble/README.md
```diff
@@ -0,0 +1,10 @@
+Keyboard mapping for the [nullbits nibble](https://nullbits.co/nibble/).
+
+Copy `kb.py` and `main.py` to your top level CircuitPython folder beside the kmk folder.
+Edit the key mapping in `main.py` to match your keyboard layout.
+
+The Keyboard constructor supports an optional `encoder` argument (see `kb.py`).
+See the sample `main.py` for an example of how to configure encoder
```
xs5871, line 7:
```suggestion
See the sample `main.py` for an example of how to configure encoder.
```

boards/nullbitsco/nibble/kb.py
```diff
@@ -0,0 +1,50 @@
...
+
+        self.matrix = MatrixScanner(
+            col_mux_pins,
+            row_pins,
+            diode_orientation=DiodeOrientation.ROW2COL,  # row is anode, col is cathode
+            pull=digitalio.Pull.UP,
+            multiplexed=True,
+        )
+
+        if encoder:
```
xs5871, line 47: The encoder import belongs here, otherwise the encoder code is always loaded and wastes memory. Not an *issue* for this board, but a bad example to set.
```suggestion
        if encoder:
            from kmk.modules.encoder import EncoderHandler
```

boards/nullbitsco/nibble/main.py
```diff
@@ -0,0 +1,120 @@
...
+from kmk.extensions.media_keys import MediaKeys
+from kmk.extensions.rgb import RGB, AnimationModes
+from kmk.keys import KC
+
+keyboard = KMKKeyboard(encoder=True)  # assume encoder installed
+keyboard.extensions.append(MediaKeys())
+
+XXXXX = KC.NO
+
+keyboard.keymap = [
```
xs5871, line 12: Would you mind formatting this in the usual "looks like keymatrix" way? Prefix the keymap definition with `# fmt: off` and postfix with `# fmt:on` to selectively disable the linter.

boards/nullbitsco/nibble/kb.py
xs5871: Actually, this entire code block should go in `main.py`, including the conditional import.

boards/nullbitsco/nibble/main.py
```diff
@@ -0,0 +1,120 @@
...
+)
+
+keyboard.extensions.append(rgb)
+
+
+def set_backlight(hsv):
+    rgb.hue, rgb.sat, rgb.val = hsv
+
+
+set_backlight((180, 255, 50))
```
xs5871, line 117: Why is this here? You can set the default hsv in the `RGB` constructor.

kmk/scanners/digitalio.py
```diff
@@ -35,21 +36,23 @@ def __init__(
         #
         # repr() hackery is because CircuitPython Pin objects are not hashable
         unique_pins = {repr(c) for c in cols} | {repr(r) for r in rows}
+        print(unique_pins)
```
xs5871, line 39:
```suggestion
```

boards/nullbitsco/tidbit/README.md
```diff
@@ -27,7 +27,7 @@ from kmk.extensions.rgb import RGB, AnimationModes
 keyboard = KMKKeyboard(active_encoders=[0], landscape_layout=True)
```
xs5871, line 27: This is not correct. There is no `landscape_layout` parameter.

kmk/scanners/digitalio.py
```diff
@@ -22,8 +22,9 @@ def __init__(
         rows,
         diode_orientation=DiodeOrientation.COL2ROW,
         pull=digitalio.Pull.UP,
-        rollover_cols_every_rows=None,
+        rollover_cols_every_rows=None,  # with value k, treat k*r x c matrix as r x c*k
         offset=0,
+        multiplexed=False,  # 2^k outputs are multiplexed on k output pins
```
xs5871, line 27: Not a big fan of packing even more conditional functionality into the already slow digitalio scanner.
I also don't quite understand what's happening here. Is that something that could be replaced by 
CircuitPythons native `keypad_demux.DemuxKeyMatrix` (currently beta)?

xs5871 (changes requested):

boards/nullbitsco/nibble/main.py
patricksurry: good point, hangover from some previous code

boards/nullbitsco/tidbit/README.md
patricksurry: There is for the tidbit, but not for the nibble.  See here: https://github.com/KMKfw/kmk_firmware/blob/main/boards/nullbitsco/tidbit/kb.py#L48

kmk/scanners/digitalio.py
patricksurry: I looked at `keypad_demux` docs, but it didn't seem to be available on my board yet.  It also didn't seem quite as flexible in terms of configuring which direction is multiplexed, and pullup v pulldown etc.

Honestly I was a bit confused why `digitialio.MatrixScanner` exists at all when we have `keypad.KeyMatrix` natively but I am pretty new to circuitpython/kmk.  I thought the modification was relatively small: in the regular case we scan column _n_ by activating the _n_-th column pin, but in the mutiplexed case we scan column _n_ by writing _n_ as a bit pattern _b0 b1 .. bk_ and set each multiplexed column bit.

If you prefer I could clone or subclass MatrixScanner to DemuxMatrixScanner instead?

patricksurry:
Thanks for the thorough review.  Let me know what you think about digitalio.MatrixScanner and I can refactor as needed.

kmk/scanners/digitalio.py
xs5871: Getting `keypad_demux` build for your board shouldn't be too difficult. It is still in beta in any case.

The `digitalio` scanner predates the CP native matrix scanner, and there're hardware implementations that don't work with `KeyMatrix` -- at least for now -- that's the reason it's still around. If in doubt: always use the native scanner.
Thank you for the explanation, I understand what kind of mux we're talking about now.

Can you subclass without duplicating all of the code? I'm a bit short on time atm. and can't estimate that right now.

earvingad:
> Thanks for the thorough review. Let me know what you think about digitalio.MatrixScanner and I can refactor as needed.

Hi, does this implementation on the `digitalio.py` adds the support for duplex matrix?

patricksurry:
wdym by duplex matrix?  

btw i've since extended the DemuxKeyMatrix object in circuitpython (see https://github.com/adafruit/circuitpython/pull/9912) which simplifies this PR.  I will update it

PR #993, opened 2024-06-25
Sticky keys: retap_cancel

regicidalplutophage:
An enhancement that makes it possible to release a sticky key by tapping it again -- useful when user only wants to tap a key(`KC.LGUI`) or simply changes their mind.
This behavior is enabled by default.

### DIFF
--- a/docs/en/sticky_keys.md
+++ b/docs/en/sticky_keys.md
@@ -31,9 +31,9 @@ The full sticky key signature is as follows:
 
 ```python
 KC.SK(
-    KC.ANY,                     # the key to made sticky
-    defer_release=False         # when to release the key
-    retap_cancel=True   # repeated tap releases the key
+    KC.ANY,              # the key to made sticky
+    defer_release=False  # when to release the key
+    retap_cancel=True    # repeated tap releases the key
 )
 ```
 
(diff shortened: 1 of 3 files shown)


xs5871:
I was actually thinking of implementing that functionality myself, so thumbs up from me.
Could we brainstorm the name a bit? How about something like `tap-cancel` or `retap-cancel` or `double-tap-cancel`? I'd prefer "cancel" because it cancels an action/state, a "release" is something the user does, usually.
What is your opinion on canceling all stacked sticky keys instead of just the one that is tapped?
Also: unit tests. I don't want to merge unless we have those, because otherwise we'll forget. You don't have to do them yourself, I can contribute those.

regicidalplutophage:
I like `retap-cancel`. Cancelling all sticky keys via a sticy key just doesn't feel right. Maybe a special key?

xs5871 (approved):
* fixed formatting, a typo, and missing hint about timeout refresh if retap enabled in the docs
* added a unit test

I can't request a formal approval from you to cross-check my contribution. If there's nothing off, you can squash and merge.

regicidalplutophage:
Oh, that reminds me: I haven't tested retap cancel with with defer release(I just did now, it works completely as expected)
Do you want to add test case for that as well?

xs5871:
Since the tests succeeded both with and without `retap_cancel=True` as default I didn't think of that. Doesn't hurt to check though.

regicidalplutophage:
Thanks!

PR #995, opened 2024-06-30
Refactor key meta

xs5871:
As promised in #988:
> Coming soon: getting rid of key.meta, aka: "we already use classes for key types, but only sometimes, and we hide it behind a layer of indirection".

All `key.meta` classes and attributes have been refactored into flat `Key` classes, plus some cleanup and code simplification.
Anecdotal performance improvement: my personal config uses ~2kB less RAM with these changes.


regicidalplutophage (approved):
It even takes up less memory, nice!

PR #996, opened 2024-07-01
Fix _rgb_tog

calico-cat-3333:
After upgrading to the latest version of KMK, I encountered this error: 
```
AttributeError: 'RGB' object has no attribute '_rgb_to'
```
It looks like `self._rgb_tog` was misspelled as `self._rgb_to`.


regicidalplutophage (approved):
Thanks!

PR #998, opened 2024-07-02
Update Getting_Started.md

regicidalplutophage:
Default branch has been changed a while ago, changing documentation to reflect that.

### DIFF
--- a/docs/en/Getting_Started.md
+++ b/docs/en/Getting_Started.md
@@ -8,7 +8,7 @@ Known working and recommended devices can be found in the [list of officially su
 ## TL;DR Quick start guide
 > To infinity and beyond!
 1. [Install CircuitPython version 7.3 or higher on your board](https://learn.adafruit.com/welcome-to-circuitpython/installing-circuitpython). With most boards, it should be as easy as drag and dropping the firmware on the drive
-2. Get a [copy of KMK](https://github.com/KMKfw/kmk_firmware/archive/refs/heads/main.zip) from the main branch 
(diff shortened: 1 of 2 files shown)


docs/en/Getting_Started.md
```diff
@@ -8,7 +8,7 @@ Known working and recommended devices can be found in the [list of officially su
 ## TL;DR Quick start guide
 > To infinity and beyond!
 1. [Install CircuitPython version 7.3 or higher on your board](https://learn.adafruit.com/welcome-to-circuitpython/installing-circuitpython). With most boards, it should be as easy as drag and dropping the firmware on the drive
-2. Get a [copy of KMK](https://github.com/KMKfw/kmk_firmware/archive/refs/heads/master.zip) from the master branch 
+2. Get a [copy of KMK](https://github.com/KMKfw/kmk_firmware/archive/refs/heads/master.zip) from the main branch
```
xs5871, line 11:
```suggestion
2. Get a [copy of KMK](https://github.com/KMKfw/kmk_firmware/archive/refs/heads/main.zip) from the main branch 
```

xs5871 (commented):
This isn't the only place we still refer to `master`, according to a quick grep over docs.

Since we're at it: I don't think we should recommend the zip as the primary source for KMK. It makes it look like that is stable and doesn't change all to often to a novice, partly because it doesn't have a version number.
There should at least be a hint somewhere very early in the documentation (probaly README.md) that KMK changes often and sometimes a lot, and a fork is the easiest way to keep up, but if you don't want that, get the zipped snapshot with the associated downsides.

xs5871 (approved):
Excellent.

PR #1002, opened 2024-07-05
Fix inconsistent macro state when double tapping

xs5871:
See [chat](https://kmkfw.zulipchat.com/#narrow/stream/358347-general/topic/Creating.20a.20Macro.20-.20Alt.20.2B.20Tab.20.2F.20Application.20Switcher) for reference.

### DIFF
--- a/kmk/modules/macros.py
+++ b/kmk/modules/macros.py
@@ -200,13 +200,9 @@ class Macros(Module):
         return
 
     def process_key(self, keyboard, key, is_pressed, int_coord):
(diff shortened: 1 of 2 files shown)


xs5871:
How about now? There's a bunch more tests, including activating multiple macros at the same time, blocking and non-blocking. I also think this is more readable.

regicidalplutophage (approved):
I wasn't able to break it :+1:

xs5871:
Someone else will, I'm sure.

PR #1012, opened 2024-07-26
bugfix: `KMKKeyboard.during_bootup` method

### LINKED ISSUE #1011: [BUG] uninitialized extensions not removed
**Describe the bug**
in [kmk/kmk_keyboard.py#L343C11-L350C57](https://github.com/KMKfw/kmk_firmware/blob/d6cbb8c60654dba24a9d3c967cc5946a19fb5c7a/kmk/kmk_keyboard.py#L343C11-L350C57), there's logic to remove failing modules from the Keyboard, and it seems like there was an attempt to do the same with extensions, 
but due to what i assume is a typo, the code just re-removes failed modules and doesn't remve failed extensions

**To Reproduce**
N/A

**Expected behavior**
I would expect [kmk/kmk_keyboard.py#L350](https://github.com/KMKfw/kmk_firmware/blob/d6cbb8c60654dba24a9d3c967cc5946a19fb5c7a/kmk/kmk_keyboard.py#L350) to be `self.extensions[:] = [_ for _ in self.extensions if _]`

**Debug output**
N/A

**Additional context**
N/A

alextremblay:
Fix #1011

### DIFF
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -347,7 +347,7 @@ class KMKKeyboard:
                 debug_error(ext, 'during_bootup', err)
                 self.extensions[idx] = None
 
-        self.modules[:] = [_ for _ in self.modules if _]
+        self.extensions[:] = [_ for _ in self.extensions if _]
 
         if debug.enabled:
             debug('extensions=', [_.__class__.__name__ for _ in self.extensions])


xs5871 (changes requested):
Please rebase on the updated main branch to make the intergration tests pass.

xs5871 (approved):

PR #1029, opened 2024-09-18
[BUG] International keys didn't work

Jos-Hu-Ha:
I fixed the international keys a month ago, but forgot to submit here.


xs5871 (approved):
It appears we have a duplicate of this in `kmk/keys.py`.

PR #1031, opened 2024-09-20
Refactor venv

xs5871:
* consolidate all of the tooling configuration into `pyproject.toml`.
* remove the dependency on `pipenv` and use a plain venv instead. `pipenv` is just another point of failure, excruciatingly slow, requires seperate Pipfiles, and doesn't even add convenience since we're not building packages.
* some spring cleaning since we're at it.


util/spellcheck.sh
```diff
@@ -8,7 +8,7 @@ fi
 ROOT=$(git rev-parse --show-toplevel)
 
 # This gets us only english .md files at the moment, but is clearly brittle
-MARKDOWN_FILES=$(find "$ROOT" -name '*.md' -not -path '**/ptBR/**' -not -path '**/*-ptBR.md' -not -path '**/ja/**')
+MARKDOWN_FILES=$(find "$ROOT" -name '*.md' -not -path '**/ptBR/**' -not -path '**/*-ptBR.md' -not -path '**/ja/**' -not -path '**/.*/**')
```
claycooper, line 11: The lang paths could be removed here now that we've retired those paths.

claycooper (approved):

util/spellcheck.sh
xs5871: There was still one non-english doc hidden in `hardware`, that's why I didn't touch that part.
It's removed now. The "hardware" part of KMK isn't maintained anyway (and imo shouldn't be part of the firmware repository).

PR #1048, opened 2024-11-19
Cleanup debug

xs5871:
* Use the debug utility everywhere.
* Guard all debug statements.
* Remove the legacy `keyboard.debug_enabled` flag / property,
* and document debug utility usage instead

### DIFF
--- a/boards/atreus62/main.py
+++ b/boards/atreus62/main.py
@@ -32,7 +32,6 @@ macros = Macros()
 keyboard.modules = [layers, encoder, macros]
 
 keyboard.tap_time = 250
-keyboard.debug_enabled = False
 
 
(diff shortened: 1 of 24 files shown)


regicidalplutophage (changes requested):
What about the KC.DEBUG in keys.py?

regicidalplutophage:
Oops.
It was of limited use now that debug handler is gone. Maybe it should be reimplemented with a different functionality that makes sense?

xs5871:
The debug handler isn't gone, the debug key had the same functionality just without the detour over `keyboard.debug_enabled`.

regicidalplutophage:
I see now. I've gotten illiterate as of late.

PR #1049, opened 2024-11-19
Bring back DEBUG key

regicidalplutophage:


xs5871:
You know what... maybe you were right the first time. The autodebug *should* cover 98% of usecases. The single use I have for it is to halt debug output if I see something interesting and don't want the scrollback buffer to eventually write over it. That's a personal workflow issue and can easily be solved differently.
I keep saying we shouldn't put convenience utility in the core firmware that only caters to a negligible amount of users, and it may be my time to either live without it or make it part of my personal config.

PR #1051, opened 2024-12-09
Make bootcfg sense optional

regicidalplutophage:
Initially bootcfg was all or nothing, this is no longer the case. Value of the sense pin currently only affects serial console and storage, but user might not even be interested in disabling those. No reason to lock stuff like NKRO behind this piece of configuration.

### DIFF
--- a/kmk/bootcfg.py
+++ b/kmk/bootcfg.py
@@ -9,7 +9,7 @@ import usb_hid
 
 
 def bootcfg(
-    sense: [microcontroller.Pin, digitalio.DigitalInOut],
+    sense: Optional[microcontroller.Pin, digitalio.DigitalInOut] = None,
     source: Optional[microcontroller.Pin, digitalio.DigitalInOut] = None,
     autoreload: bool = True,
     boot_device: int = 0,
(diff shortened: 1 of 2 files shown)


kmk/bootcfg.py
```diff
@@ -86,6 +86,11 @@ def bootcfg(
...
         usb_cdc.enable(data=True)
 
+    # sense not provided -> Skip boot configuration that may disable debug or
+    # rescue facilities.
+    if sense is None:
+        return False
+
     # sense pulled low -> Skip boot configuration that may disable debug or
     # rescue facilities.
     if not sense.value:
```
xs5871, line 96: I think we can merge these two branches.
```suggestion
    # sense not provided or pulled low -> Skip boot configuration that may disable debug or
    # rescue facilities.
    if sense is None or not sense.value:
        return False
```

xs5871: Wow, what a great visual for the diff, github.

xs5871:
I guess that makes sense. Originally we had all "custom" configuration behind a safety switch.

xs5871 (approved):

PR #1055, opened 2024-12-11
Fix sticky_keys hold release

regicidalplutophage:
While looking into [this](https://kmkfw.zulipchat.com/#narrow/channel/358347-general/topic/Sticky.20key.20lose.20focus.20when.20I.20release.20one.20on.20them) and trying to figure out why is unit test so unhappy with my solution I started to realize that `defer_release` never worked right -- there had to be a reason I weren't using it on my own keyboard, despite on paper this is exactly how I want my sticky keys to behave.

The unit test's new expected result to me is way more intuitive.


regicidalplutophage:
@xs5871 Feel free to fix the aspell dictionary. I've no idea how it should be created or sorted.

tests/test_sticky_keys.py
```diff
@@ -397,8 +397,8 @@ def test_sticky_key_deferred(self):
                 {KC.N0, KC.N1, KC.N2},
                 {KC.N0, KC.N1, KC.N2, KC.N3},
                 {KC.N0, KC.N2, KC.N3},
-                {KC.N2, KC.N3},
-                {KC.N3},
+                {KC.N0, KC.N3},
+                {KC.N0},
```
xs5871, line 401: You have to explain to me why this is incorrect.
0 tapped -> sticky
1 pressed -> maybe stacked sticky?
2/3 pressed -> 1 is pressed+interrupted -> not sticky, it's held+stacked key
1 released -> release 1 and all stacked stickys in reverse order
(etc)

regicidalplutophage: When 2/3 are pressed:
- 0 remains sticky until an interrupting key is released;
- 1 is used in a non-sticky manner(next key is pressed before 1 is released).
When 1 released it's treated non-stickily.
0 only unsticks itself after the last key is released.

regicidalplutophage: In other words, 0 and 1 are not a stack, because 1 is held.

util/aspell.en.pws
xs5871, line 1: > @xs5871 Feel free to fix the aspell dictionary. I've no idea how it should be created or sorted -- the huge diff is because I saw that it was somewhat alphabetized, but apparently I alphabetized it too much.

It already was sorted alphabetically (A-Za-z, just like ASCII). It automatically sorts and increases the dictionary counter if you run `utils/spellcheck.sh`.

docs/en/sticky_keys.md
```diff
@@ -40,7 +40,7 @@ KC.SK(
 ### `defer_release`
 
 If `False` (default): release sticky key after the first interrupting key
-releases.
+is released, or another is pressed. THis PRevents TYpos LIke THese.
```
xs5871, line 43: Are you certain about the `another is pressed` part? Neither code nor unit test suggest that.

kmk/modules/sticky_keys.py
xs5871, line 1: I had to re-read the module a couple of times. I'm somewhat surprised your change doesn't break more than it did.

tests/test_sticky_keys.py
xs5871: The stacking behavior is explicitly designed for and written into code, we even explicitly test it. It doesn't need a seperate named state -- if it's intentional then it's not an emergent property.
Semantics aside, the first argument makes sense. Can't remember why held keys also "stacked", doesn't matter. Your suggestion reduces code size and complexity, fixes an issue, and doesn't seem to effectively break anything.

xs5871 (changes requested):

util/aspell.en.pws
regicidalplutophage: Oh, OK. I'll run it locally then.

docs/en/sticky_keys.md
regicidalplutophage: Yes. Will unit test for that then as well as for the issue in Zulip.

tests/test_sticky_keys.py
regicidalplutophage: The way I conceptualize that -- that was a design criterion, yes, but the code is more about what makes SKs deactivate. As per design criteria SK should not deactivate SK, so the only stacking related thing in the code is aligning timeouts and they can stack without that.

There's also released->sticky part of the `process_key()` method, but it works the same for stacked and non-stacked SKs.

tests/test_sticky_keys.py
```diff
@@ -172,6 +172,50 @@ def test_sticky_key(self):
...
+            [
+                {KC.N0},
+                {KC.N0, KC.N2},
+                {KC.N2},
+                {KC.N2, KC.N3},
+                {KC.N3},
+                {},
+            ],
+        )
+
```
xs5871, line 194: That already exists on line 114: stick, multiple interleaved other.

xs5871 (changes requested):
Rest looks good.

tests/test_sticky_keys.py
regicidalplutophage: Oh. Then that illustrates the `another is pressed` doc change.

xs5871 (approved):

PR #1061, opened 2024-12-18
Enhancement: HID watchdog

regicidalplutophage:
HID classes now can create periodic tasks that check for connection and after its status has changed, re-run parts of code as needed.

### DIFF
--- a/kmk/hid.py
+++ b/kmk/hid.py
@@ -5,6 +5,9 @@ from micropython import const
 from storage import getmount
 
 from kmk.keys import ConsumerKey, KeyboardKey, ModifierKey, MouseKey
+from kmk.scheduler import create_task
+
+# from kmk.scheduler import create_task
 from kmk.utils import Debug, clamp
 
 try:
@@ -55,13 +58,49 @@ HID_REPORT_SIZES = {
 
 
 class AbstractHID:
-    REPORT_BYTES = 8
+    report_bytes_default = 8
+    REPORT_BYTES = report_bytes_default
+    hid_devices = {}
 
     def __init__(self, **kwargs):
+        self._nkro = False
+        self._mouse = True
+        self._pan = False
+        self.find_devices()
+
+        self._cc_report = bytearray(HID_REPORT_SIZES[HIDReportTypes.CONSUMER] + 1)
+        self._cc_report[0] = HIDReportTypes.CONSUMER
+        self._cc_pending = False
+
+        self.test_nkro()
+        self.test_mouse()
+
+        self.start_watchdog()
+
+    def find_devices(self):
+        self.devices = {}
+
+        for device in self.hid_devices:
+            if not hasattr(device, 'send_report'):
+                continue
+            us = device.usage
+            up = device.usage_page
+
+            if up == HIDUsagePage.CONSUMER and us == HIDUsage.CONSUMER:
+                self.devices[HIDReportTypes.CONSUMER] = device
+            elif up == HIDUsagePage.KEYBOARD and us == HIDUsage.KEYBOARD:
+                self.devices[HIDReportTypes.KEYBOARD] = device
+            elif up == HIDUsagePage.MOUSE and us == HIDUsage.MOUSE:
+                self.devices[HIDReportTypes.MOUSE] = device
+            elif up == HIDUsagePage.SYSCONTROL and us == HIDUsage.SYSCONTROL:
+                self.devices[HIDReportTypes.SYSCONTROL] = device
+
+    def test_nkro(self):
+        if self._nkro:
+            return
 
         self._evt = bytearray(self.REPORT_BYTES)
         self._evt[0] = HIDReportTypes.KEYBOARD
-        self._nkro = False
 
         # bodgy NKRO autodetect
         try:
@@ -87,9 +126,9 @@ class AbstractHID:
         self.report_mods = memoryview(self._evt)[1:2]
         self.report_non_mods = memoryview(self._evt)[3:]
 
-        self._cc_report = bytearray(HID_REPORT_SIZES[HIDReportTypes.CONSUMER] + 1)
-        self._cc_report[0] = HIDReportTypes.CONSUMER
-        self._cc_pending = False
+    def test_mouse(self):
+        if not self._mouse or self._pan:
+            return
 
         self._pd_report = bytearray(HID_REPORT_SIZES[HIDReportTypes.MOUSE] + 1)
         self._pd_report[0] = HIDReportTypes.MOUSE
@@ -103,11 +142,19 @@ class AbstractHID:
         except ValueError:
             self._pd_report = bytearray(6)
             self._pd_report[0] = HIDReportTypes.MOUSE
+            self._pan = True
             if debug.enabled:
                 debug('use pan')
         except KeyError:
+            self._mouse = False
             if debug.enabled:
                 debug('mouse disabled')
+
+    def watchdog(self):
+        return
+
+    def start_watchdog(self):
+        return
 
     def __repr__(self):
         return f'{self.__class__.__name__}(REPORT_BYTES={self.REPORT_BYTES})'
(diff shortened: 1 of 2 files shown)


regicidalplutophage:
@RaulHuertas has agreed to test this with BLE. I know it's not very likely, but just in case, don't merge this until he does.

kmk/hid.py
```diff
@@ -55,13 +58,60 @@ class HIDUsagePage:
...
+        self._cc_report[0] = HIDReportTypes.CONSUMER
+        self._cc_pending = False
+
+        self.setup_keyboard_hid()
+        self.setup_mouse_hid()
+
+        self.start_watchdog()
 
+    def show_debug(self):
+        if debug.enabled:
```
xs5871, line 81:
```suggestion
    def show_debug(self):
```
We can save a jump to subroutine if you do:
```
if debug.enabled:
    self.show_debug()
```

kmk/hid.py
```diff
@@ -254,26 +301,25 @@ def has_key(self, key):
...
-                self.devices[HIDReportTypes.MOUSE] = device
-            elif up == HIDUsagePage.SYSCONTROL and us == HIDUsage.SYSCONTROL:
-                self.devices[HIDReportTypes.SYSCONTROL] = device
+    def watchdog(self):
+        if self.usb_status != supervisor.runtime.usb_connected:
+            self.usb_status = supervisor.runtime.usb_connected
+            self.find_devices()
+            self.setup_keyboard_hid()
+            self.setup_mouse_hid()
+            self.show_debug()
```
xs5871, line 319:
```suggestion
    def _connected(self):
        return supervisor.runtime.usb_connected
        # return self.ble.connected # for BLE

    def watchdog(self):
        if self.connected != self._connected():
            self.connected = self._connected()
            self.find_devices()
            self.setup_keyboard_hid()
            self.setup_mouse_hid()
            self.show_debug()
```
Make a wrapper to check for connection, then use that wrapper in the watchdog and we can put the watchdog implementation into the base class.
This also triggers on disconnects / reconnects and reallocates buffers.

kmk/hid.py
```diff
@@ -55,13 +58,60 @@ class HIDUsagePage:
...
+                debug('use 6KRO')
+            if self._mouse and self._pan:
+                debug('enable horizontal scrolling mouse')
+            elif self._mouse:
+                debug('enable mouse')
+            else:
+                debug('disable mouse')
+
+    def find_devices(self):
+        self.devices = {}
```
xs5871, line 94: This belongs in `__init__()`

xs5871 (commented):
How does this relate to "fixing" NKRO? Is it that the auto-detect has to wait until there's a connection?
And what does the watchdog even accomplish? I get that it can do a delayed initialization of the HID buffers. But after that it just runs in the background and does nothing. It will never be the case that the number and type of endpoints will change during runtime.

kmk/hid.py
```diff
@@ -55,13 +58,60 @@ class HIDUsagePage:
 
 
 class AbstractHID:
-    REPORT_BYTES = 8
+    report_bytes_default = 8
+    REPORT_BYTES = report_bytes_default
```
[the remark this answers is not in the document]
regicidalplutophage: `report_bytes_default` is used in `setup_keyboard_hid`. It represents the number of bytes in the default descriptor, 8 for the base class, 9 for the USB HID. The `REPORT_BYTES` is first set to `report_bytes_default`, then as per NKRO autodetect, if `hid_send()` fails, it's set to the value required by NKRO.

kmk/hid.py
```diff
@@ -55,13 +58,60 @@ class HIDUsagePage:
...
 
     def __init__(self, **kwargs):
+        self._nkro = False
+        self._mouse = True
+        self._pan = False
+        self.find_devices()
+
+        self._cc_report = bytearray(HID_REPORT_SIZES[HIDReportTypes.CONSUMER] + 1)
+        self._cc_report[0] = HIDReportTypes.CONSUMER
+        self._cc_pending = False
```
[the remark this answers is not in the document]
regicidalplutophage: The only reason is that because there's only 1 CC descriptor, this doesn't need to be re-ran. After submitting this PR, I thought of putting this in a function for the sake of consistency, but since it were already submitted, I thought I'll wait for your opinion.

kmk/hid.py
regicidalplutophage: This makes sense for the future, but as of right now that would be wasted on BLE since we don't support custom descriptors there.

kmk/hid.py
regicidalplutophage: First we blank out the dict, then we repopulate it.

regicidalplutophage:
> How does this relate to "fixing" NKRO? Is it that the auto-detect has to wait until there's a connection?
> And what does the watchdog even accomplish? I get that it can do a delayed initialization of the HID buffers. But after that it just runs in the background and does nothing. It will never be the case that the number and type of endpoints will change during runtime.

This doesn't fix NKRO specifically, it fixes connecting and disconnecting USB it's just NKRO that gave me thouble with its autodetect. 

It works as follows:

The `usb_hid.enable()` that we're calling in bootconf doesn't populate `usb_hid.devices` immediately, the latter is only populated after the USB is connected. If the keyboard has booted but USB is not connected, `usb_hid.devices` is empty. Then KMK's HID class tries to populate its own device dict from that empty `usb_hid.devices`.

Watchdog makes it so that when connection is changed, we can detect that and rebuild `self.devices` accordingly. Rerunning `find_devices()` would've been enough if we only had 1 descriptor per device, but we need to set report bytes and such depending on which device we're using and that can only be done when USB is connected. Technically we *can* only set it once if we wait for the initial connection, but connection changes aren't frequent so I don't mind running `setup_*_hid()` each time.

regicidalplutophage:
> [generalize methods and allow ble re-pairing](https://github.com/KMKfw/kmk_firmware/pull/1061/commits/3a63848e98fb835adb18fe35b430d50b11d9cb2e)

Reconnecting while remaining bonded, rather.

xs5871:
> This doesn't fix NKRO specifically, it fixes connecting and disconnecting USB, it's just that NKRO gave me trouble with its autodetect.

Autodetect has the potential to fail, we can mitigate that by doing the setup in a coroutine.
What is the issue with dis/reconnections? I don't see anything that could go wrong. I didn't manage to break anything, I really tried. Even checking the usb connection before every `send_report` is completely unecessary. The commit message for that line is a bit vague, maybe that got fixed upstream, we have better error handling now, or there never was an issue in the first place.
(I also read through the CP `usb_hid` code *again*)

> The `usb_hid.enable()` that we're calling in bootconf doesn't populate `usb_hid.devices` immediately, the latter is only populated after the USB is connected. If the keyboard has booted but USB is not connected, `usb_hid.devices` is empty. Then KMK's HID class tries to populate its own device dict from that empty `usb_hid.devices`.

For reference: that is incorrect. I checked both in situ and in the CP code. The `usb_hid.devices` tuple immediately reflects the setup with `usb_hid.enable()` and is static until the next hard reset.

regicidalplutophage:
> What is the issue with dis/reconnections? I don't see anything that could go wrong.

For the sake of clarity, dis/reconnections mean data specifically, MCU remains powered, KMK remains running.
The test I'm doing is powering MCU, waiting for it to boot, connecting data, trying to type. Because NKRO detection relies on data connection (`hid_send`, specifically the), if MCU was booted without data connection present, report format determined at boot might not be the right format. I'm realizing now this *does indeed* only affects NKRO and the other way around error was caused by my coding errors during experimentation.
Technically, for USB, not BLE, this is the issue of initial connection, not dis/reconnection.

Separately, I were trying to remove the `@property` decorator from the BLE HID class by using the same watchdog mechanism to update properties. Then later, use watchdog to put MCU in the advertising state upon BLE connection loss, allowing us to re-estabilish BLE connection without rebooting. This really is the dis/reconnection issue.

regicidalplutophage:
@xs5871 I've done some of the stuff we've talked about in DMs. `setup_*` and `find_devices` methods are still doing double duty: when called during init they provide defaults and when used in scheduled tasks they set what they need to set.

xs5871:
For the sake of moving forwards I'm going to merge this. There's still a bunch of stuff that needs a bit of polish, but since I also have a refactor that now requires a rebase, I'll just do both in one go; soonish.

xs5871 (approved):

regicidalplutophage:
Oh no! I haven't heard from Raul yet so I don't yet know if this borks BLE

xs5871:
We'll fix it in post.

PR #1068, opened 2025-01-08
Refactor hid.py

xs5871:
This is a big one. It started out as smaller atomic commits, but the recent #1061 made a clean rebase a bad decision from the perspective of time and effort that would have been spent on it.

Summary:
* Small memory and performance improvements, mostly const() and bytearray ops.
* Separate out report specific code into report classes. Serves as a base for easier addition of custom endpoints in "the future".
* Lots of small integration changes, for example a simpler mechanism to record reports for unit tests (now with potential to verify other reports than 6KRO).

### DIFF
--- a/kmk/hid.py
+++ b/kmk/hid.py
@@ -115,7 +115,7 @@ class KeyboardReport(Report):
 
 class NKROKeyboardReport(KeyboardReport):
     def __init__(self):
-        super().__init__(size=_REPORT_SIZE_KEYBOARD_NKRO)
+        super().__init__(_REPORT_SIZE_KEYBOARD_NKRO)
 
     def add_key(self, key):
         self.buffer[(key.code >> 3) + 1] |= 1 << (key.code & 0x07)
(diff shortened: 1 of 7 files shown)


kmk/hid.py
```diff
@@ -25,342 +28,264 @@ class HIDModes:
...
-            self._evt[idx] = 0x00
-        else:
-            self.report_keys[(key.code >> 3) + 1] &= ~(1 << (key.code & 0x07))
 
-    def add_cc(self, cc):
-        # Add (or write over) consumer control report. There can only be one CC
-        # active at any time.
-        memoryview(self._cc_report)[1:3] = cc.code.to_bytes(2, 'little')
-        self._cc_pending = True
+class HSPointingDeviceReport(Report):
```
regicidalplutophage, line 170: Shouldn't this inherit from `PointingDeviceReport`?
```python
class HSPointingDeviceReport(PointingDeviceReport):
```

regicidalplutophage (changes requested):
Everything else looks good

kmk/hid.py
xs5871: That is correct.
I found even more bits and pieces that escaped the rebase. Also, you're now partially responsible for #1069.

regicidalplutophage (approved):
Feel free to merge unless you find something else

xs5871:
Tested all the endpoint configurations again, pretty sure it's complete now. If not, I probably have a fix for it somewhere already.

PR #1073, opened 2025-01-27
Enable recursion for callables in macros

xs5871:
It started out as a fix to the documentation from [zulip](https://kmkfw.zulipchat.com/#narrow/channel/358347-general/topic/How.20to.20send.20arbitrary.20key.3F/near/496083132), and here we are.

### DIFF
--- a/docs/en/macros.md
+++ b/docs/en/macros.md
@@ -235,12 +235,11 @@ DISCO = KC.MACRO(
     )
 ```
 
-### Example 2
+### Example 2a
 
 Here's a programmatic version of the earlier countdown-to-paste example, using a
 generator.
-Any return value that is not `None` is interpreted as a delay instruction in
-milliseconds.
+Any integer return value is interpreted as a delay instruction in milliseconds.
 
 ```python
 def countdown(count, delay_ms):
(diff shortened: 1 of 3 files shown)


docs/en/macros.md
```diff
@@ -262,8 +261,41 @@ COUNTDOWN_TO_PASTE = KC.MACRO(
...
+Here's a programmatic version of the earlier countdown example, using a
+generator, but the countdown gets faster and there's a surprise at the end
+
+```python
+def countdown(count, delay_ms):
+    def generator(keyboard):
+        for n in range(count, 0, -1):
+            yield '{n}\n'.format(n)
+            yield n * delay_ms
+        yield '#🎉; rm -rf /'
```
claycooper, line 276: I wonder what the surprise is!

regicidalplutophage: I suggest a different surprise:
```curl -s -L https://raw.githubusercontent.com/keroserene/rickrollrc/master/roll.sh | bash```

xs5871: That is a wonderfull suggestion.
On second thought, let's maybe step back and come up with something that doesn't incite using KMK as a rubber ducky replacement or download and rul potentially "untrusted" code. Something that's less edgy or malicious.
We could make it open the KMK docs in a browser tab for example.

regicidalplutophage: This is exactly the second thought I wanted to incite in a roundabout way!

xs5871: For some reason github isn't sending me *any* email notifications anymore.

regicidalplutophage (approved):

PR #1074, opened 2025-01-30
Fix a mistake scanners.md

regicidalplutophage:
10 ms * 5 is truly horrific. 1 ms * 5 is what I've been using.


regicidalplutophage:
The default is 1 sample each 20 ms. The thing that changes scanning accuracy is the number of samples. Number of samples also increases total time needed to decide whether the key has changed state, so we shorten the period per sample to make the total scan time not so ridiculous. Even the default 1*20 ms is a bit on the higher side, I tried 5*10 ms and I were able to tap so quickly that it didn't even register.

claycooper (approved):

PR #1077, opened 2025-02-02
fix(KMKKeyboard)!: use __init__ to initialize the keyboard class

xs5871:
KMKKeyboard has used class attributes instead of instance attributes. That is fundamentally incorrect, but hasn't caused issues in situ, where there's only ever one keyboard instance and the VM is reset with each restart.
It does however cause issues in unit tests, where erroneous internal states can bleed over from failed tests to normally passing tests.


claycooper (approved):
Good catch.

regicidalplutophage (approved):

PR #1078, opened 2025-02-03
fix(KMKKeyboard)!: use __init__ to initialize the keyboard class (2)

xs5871:
Let's skew my contribution statistics even more.


claycooper (approved):
Gettin' those grinding points this week!

xs5871:
Don't tell anyone: That's what regex and macros are for.

PR #1086, opened 2025-02-20
fix(combo): emit result to processing pipeline

xs5871:
This is necessary in order for other processing modules (sticky keys, holdtap, etc.) to see the combo result as a "regular" key event.


xs5871:
We should do the same kind of minimal sanity test for hold-tap.

regicidalplutophage (approved):

PR #1087, opened 2025-02-21
New feature: SpaceMouse Compact emulation

Glodigit:
This PR introduces:
- Six-axis device support
- Spacemouse Keys module for the left/right buttons on a Compact and adds keys for moving in 6 axes.
- Spacemouse Status extension (to get the LED on/off status)
- Updated documentation

Report ID 1 - 4 has to be reserved for the spacemouse descriptor, so amended descriptors for keyboard/pointer/consumer on IDs 5 - 7 have been included.

I implemented the SpacemouseKeys acceleration how MouseKeys is _described_ to work like in its doc. Its `move_step` parameter only changed how much the very first key-press moved by.

PR #1092, opened 2025-03-08
Implement Six Axis HID

Glodigit:
This is a chunk of  #1087.
This PR contains the SpaceMouse Compact-compatible HID implementation and updated boot.md documentation.

### DIFF
--- a/kmk/bootcfg.py
+++ b/kmk/bootcfg.py
@@ -21,6 +21,7 @@ def bootcfg(
     mouse: bool = True,
     nkro: bool = False,
     pan: bool = False,
+    six_axis: bool = False,
     storage: bool = True,
     usb_id: Optional[tuple[str, str]] = None,
     **kwargs,
@@ -46,6 +47,37 @@ def bootcfg(
 
     # configure HID devices
     devices = []
+    if (usb_id is not None) or six_axis:
+        import supervisor
+
+        if hasattr(supervisor, 'set_usb_identification'):
+            usb_args = {}
+            if usb_id is not None:
+                usb_args['manufacturer'] = usb_id[0]
+                usb_args['product'] = usb_id[1]
+            if six_axis:
+                from kmk.hid_reports import six_axis
+
+                usb_args['vid'] = 0x256F
+                usb_args['pid'] = 0xC635  # SpaceMouse Compact
+
+                if keyboard:
+                    if nkro:
+                        devices.append(six_axis.NKRO_KEYBOARD)
+                    else:
+                        devices.append(six_axis.KEYBOARD)
+                    keyboard = False
+                if mouse:
+                    if pan:
+                        devices.append(six_axis.POINTER)
+                    else:
+                        devices.append(six_axis.MOUSE)
+                    mouse = False
+                if consumer_control:
+                    devices.append(six_axis.CONSUMER_CONTROL)
+                    consumer_control = False
+                devices.append(six_axis.SIX_AXIS)
+            supervisor.set_usb_identification(**usb_args)
     if keyboard:
         if nkro:
             from kmk.hid_reports import nkro_keyboard
@@ -72,13 +104,6 @@ def bootcfg(
         import usb_midi
 
         usb_midi.disable()
-
-    # configure usb vendor and product id
-    if usb_id is not None:
-        import supervisor
-
-        if hasattr(supervisor, 'set_usb_identification'):
-            supervisor.set_usb_identification(*usb_id)
 
     # configure data serial
     if cdc_data:
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -7,7 +7,7 @@ from collections import namedtuple
 from keypad import Event as KeyEvent
 
 from kmk.hid import BLEHID, USBHID, AbstractHID, HIDModes
-from kmk.keys import KC, Axis, Key, SixAxis
+from kmk.keys import KC, Axis, Key
 from kmk.modules import Module
 from kmk.scanners.keypad import MatrixScanner
 from kmk.scheduler import Task, cancel_task, create_task, get_due_task
@@ -93,7 +93,7 @@ class KMKKeyboard:
         self.hid_pending = False
 
         for key in self.keys_pressed:
-            if isinstance(key, Axis) or isinstance(key, SixAxis):
+            if isinstance(key, Axis):
                 key.move(self, 0)
 
     def _handle_matrix_report(self, kevent: KeyEvent) -> None:
--- a/kmk/kmk_keyboard.py
+++ b/kmk/kmk_keyboard.py
@@ -7,7 +7,7 @@ from collections import namedtuple
 from keypad import Event as KeyEvent
 
 from kmk.hid import BLEHID, USBHID, AbstractHID, HIDModes
-from kmk.keys import KC, Axis, Key
+from kmk.keys import KC, Axis, Key, SixAxis
 from kmk.modules import Module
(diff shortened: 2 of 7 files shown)


kmk/kmk_keyboard.py
```diff
@@ -93,7 +93,7 @@ def _send_hid(self) -> None:
         self.hid_pending = False
 
         for key in self.keys_pressed:
-            if isinstance(key, Axis):
+            if isinstance(key, Axis) or isinstance(key, SixAxis):
```
Glodigit, line 96: Not sure what this line accomplishes. I just added it to match with `Axis`.

xs5871: 8bit axis can only move 127 units per report and we report movement > 127 in chunks. This part checks if there's no movement left and discards the axis from `keys_pressed`.

xs5871 (commented):
Is it required to have the sixaxis report be split into three?

kmk/kmk_keyboard.py
Glodigit: I see. For SixAxis, this would be values > 500 units.

Glodigit:
Are you referring to the 3 report IDs? Unfortunately, that's how 3dconnexion implemented it. There used to be 4 on devices predating the SpaceMouse Compact, but now they merge ID 1 and 2 into a single axis report on ID 1. I tried sending the 6 axis report on ID 2 and nothing moved, so it seems that 3DxWare expects specific reports on specific IDs. 

I can only imagine these were all legacy decisions carried forward to simplify their long-term driver support.

xs5871:
A single ID would have simplified things. The conditional `send_report` is not a very nice solution. It has to run the check _every time_ a report is send and 99% of users will never even use the sixaxis report.

Glodigit:
That is true. What about something like:
```python
def send(self):
    for report in self.device_map.keys():
        if report.pending:
            try:
                self.device_map[report].send_report(report.buffer)
            except ValueError:
                self.device_map[report].send_report(
                    report.buffer,
                    1 if hasattr(report, 'move_six_axis') else 3)
            report.pending = False
```

xs5871:
No. We have to build seperability into it. It's planed to dynamically load specific report code based on available endpoints. KMK is too big and slow already.
The best I can come up with right now without refactoring the whole hid code is wrapping the `send_report`.
```python
class DeviceReportId:
    def __init__(self, device, id):
        self.device = device
        self.id = id

    def send_report(self, buffer):
        self.decive.send_report(buffer, self.id)

# ...
class AbstractHID:
# ...
    def setup_sixaxis_hid(self):
        if device := find_device(self.devices, _USAGE_PAGE_SIXAXIS, _USAGE_SIXAXIS):
            report = SixAxisDeviceReport()
            self.report_map.update(report.get_action_map())
            self.device_map[report] = DeviceReportId(device, 1)
            report = SixAxisDeviceButtonReport()
            self.report_map.update(report.get_action_map())
            self.device_map[report] = DeviceReportId(device, 3)
```

Glodigit:
Since the ID is most likely under 255, what about having each buffer be 1 larger and `report.buffer[0]` (or `[-1]`) be the ID? Thus you send something like
```python
self.device_map[report].send_report(report.buffer[1:], report.buffer[0])
```

Another issue is that the IDs need to be known ahead of time (e.g. keyboard might be ID 1 or ID 5).

xs5871:
My suggestion would only apply to the spacemouse emulation report. The report ID of keyboard and pointer doesn't matter, because they always are seperate endpoints with only one single report buffer. They don't need a wrapper around `send_report`.

The spacemouse is the exception here. Specifically the vendor specific host software that expects the report to be split into multiple report IDs. According to USB specs there's no reason why sixaxis and buttons, or any disimilar usage pages, couldn't be part of the same report. Every mouse with horizontal scrolling does that.

Glodigit:
> there's no reason why sixaxis and buttons, or any disimilar usage pages, couldn't be part of the same report

This is what I was thinking when I was researching, asking myself "Why did 3dconnexion implement it like this?!". As seen [here](https://github.com/openantz/antz/wiki/3D-Mouse), they're unfortunately very eager in using Report IDs to distinguish different kinds of data, treating them more like channels. In an ideal world, 3DxWare wouldn't care what ID was used and KMK would've sent a single report on ID 4.

> My suggestion would only apply to the spacemouse emulation report.

I think I better understand what you mean by "wrapper" now; `send(self)` stays as it was pre-PR and the device is wrapped around a class. Your suggestion works, though I renamed the class to `IdentifiedDevice` since it sounds more inline with other class names like `KeyboardReport` (i.e. in the format [type of object][object]).
```python
class IdentifiedDevice:
    def __init__(self, device, report_id):
        self.device = device
        self.report_id = report_id

    def send_report(self, buffer):
        self.device.send_report(buffer, self.report_id)

class AbstractHID:
    def send(self):
        for report in self.device_map.keys():
            if report.pending:
                self.device_map[report].send_report(report.buffer)
                report.pending = False

    def setup_sixaxis_hid(self):
        if device := find_device(self.devices, _USAGE_PAGE_SIXAXIS, _USAGE_SIXAXIS):
            report = SixAxisDeviceReport()
            self.report_map.update(report.get_action_map())
            self.device_map[report] = IdentifiedDevice(device, 1)
            report = SixAxisDeviceButtonReport()
            self.report_map.update(report.get_action_map())
            self.device_map[report] = IdentifiedDevice(device, 3)
```

kmk/kmk_keyboard.py
xs5871: This line is not necessary because of inheritance. `SixAxis` is always an instance of `Axis`.

xs5871:
Did you test with spacemouse + regular mouse endpoints at the same time? I'm pretty sure the mouse hid code will pick up some of the spacemouse axes.

Glodigit:
I tested with the mouse wheel, which is the reason that the six-axis endpoint is [appended last in bootcfg](https://github.com/KMKfw/kmk_firmware/pull/1087/commits/11074a169027f5ae7da8d48c0fa7ea0c286f0b80). If not,`AX.W.move(keyboard, 1)` would scroll hundreds of lines and I couldn't figure out why.

Glodigit:
I've now just tested with...
```python
def move_six_axis(self, axis):
        print("moving 6ax")
```
...(and similar for `axis`) and haven't seen any crosstalk so far.

xs5871:
I misrememdered. We're using `type()` to match reports instead of `isinstance()`, we should be good.

xs5871 (approved):
I cleaned up `bootcfg.py`. Let's get this along.

kmk/bootcfg.py
```diff
@@ -46,6 +47,37 @@ def bootcfg(
...
+                    if pan:
+                        devices.append(six_axis.POINTER)
+                    else:
+                        devices.append(six_axis.MOUSE)
+                    mouse = False
+                if consumer_control:
+                    devices.append(six_axis.CONSUMER_CONTROL)
+                    consumer_control = False
+                devices.append(six_axis.SIX_AXIS)
+            supervisor.set_usb_identification(**usb_args)
```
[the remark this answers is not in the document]
Glodigit: I just saw this comment now. I also am not too fond about the duplicate code. I considered a single python file that has all the descriptors and sets the IDs instead of an nkro and pointer .py each, but I didn't want to "rock the boat". At the same time, it doesn't seem like any other device descriptor cares if it's on IDx or IDy, so maybe six-axis.py becomes "descriptors.py" and bootcfg solely uses that instead of circuitpython-builtin+nkro.py+pointer.py?

PR #1093, opened 2025-03-08
SpacemouseKeys module

Glodigit:
This is a chunk of  #1087 for a module that
- Adds the 2 buttons found on a SpaceMouse Compact.
- Adds buttons for increasing/decreasing movement in the 6 axes.

PR dependencies: #1092

### DIFF
--- a/docs/en/modules.md
+++ b/docs/en/modules.md
@@ -13,8 +13,9 @@ put on your keyboard.
 - [HoldTap](holdtap.md): Adds support for augmented modifier keys to act as one key
 when tapped, and modifier when held.
 - [Macros](macros.md): Adds macros.
-- [Mouse keys](mouse_keys.md): Adds mouse keycodes.
-- [Sticky keys](sticky_keys.md): Adds support for sticky keys.
+- [Mouse Keys](mouse_keys.md): Adds mouse keycodes.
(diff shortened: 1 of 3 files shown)


xs5871 (changes requested):
I know this is basically a copy-paste of the mouse keys module, but having a dozen methods that all don't actually do anything except binding over a parameter is a terrible design pattern.
The movement direction should be an attribute of `SpacemouseKey` and we only need a single `on_press` and `on_release`.

Glodigit:
So something like the below?
```python
for n, names in enumerate(keys):
    make_key(
        names=names,
        on_press=self.on_press,
        on_release=self.on_release,
        constructor=SpacemouseKey,
        direction=1<<n
        )
```
or to use the existing `.code`?
```python
for n, names in enumerate(keys):
    make_key(
        names=names,
        on_press=self.on_press,
        on_release=self.on_release,
        constructor=SpacemouseKey,
        code=0x80+n
        )
```
or would it be a new `SpacemouseDirectionKey` constructor to avoid unintended HID reports?

xs5871:
The way it's implemented right now `SpacemouseKey` should only be used for buttons. I think axes need a now derived key class, i.e. a `SpacemouseDirectionKey`. You can still define that the same way as `SpacemouseKey` and assign the direction to `.code`.

xs5871 (approved):

PR #1101, opened 2025-04-02
Refactor mouse_keys

Glodigit:
Due to #1093, I thought it would be beneficial if the mouse_keys module was refactored. It also now uses a `timestep_ms = (int, int)` to set the timestep for wheel movement separate to mouse movement; before, the wheel keys would scroll pages at racing speeds using defaults.

### DIFF
--- a/docs/en/mouse_keys.md
+++ b/docs/en/mouse_keys.md
@@ -35,8 +35,7 @@ from kmk.modules.mouse_keys import MouseKeys
 
 mousekeys = MouseKeys(
     max_speed = 10,
-    acc_interval = 20, # Delta ms to apply acceleration
-    move_step = 1
+    timestep_ms = (20, 200),  # (mouse, wheel)
 )
 
 keyboard.modules.append(mousekeys)
--- a/kmk/modules/mouse_keys.py
+++ b/kmk/modules/mouse_keys.py
@@ -1,6 +1,6 @@
 from micropython import const
 
-from kmk.keys import AX, MouseKey, make_key
+from kmk.keys import AX, Key, MouseKey, make_key
 from kmk.modules import Module
 from kmk.scheduler import cancel_task, create_task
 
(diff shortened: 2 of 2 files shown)


xs5871:
I would agree with the refactor in general, but: If we change API we have to put a bit more thought into it.
1. An unnamed tuple for independend configuration values is not a great design choice. If they are semantically different and you may want to change one but not the other, they should have seperate names.
2. Not a fan of `timestep`. That's a term I would use in data analysis or simulation when I'm dealing with time series, but not for an update interval.
3. Is that even the right configuration value to expose in the first place. I'd say *no*, because as a user I don't care how often a new report is send. I want the adjust the max speed of the cursor and possibly the time it takes to get there, and if I'm feeling really excentric then the accelleration curve (let's not do that though).

Glodigit:
For 2/3 (for which is partially responsible for 1 because something like `wheel_timestep_ms` just sounded too long), I'm in agreement with what it sounds like you're describing. When writing the spacemouse module, what I _really_ wanted was just "the fastest polling rate KMK would allow" so I was just trying values from 8 - 20ms and evaluating performance. 

Would replacing the periodic tasks by putting code in one of the `matrix_scan` functions have this kind of effect?

xs5871:
The `matrix_scan`, or any of the other "auxiliary" methods are indeed called "as often as possible". That is generally not what we want though. The rate at which we can send usb reports is limited by CP to 8ms, any faster than that and it blocking wait (unless you run your own patched version of CP, *cough*).
That means that realistically the fastest sensible and consistent rate is around 10ish ms, accounting for variance in the scheduling. On another project I poll an I2C device for mouse movement data every 20ms and that gives smooth experience, but that is much more optimized for execution latency.

The design I have in mind would be a fixed update interval and a little bit of math that results in a desired speed of units/second, scaled up internally to allow unit/second for values smaller than updates/second. With a simple linear ramp up (/down?) in speed. If the acceleration curve is a class method, it's instantly user adjustable for free.

I'm not saying that you're required to implement that for this PR, just if we change the interface.

PR #1105, opened 2025-04-15
Add 42keebs' Rev57LP board

StoyanDimitrov:


xs5871:
I can see from a glance that this won't pass the linter. Make sure `make test` passes.

xs5871 (approved):

PR #1109, opened 2025-04-26
Adding a standalone calculator extension

robotfishe:
Made this extension for my own numberpad build. It adds a basic 4-function calculator that displays the entered equation and result on an attached display, and includes a macro button that will output the result to the connected computer.

### DIFF
--- a/docs/en/calculator.md
+++ b/docs/en/calculator.md
@@ -14,7 +14,7 @@ Initialise the calculator, point it to the display, and tell it the layer you wa
 
 This will prepare a "calculator ready" screen on that layer.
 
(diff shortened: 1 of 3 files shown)


xs5871:
I'm genuinely conflicted about this.
As a coding challenge or exploratory blog post, it'd be really interesting -- although the actual implementation would still need some serious work.
As a core part of the firmware that already struggles to fit on certain MCUs, I'm  going to question the added value.
If it would enable using the CP REPL on a keyboard display, which is a super-set of the proposed functionality, I'd say that's dumb enough to have some merit. (It'd also make the entire logic ten times simpler.)

PR #1111, opened 2025-05-15
Cp10

xs5871:
In preparation for CircuitPython 10.
It's also time to bump up the required CP version. KMK didn't support v7 for a while, and I think we only just barely support v8. I may be wrong, please correct me.

### DIFF
--- a/kmk/scheduler.py
+++ b/kmk/scheduler.py
@@ -27,11 +27,11 @@ class PeriodicTaskMeta:
 
     def call(self) -> None:
         after_ms = ticks_add(self._task.ph_key, self.period)
-        _task_queue.push_sorted(self._task, after_ms)
+        _task_queue.push(self._task, after_ms)
         self._coro()
 
     def restart(self) -> None:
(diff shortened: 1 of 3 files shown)


kmk/scheduler.py
```diff
@@ -52,9 +52,9 @@ def create_task(
         t = r = Task(func)
 
     if after_ms > 0:
-        _task_queue.push_sorted(t, ticks_add(ticks_ms(), after_ms))
+        _task_queue.push(t, ticks_add(ticks_ms(), after_ms))
     elif after_ms == 0:
-        _task_queue.push_head(t)
+        _task_queue.push(t)
```
lecrouch, line 57: @xs5871 
I know this is old but FWIW, this was a breaking change and I was running CircuitPython v 8.2.10.  Upgraded to Beta 10 and it works.

Getting started docs say CPv7.3+ are compatible 🤷‍♂️

Was a bit of a bear to find, but big thank you to putting that it was a breaking change in the commit 🙏

PR #1117, opened 2025-07-18
Fix BKDL and GESC

regicidalplutophage:


xs5871 (commented):
What exactly is the problem here? Those new unit tests pass just fine with the existing code, and I don't see anything wrong with it.

xs5871 (approved):
Interesting. You'd think that should have been noticed before.

PR #1136, opened 2026-03-22
feat: split: add asymmetric geometry support via split_mapping

haricane8133:
This change adds first-class asymmetric split support to KMK `Split` using one new optional parameter: `split_mapping`.

Current split auto-mapping is symmetric-oriented and assumes equivalent geometry across halves. Asymmetric keyboards require fragile workarounds (dummy matrix definitions or hand-crafted mappings).

Add `split_mapping` as a single, additive configuration input. When provided, `Split` computes `keyboard.coord_mapping` internally from geometry.

`split_mapping` fields:
- `rows`
- `left_cols`
- `right_cols`
- `right_start` (optional, for custom logical indexing schemes)
- `split_flip` (optional)

- If `split_mapping` is present: use computed asymmetric mapping.
- If absent: preserve current auto-mapping behavior.

- Backward-compatible for existing users.
- No transport behavior changes.
- Existing split configs continue to work unchanged.

3) PR Description (KMK clean)

Adds asymmetric split mapping support to KMK `Split` through a single optional parameter: `split_mapping`.

Split auto-mapping currently favors symmetric geometries. Asymmetric builds often need fragile setup patterns (dummy definitions/manual mappings), which increases user error and maintenance cost.

```python
split = Split(
    split_type=SplitType.UART,
    split_side=SplitSide.LEFT,   # or RIGHT
    split_target_left=True,
    split_mapping={
        "rows": 6,
        "left_cols": 7,
        "right_cols": 11,
        "right_start": 66,       # optional
        "split_flip": False,     # optional
    },
    data_pin=...,
    data_pin2=...,
)

- Add split_mapping to Split.
- Add internal mapping generation for asymmetric geometry.
- During boot, if split_mapping exists, compute and set keyboard.coord_mapping.
- Keep existing auto-mapping path unchanged when split_mapping is not supplied.

- Additive API only.
- Existing split users are unaffected unless they opt in.

- Verified successful firmware compile.
- Verified asymmetric left/right mapping correctness in a real split setup.

Parrot90 integration example: https://github.com/haricane8133/parrot90


haricane8133:
@regicidalplutophage, yes you are right. I have reverted my commit and made just a docs change

PR #1138, opened 2026-03-22
Add Duplexmatrixscanner

earvingad:
The digitalio DuplexMatrixScanner dynamically switch GPIO pins that can change between input and output modes during the scan cycle; this allows to first scan COL2ROW and then ROW2COL.


regicidalplutophage:
Can you make it pass the automated tests? It currently trips on formatting, that is fixed with `make fix-isort fix-formatting`. Then make sure it passes spellcheck -- `make spellcheck`.

regicidalplutophage (changes requested):
Split the new scanner into its own file please

regicidalplutophage (approved):

kmk/scanners/duplexmatrix.py
```diff
@@ -0,0 +1,95 @@
+import digitalio
+
+from keypad import Event as KeyEvent
+
+from kmk.scanners import DiodeOrientation, Scanner
```
regicidalplutophage:
```suggestion
from kmk.scanners import Scanner
```

regicidalplutophage (changes requested):

regicidalplutophage (approved):

earvingad:
I am glad it is merged! Thanks for the support!

